PR #3, opened 2022-04-12
[move-lang][vm] Allow non-`'static` native context extensions.

wrwg:
(This is a port of [PR #208](https://github.com/diem/move/pull/208) from the old repo where it was already approved.)

This uses the `Tid<'a>` type from the [better_any](https://docs.rs/better_any/latest/better_any/) crate which acts as a replacement for the standard `Any` type in order to implement `NativeContextExtensions`. This allows the polymorphic values stored in the extensions to be non-`'static`, which `Any` does not.

As a consequence, it is now `NativeContextExtensions<'a>`, where `'a` is placeholder for the lifetime of any references stored in some extensions. Only one lifetime parameter is allowed for such extensions because of restrictions of `better_any`.

A user of extensions should not see any of this, however, structures describing extensions must use `#[derive(Tid)]`, and any type parameters those have must do as well. This is so the crate can compute a `TypeId` for such types.

The `NativeContextExtensions<'a>` type has been moved in its own module, and a test has been added to verify that this new feature works.

## Motivation

Allow extensions to aggregate references.

## Test Plan

Added tests


wrwg:
I can't add @msmouse yet, he need to first accept his invite.

msmouse (approved):

wrwg:
@bmwill @sblackshear @tnowacki 

There were quite a few things needed to get this finally working, but now it seems to do.

- Needed a custom build profile for CI test to get past the linker errors. `profile.ci` now defines opt-level = 3 and sets debug = 1, to include still line # info, but exclude all other debug info. My guess is the last step did the job, we can play with decreasing the opt level (since compiling with this level is *really* slow with the github action machines; compilation takes like 20 minutes)
- In order to use custom profiles, need to upgrade rust toolchain to 1.60.0, because it is unstable with our current toolchain.
- The new tool chain produces dozens and dozens of xclippy errors all over the code. It is unclear to me why, as some of those errors where fairly basic and not new (e.g. `|x| f(x)` --> `f`). I fixed most of them, but at some point lost patience with this, and added a few new exclusion rules to x.toml. We should followup on removing those exceptions and fixing the code.
- With a custom profile, the move-cli test suite was broken, because it expects the program to be either in target/debug or target/release. Fortunately, cargo sets an env var to read the binary path for integration tests.

All those changes are amended to this PR, which served as my test case. While this is not very clean, I'm going to land this as soon as all checks are green to unblock other PRs.

PR #4, opened 2022-04-12
Macros

wrwg:
This PR realizes the 1st step towards a high fidelity implementation of _macros_. Macros are expanded at compile time and do not appear in the Move bytecode. Most importantly,  macros can take lambda expressions as parameters, which allows to define collection operations like contains, filter, foreach, etc. 

A macro is declared similar to a function; below is an example:

```
    public macro fun foreach<T>(v: &vector<T>, action: |&T|) {
        let i = 0;
        while (i < Vector::length(v)) {
            action(Vector::borrow(v, i));
            i = i + 1;
        }
    }

   // Using the macro
   fun use_foreach() {
        let v = vector[1, 2, 3];
        let sum = 0;
        foreach!(&v, |e| sum = sum + *e);
    }

```

This PR implements parsing and type checking of this style of macros up to the typing AST. Macro expansion is not yet implemented and intended to be added in a followup.

The biggest change is enabling function types, lambda expressions, and calls of function values for naming and typing. Those had been supported before for the spec language, but weren't supported in move-lang beyond expansion. In the design, we represent function types by a new `BuiltinTypeName_::Fun` which allows to minimize changes to the code, as no new type variant needs to be introduced, and existing functionality for inference et. al can be adopted with minimal changes. This consistent e.g. with the treatment of function types in Rust, where they are just type constructors with some syntactic sugar.

A new test case has been added at [lambda.move](language/move-compiler/tests/move_check/typing/lambda.move) which shows usage and intends to cover all type check failure scenarios.

After this PR, to complete this style of macros, we still need to:

- Implement actual expansion of macros. This is intended to happen in a new phase after typing.
- Implement representation of macros in compiled, persisted modules. We need to be able to fetch the macro definition from an imported, precompiled module.
- Implement techniques (in the prover) to refer to specifications of passed lambda arguments to macros. (See https://github.com/diem/move/blob/main/language/move-prover/tests/sources/functional/macro_verification.move for what we want to do here.)

## Motivation

Macros

Yes

## Test Plan

Added new tests

### DIFF
--- a/language/move-compiler/src/naming/ast.rs
+++ b/language/move-compiler/src/naming/ast.rs
@@ -109,6 +109,7 @@ pub type FunctionBody = Spanned<FunctionBody_>;
 #[derive(PartialEq, Debug, Clone)]
 pub struct Function {
     pub attributes: Attributes,
+    pub is_macro: bool,
     pub visibility: Visibility,
     pub entry: Option<Loc>,
     pub signature: FunctionSignature,
@@ -154,6 +155,8 @@ pub enum BuiltinTypeName_ {
     Vector,
     // bool
     Bool,
+    // Function (last type arg is result type)
+    Fun,
 }
 pub type BuiltinTypeName = Spanned<BuiltinTypeName_>;
 
@@ -166,6 +169,12 @@ pub enum TypeName_ {
     ModuleType(ModuleIdent, StructName),
 }
 pub type TypeName = Spanned<TypeName_>;
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
+pub enum Variance {
+    CoVariant,
+    ContraVariant,
+}
 
 #[derive(Debug, Hash, Eq, PartialEq, Ord, PartialOrd, Copy, Clone)]
 pub struct TParamID(pub u64);
@@ -239,9 +248,11 @@ pub enum Exp_ {
     ModuleCall(
         ModuleIdent,
         FunctionName,
+        bool,
         Option<Vec<Type>>,
         Spanned<Vec<Exp>>,
     ),
+    VarCall(Var, Spanned<Vec<Exp>>),
     Builtin(BuiltinFunction, Spanned<Vec<Exp>>),
     Vector(Loc, Option<Type>, Spanned<Vec<Exp>>),
 
@@ -249,6 +260,7 @@ pub enum Exp_ {
     While(Box<Exp>, Box<Exp>),
     Loop(Box<Exp>),
     Block(Sequence),
+    Lambda(LValueList, Box<Exp>),
 
     Assign(LValueList, Box<Exp>),
     FieldMutate(ExpDotted, Box<Exp>),
@@ -343,6 +355,7 @@ impl BuiltinTypeName_ {
     pub const U_256: &'static str = "u256";
     pub const BOOL: &'static str = "bool";
     pub const VECTOR: &'static str = "vector";
+    pub const FUN: &'static str = "|..|..";
 
     pub fn all_names() -> &'static BTreeSet<Symbol> {
         &BUILTIN_TYPE_ALL_NAMES
@@ -390,10 +403,11 @@ impl BuiltinTypeName_ {
             }
             B::Signer => AbilitySet::signer(loc),
             B::Vector => AbilitySet::collection(loc),
+            B::Fun => AbilitySet::functions(),
         }
     }
 
-    pub fn tparam_constraints(&self, _loc: Loc) -> Vec<AbilitySet> {
+    pub fn tparam_constraints(&self, _loc: Loc, arity: usize) -> Vec<AbilitySet> {
         use BuiltinTypeName_ as B;
         // Match here to make sure this function is fixed when collections are added
         match self {
@@ -407,6 +421,25 @@ impl BuiltinTypeName_ {
             | B::U256
             | B::Bool => vec![],
             B::Vector => vec![AbilitySet::empty()],
+            B::Fun => (0..arity).map(|_| AbilitySet::empty()).collect(),
+        }
+    }
+
+    pub fn variance(&self, pos: usize, arity: usize) -> Variance {
+        match self {
+            // Function variance: given g: T1 -> R1 and f: T2 -> R2, then
+            // f can substitute g if T2 >= T1 && R1 <= R2
+            BuiltinTypeName_::Fun if pos < arity - 1 => Variance::ContraVariant,
+            _ => Variance::CoVariant,
+        }
+    }
+}
+
+impl TypeName_ {
+    pub fn variance(&self, pos: usize, arity: usize) -> Variance {
+        match self {
+            TypeName_::Builtin(bn) => bn.value.variance(pos, arity),
+            _ => Variance::CoVariant,
         }
     }
 }
@@ -486,7 +519,7 @@ impl Type_ {
                 Some(AbilitySet::primitives(b.loc))
             }
             B::Signer => Some(AbilitySet::signer(b.loc)),
-            B::Vector => None,
+            B::Vector | B::Fun => None,
         };
         let n = sp(b.loc, TypeName_::Builtin(b));
         Type_::Apply(abilities, n, ty_args)
@@ -595,6 +628,7 @@ impl fmt::Display for BuiltinTypeName_ {
                 BT::U256 => BT::U_256,
                 BT::Bool => BT::BOOL,
                 BT::Vector => BT::VECTOR,
+                BT::Fun => BT::FUN,
             }
         )
     }
@@ -732,6 +766,7 @@ impl AstDebug for (FunctionName, &Function) {
             name,
             Function {
                 attributes,
+                is_macro,
                 visibility,
                 entry,
                 signature,
@@ -747,7 +782,11 @@ impl AstDebug for (FunctionName, &Function) {
         if let FunctionBody_::Native = &body.value {
             w.write("native ");
         }
-        w.write(&format!("fun {}", name));
+        if *is_macro {
+            w.write(&format!("macro {}", name));
+        } else {
+            w.write(&format!("fun {}", name));
+        }
         signature.ast_debug(w);
         if !acquires.is_empty() {
             w.write(" acquires ");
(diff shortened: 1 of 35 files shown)


wrwg:
@bmwill Awesome, CI seems to work.

bmwill:
@wrwg unfortunately it does not yet

language/move-compiler/tests/move_check/typing/lambda.move
```diff
@@ -0,0 +1,100 @@
...
+
+    public fun correct_reduce(): u64 {
+        let v = vector[1, 2, 3];
+        reduce!(v, 0, |t, r| t + r)
+    }
+
+    public fun corrected_nested() {
+        let v = vector[vector[1,2], vector[3]];
+        let sum = 0;
+        foreach!(&v, |e| sum = sum + reduce!(*e, 0, |t, r| t + r));
```
davidiw, line 34: I might be nitpicking here, but the lack of separator between the input and output is slightly jarring, is there no way to provide a ":" syntax here as well in the macro action?

davidiw (commented):
This is really great, there were many times back when working on validator functions where we'd want to basically return a reference, but move didn't (does it now?) support returning a reference from a function. There was also no lambda / macro support, so there was a bit of copypasta

language/move-compiler/tests/move_check/typing/lambda.move
wrwg: Do you mean the lambda or the function type signature? The lambda is obviously inspired by Rust, I feel its fine.

The signature is a different problem. As we do not use `->` but `:` for return types, we would go strange looking stuff like
`fun foo(p: |int|:bool)` (notice the consecutive colons). I tried to align the current syntax with lambda, and once used to it, I like it ;)

tnowacki (commented):
Overall, I like where this is going. Big fan of the `|T, T, T|T` syntax! I like syntax where the type is matches the expression like this! (I was, and still am, always bothered by the syntax soup around lambdas in Rust)

I have some concerns about the usability of the macros, just in how it lines up with the rest of the system. Maybe, @wrwg, if you could walk me through you you see this working out that would be helpful :) 

To be clear *none of these are blocking landing*. We don't have to solve all these today. I just want to make sure we are set up for being able to solve these in the future (which might influence this PR)

### 1. Type parameters don't normally take references or "tuples"

At the bytecode level, references cannot instantiate type parameters. And at the bytecode level, these "tuples" don't even exist. 
So in the source language, every type parameter has to be instantiated with a single non-reference type. 
This will be unnecessarily restrictive, and kind of annoying for some of these macros. For example I imagine it would be nice for `Vector::reduce` to have the accumulator be a "tuple", e.g. 
```
let v: vector<u64> = ...;
let (num_even, num_odd) = Vector::reduce(
    v, 
    (0, 0), 
    |x, (even, odd)| if (x % 2 == 0) even = even + 1 else odd = odd + 1,
});
```
But this gets a bit tricky if we just remove the single type constraint from macro type parameters in that we need to be sure it is still applied in the appropriate spots (which I think it should happen automatically, even in cases like `macro fun foo<T>(v: vector<T>, f: ||T)` and 
`foo!<(u64, bool)>(...)`)

My bigger concern really is, do you think this will be confusing to programmers? (That the rules
for type parameters are different for macro vs non-macro funs)

### 2. Substitution and shadowing 

There is a whole can of worms once we try to expand the macro. The two that I'm most concerned about 
are 
A. shadowing macro parameters 
B. "captured" local variables in lambdas 

For (A), I see two possibilities, we could either have a check that says "You cannot shadow or assign macro fun parameters" or we could put `$` in front of macro parameters so they are in a different name space from other variables

For (B), I think we should probably just not allow capturing of locals in these lambdas in the short term? I think it will just make the macro expansion a lot cleaner and we can always come back to later 

### 3. Receiver syntax inside lambdas 

I think this is a non-issue as long as we restrict macro lambda usage heavily (which I think we have to anyway), but I'll write it up anyway.

Normally, when type checking the body of lambdas you have that you don't really want to type check the body until the lambda is "used". This is so you can be more permissive in not requiring annotations. For example,
```
let v: vector<MyStruct> = ...;
let f = |x| x.foo();
some_cool_macro!(x, f)
```
If you type check `|x| x.foo()` when it is first seen, you have to error since you don't know what `x` is. But if you wait until after it is "used" (like in the call `some_cool_macro`), you can give a better programming experience.

I think this shouldn't be an issue because we likely won't support anything other than using the lambda in the macro call, e.g. `let v: vector<MyStruct> = ...; some_cool_macro!(x, f)`

### 4. Receiver syntax inside macro funs

Where did we land on type checking the body of `macro fun`s?
I remember discussing this particular issue but don't remember where we landed.

It could be annoying if we don't (since you might see an error over and over again like you do in Rust macro expansions)
But conversely it would be more expressive if we don't!
That way you could write macros such as 
```
macro fun call_foo<T>!(v: &vector<T>) {
    Vector::foreach!(v, |x| x.foo())
}
```

As I write this out, I think you will want to expand the macro *during* type checking, which will I think mean you cannot type check the `macro fun` before it is used everywhere. 

But let me know your thoughts

### 5. Errors in macro expansion 

I think a managable concern but just something I thought of: 

I think we want to make sure we add a note or label of some sort to any diagnostic that happens inside of the code of an expanded macro. We could maybe accomplish this by adding a node `UnannotatedExp_::ExandedMacro(Box<Exp>, /* inside of */ (ModuleIdent, FunctionName))` 
Or something similar, but definitely doesn't have to be done right now :)

wrwg:
> Overall, I like where this is going. Big fan of the `|T, T, T|T` syntax! I like syntax where the type is matches the expression like this! (I was, and still am, always bothered by the syntax soup around lambdas in Rust)
> 
> I have some concerns about the usability of the macros, just in how it lines up with the rest of the system. Maybe, @wrwg, if you could walk me through you you see this working out that would be helpful :)
> 
> To be clear _none of these are blocking landing_. We don't have to solve all these today. I just want to make sure we are set up for being able to solve these in the future (which might influence this PR)
> 
> ### 1. Type parameters don't normally take references or "tuples"
> At the bytecode level, references cannot instantiate type parameters. And at the bytecode level, these "tuples" don't even exist. So in the source language, every type parameter has to be instantiated with a single non-reference type. This will be unnecessarily restrictive, and kind of annoying for some of these macros. For example I imagine it would be nice for `Vector::reduce` to have the accumulator be a "tuple", e.g.
> 
> ```
> let v: vector<u64> = ...;
> let (num_even, num_odd) = Vector::reduce(
>     v, 
>     (0, 0), 
>     |x, (even, odd)| if (x % 2 == 0) even = even + 1 else odd = odd + 1,
> });
> ```
> 
> But this gets a bit tricky if we just remove the single type constraint from macro type parameters in that we need to be sure it is still applied in the appropriate spots (which I think it should happen automatically, even in cases like `macro fun foo<T>(v: vector<T>, f: ||T)` and `foo!<(u64, bool)>(...)`)
> 
> My bigger concern really is, do you think this will be confusing to programmers? (That the rules for type parameters are different for macro vs non-macro funs)

I think this will be ok if our error messages are good enough. 

I wonder whether we already considered a different (not uncommon) approach to the tuple problem. Let them be just some up to a given N predefined value structs (say N=12). The type and pattern notation is then just syntactic sugar. The return value of a function, if it is a tuple, may then have some optimized compilation to exploit direct VM support.



> 
> ### 2. Substitution and shadowing
> There is a whole can of worms once we try to expand the macro. The two that I'm most concerned about are A. shadowing macro parameters B. "captured" local variables in lambdas
> 
> For (A), I see two possibilities, we could either have a check that says "You cannot shadow or assign macro fun parameters" or we could put `$` in front of macro parameters so they are in a different name space from other variables
> 
> For (B), I think we should probably just not allow capturing of locals in these lambdas in the short term? I think it will just make the macro expansion a lot cleaner and we can always come back to later

I actually think this is  not that problematic, because after all, we do not have Rust macros but just regular higher-order functions, only that we expand them on bytecode level.

A. Where the lambda body is substituted, all locals in the body need to be renamed away from all the locals in the context. On Move level, regular static scoping will apply for the lambda.

B. Capturing is fine, because the lambda is only substituted *downwards* of the invocation point of `m!(|x|e)`, inheriting the context of exactly this expression. The lambda always appears as a parameter. One cannot say `let f = |x|e; m!(f)`. Thus basically, capturing comes for free via substitution (A).

There is a problem C: Borrow errors if a captured value is consumed in a loop defined by the macro. Perhaps at some point we need to have better support for errors stemming from macro body expansion.

> 
> ### 3. Receiver syntax inside lambdas
> I think this is a non-issue as long as we restrict macro lambda usage heavily (which I think we have to anyway), but I'll write it up anyway.
> 
> Normally, when type checking the body of lambdas you have that you don't really want to type check the body until the lambda is "used". This is so you can be more permissive in not requiring annotations. For example,

I think for `fun macro` we should assume they are fully independently and eagerly type checked. Basically just higher order functions which we happen to expand.

We may want to add other kind of macros down the road where this is not the case.

> 
> ```
> let v: vector<MyStruct> = ...;
> let f = |x| x.foo();
> some_cool_macro!(x, f)
> ```
> 
> If you type check `|x| x.foo()` when it is first seen, you have to error since you don't know what `x` is. But if you wait until after it is "used" (like in the call `some_cool_macro`), you can give a better programming experience.
> 
> I think this shouldn't be an issue because we likely won't support anything other than using the lambda in the macro call, e.g. `let v: vector<MyStruct> = ...; some_cool_macro!(x, f)`
> 
> ### 4. Receiver syntax inside macro funs
> Where did we land on type checking the body of `macro fun`s? I remember discussing this particular issue but don't remember where we landed.

Fully type checked, currently.

> 
> It could be annoying if we don't (since you might see an error over and over again like you do in Rust macro expansions) But conversely it would be more expressive if we don't! That way you could write macros such as
> 
> ```
> macro fun call_foo<T>!(v: &vector<T>) {
>     Vector::foreach!(v, |x| x.foo())
> }
> ```
> 
> As I write this out, I think you will want to expand the macro _during_ type checking, which will I think mean you cannot type check the `macro fun` before it is used everywhere.
> 
> But let me know your thoughts
> 
> ### 5. Errors in macro expansion
> I think a managable concern but just something I thought of:
> 
> I think we want to make sure we add a note or label of some sort to any diagnostic that happens inside of the code of an expanded macro. We could maybe accomplish this by adding a node `UnannotatedExp_::ExandedMacro(Box<Exp>, /* inside of */ (ModuleIdent, FunctionName))` Or something similar, but definitely doesn't have to be done right now :)

Absolutely!

language/move-compiler/src/naming/ast.rs
```diff
@@ -140,6 +141,8 @@ pub enum BuiltinTypeName_ {
     Vector,
     // bool
     Bool,
+    // Function (last type arg is result type)
+    Fun,
```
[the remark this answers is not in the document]
wrwg: Yes, we have discussed this already extensively ;) 

To summarize, I disagree on this one. Functions are just a new kind of type constructors. Type constructors are currently represented by TypeName and BuiltinTypeName.

The only difference of function type constructors to existing ones is there contra-variance in the argument types when it comes to type unification. 

It's fairly standard in languages with type unification to treat functions just like other type constructors, conceptually and in the implementation.

wrwg:
@tnowacki @sblackshear @meng-xu-cs @rahxephon89 @junkil-park @runtian-zhou 

Folks, this resurrects the macro PR from April of this year. Macros are really, really important for the Move prover, so I'd like to push this forward before the end of the year.

I think most of the design we discussed back in April is still valid. I would probably suggest to discuss whether we really call those guys macros and instead preserve this name for something more on a lexical level. I personally still like the name 'template' for what is implemented here.

wrwg:
PS. Also want to add that after discussion also with @runtian-zhou I do not see an overlap with the interface proposal #449. Interfaces serve a different equally valid purpose. Macros + Interfaces together make traits unnecessary.

PR #7, opened 2022-04-12
[account-address] Input and output with a 0x for hex output

gregnazario:
### Summary

Accepts 0x as an input, and will print the output as a 0x for AccountAddress.
This removes confusion on whether it's in a different encoding, and keeps
it simple.

### Motivation

Addresses should start with a `0x` as I have been described by Todd how a hex literal must have `0x` in front of it.  Otherwise, we're being inconsistent and having `0x` in code, but no `0x` outside of it for the same representation.

Updated version from: https://github.com/diem/move/pull/199

### DIFF
--- a/language/move-core/types/src/account_address.rs
+++ b/language/move-core/types/src/account_address.rs
@@ -68,7 +68,7 @@ impl AccountAddress {
 
         let hex_len = literal.len() - 2;
 
-        // If the string is too short, pad it
+        // If the string is too short, pad it because it needs to be exactly the right number of bytes
         if hex_len < Self::LENGTH * 2 {
             let mut hex_str = String::with_capacity(Self::LENGTH * 2);
             for _ in 0..Self::LENGTH * 2 - hex_len {
@@ -92,7 +92,7 @@ impl AccountAddress {
     }
 
     pub fn to_hex(&self) -> String {
-        format!("{:x}", self)
+        self.short_str_lossless()
     }
 
     pub fn from_bytes<T: AsRef<[u8]>>(bytes: T) -> Result<Self, AccountAddressParseError> {
@@ -206,7 +206,7 @@ impl From<&AccountAddress> for [u8; AccountAddress::LENGTH] {
 
 impl From<&AccountAddress> for String {
     fn from(addr: &AccountAddress) -> String {
-        ::hex::encode(addr.as_ref())
+        addr.to_hex_literal()
     }
 }
 
@@ -214,7 +214,7 @@ impl TryFrom<String> for AccountAddress {
     type Error = AccountAddressParseError;
 
     fn try_from(s: String) -> Result<AccountAddress, AccountAddressParseError> {
-        Self::from_hex(s)
+        AccountAddress::from_str(&s)
     }
 }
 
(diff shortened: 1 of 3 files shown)


wrwg (changes requested):
Can you please please rebase and push again? The rust toolchain was upgraded and I'm concerned we merge this as this, new linter errors will slip in, as the new toolchain is more  pedantic.

gregnazario:
> Can you please please rebase and push again? The rust toolchain was upgraded and I'm concerned we merge this as this, new linter errors will slip in, as the new toolchain is more pedantic.

@wrwg I've rebased it in

gregnazario:
Turns out I had a test in a rust block in a comment.  I created a new task to track these kinds of issues https://github.com/aptos-labs/aptos-core/issues/503

/land

wrwg:
Adding @tnowacki which I think reviewed this PR in the old repo.

gregnazario:
> I think its totally reasonable to require `0x` if you aren't using the source language, where it is a bit more obvious what is going on (I hope).

Yeah, it's mostly about the output from the Rust attached code e.g. if I print the address, it should print out in a way that I can copy it and put it into move code the same way, so that's why I decided to make this match the hex literal.  Everything should be consistent on the hex output, and then you can simply copy paste from one place to another, without thinking about whether you need to add 0x or not.

> Do you know if this code affects the package system? We had an open issue for that

I'm not aware if it affects the package system.  Is there a test for that?

tnowacki:
> I'm not aware if it affects the package system. Is there a test for that?

Would you mind adding a quick one? :D It should be fairly easy

If you look at `language/tools/move-package/tests/test_sources/parsing/invalid_hex_address_in_subst` as an example, you could just copy this and do `language/tools/move-package/tests/test_sources/parsing/invalid_non_hex_address` with  a `Move.toml` that looks like 

```
[package]
name = "name"
version = "0.1.2"

[addresses]
A = "1234"
```

It doesn't need to be fixed for this PR, just we should maybe test if it is also fixed, would be a fun bonus :)

gregnazario:
> > I'm not aware if it affects the package system. Is there a test for that?
> 
> Would you mind adding a quick one? :D It should be fairly easy
> 
> If you look at `language/tools/move-package/tests/test_sources/parsing/invalid_hex_address_in_subst` as an example, you could just copy this and do `language/tools/move-package/tests/test_sources/parsing/invalid_non_hex_address` with a `Move.toml` that looks like
> 
> ```
> [package]
> name = "name"
> version = "0.1.2"
> 
> [addresses]
> A = "1234"
> ```
> 
> It doesn't need to be fixed for this PR, just we should maybe test if it is also fixed, would be a fun bonus :)

I'm not sure what's supposed to be in the `.exp` file, is it supposed to be the error that's thrown?

tnowacki:
> I'm not sure what's supposed to be in the .exp file, is it supposed to be the error that's thrown?

If your change affects the package system, there will be an error. If it does not, there shouldn't be one. (but we probably want this to be an error)
You can double check this by running the test on main and again on your branch.

You can run the test by `cargo test -p move-package`.
You can create the `.exp` file once you think the output is correct with `UB=1  cargo test -p move-package` which will update the `.exp` files for all tests.

gregnazario:
@tnowacki I added the test as expected :)

davidiw (approved):
let's get the linter issues fixed and get this landed.

PR #8, opened 2022-04-12
[table] adjustments

msmouse:
## Motivation

1. manage the size from Move, instead of the extention. This way, the storage doesn't need to store a metadata item for the table.
2. fix issues with items created and removed in the same session before hitting storage.

y

## Test Plan

unitests in Move

### DIFF
--- a/language/extensions/move-table-extension/sources/Table.move
+++ b/language/extensions/move-table-extension/sources/Table.move
@@ -2,7 +2,11 @@
 module Extensions::Table {
     use Std::Errors;
 
-    const ENOT_EMPTY: u64 = 0;
+    // native code raises this with Errors::invalid_arguments()
+    const EALREADY_EXISTS: u64 = 100;
+    // native code raises this with Errors::invalid_arguments()
+    const ENOT_FOUND: u64 = 101;
+    const ENOT_EMPTY: u64 = 102;
 
     /// Type of tables
     struct Table<phantom K, phantom V> has store {
--- a/language/extensions/move-table-extension/sources/Table.move
+++ b/language/extensions/move-table-extension/sources/Table.move
@@ -1,17 +1,26 @@
 /// Type of large-scale storage tables.
 module Extensions::Table {
+    use Std::Errors;
+
+    const ENOT_EMPTY: u64 = 0;
+
     /// Type of tables
     struct Table<phantom K, phantom V> has store {
-        handle: u128
+        handle: u128,
+        length: u64,
     }
 
     /// Create a new Table.
     public fun new<K, V: store>(): Table<K, V> {
-        Table{handle: new_table_handle()}
+        Table{
+            handle: new_table_handle(),
+            length: 0,
+        }
     }
 
     /// Destroy a table. The table must be empty to succeed.
     public fun destroy_empty<K, V>(table: Table<K, V>) {
+        assert!(table.length == 0, Errors::invalid_state(ENOT_EMPTY));
         destroy_empty_box<K, V, Box<V>>(&table);
         drop_unchecked_box<K, V, Box<V>>(table)
     }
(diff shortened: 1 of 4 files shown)


wrwg:
Can you please please rebase and push again? The rust toolchain was upgraded and I'm concerned we merge this as this, new linter errors will slip in, as the new toolchain is more  pedantic.

msmouse:
@wrwg  your table diff did that, right? This is already on top of that. (And looks like it's still the head of main)

language/extensions/move-table-extension/sources/Table.move
```diff
@@ -1,17 +1,26 @@
...
-        Table{handle: new_table_handle()}
+        Table{
+            handle: new_table_handle(),
+            length: 0,
+        }
     }
 
     /// Destroy a table. The table must be empty to succeed.
     public fun destroy_empty<K, V>(table: Table<K, V>) {
+        assert!(table.length == 0, Errors::invalid_state(ENOT_EMPTY));
```
wrwg, line 23: I think we still may want to return a unique abortion code here, as it is/was defined in the Rust code. You could have a native function here to discover this abortion code.

language/extensions/move-table-extension/tests/TableTests.move
```diff
@@ -41,7 +41,7 @@ module Extensions::TableTests {
     }
 
     #[test]
-    #[expected_failure(abort_code = 51155)]
+    #[expected_failure(abort_code = 1)]
```
wrwg, line 44: This kind of shows why having a user level abortion code is not the best idea. This clashes with the assert code used below.

wrwg (commented):
Great idea to move this to the Table struct. Some nit about abort codes.

language/extensions/move-table-extension/tests/TableTests.move
msmouse: I was ready to make a native function to expose a error code.. but it felt not quiet right..

To be fair a data structure implemented in pure Move will need to use an error like this, right?
Also to be fair `Errors::invalid_state(1)` is `257`,  `Errors::invalid_state(10)` is `2566`, and `Errors::already_published(1)` is `262` .. `Error::invalid_state(0)` is just an unfortunate combination..


Maybe it's more of the framework /stdlib's responsibility to provide a better `Error` facility? (maybe just offset everything by 10000)
Shall we just make ENOT_EMPTY more unique in this case?

wrwg: Good point. I forgot about the practice to use categories (in the first byte of the code) to provide some kind of canonical error code, and distinguish it from the reason (in your case `0`) -- a thing I pushed for myself a while ago. I think most of our native function implementations do the same and do not respect the category/reason convention.

Categories are good, but what bothers me with our current setup is that reasons are not unique across modules. That is why I played with the crypto approach: actually one can generate a (sufficiently) unique code by sha256 of the full module name and a local counter, without the need of some global error registry or convention. But perhaps we should roll out something like this systematically across all code (native and not) only once we had more sorough design discussion about this.

Can I then ask to implement for now the category/reason thing consistently for all table operations, whether native or not? That is we should also change the other abort codes in `lib.rs` to follow the same convention like `Errors.move`.

If this is PR is blocking more important work, we can also do this after it is merged, so I leave you an approve to move forward either way you prefer.

wrwg (approved):

language/extensions/move-table-extension/tests/TableTests.move
msmouse: I pushed another commit. Lmk if we want to list all error codes on both sides although not used on either side (as done in the diff), or if we want to be lean.

wrwg:
The hardtests should be fixed, please rebase and try again.

wrwg:
> /canary

Bors not longer in place, so these are now just comments.

msmouse:
I'm gonna land for now. Lmk if we want adjustments in the error code implementation, I'll follow up later.
/land

PR #14, opened 2022-04-13
[evm] initial implementation of the Move plugin for hardhat

vgao1996:
## Overview
This is the initial implementation of the Move plugin for hardhat. This plugin extends hardhat's `compile` task, making it possible to compile contracts written in Move and generate the require artifacts for deployment and testing. 

## How to use the plugin?
Please refer to `hardhat-move/README.md` for a detailed tutorial.

## Testing
This PR also updates the CI test for hardhat so that it uses this plugin, offering coverage for both. As a side effect, the old scripts were removed as they are no longer needed.

## Known Issues
There are still some issues about the error handling path:
1. Chained errors are not printed out in a nice way
2. There are still a handful of places where exceptions could be thrown, which need to be switched to Result-based error handling

These are more about the obscure cases and should not be easily triggerable by normal users. I'm planning to address them in a followup PR.

I'd also love to have a little bit more documentation for the implementation of the plugin, but again, I'll leave it for a separate PR.


wrwg (approved):
Wow! The typescript looks like fun.

PR #22, opened 2022-04-14
[ci] Run CI on merge commit instead of head commit

wrwg:
Running on head commit increases the chance that CI run passes standalone but fails once the changes are merged.

We had already the first case where head commit lead to a broken test state after merge (fixed in #21). 

Also removes checking out all branches because that seems to be not needed in the Move repo.

### DIFF
--- a/.github/workflows/ci-test.yml
+++ b/.github/workflows/ci-test.yml
@@ -15,11 +15,6 @@ jobs:
       test-rust: ${{ steps.rust-changes.outputs.changes-found }}
     steps:
       - uses: actions/checkout@v2.4.0
-        with:
-          # This ensures that the tip of the PR is checked out instead of the merge between the base ref and the tip
-          # On `push` this value will be empty and will "do-the-right-thing"
-          ref: ${{ github.event.pull_request.head.sha }}
-          fetch-depth: 0 #get all the history!!!
       - name: Git Hooks and Checks
         run: ./scripts/git-checks.sh
       - id: changes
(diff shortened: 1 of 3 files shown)


vgao1996:
One concern I have is that might make it harder to work on bigger PRs by stripping away the ability to test them in CI without forcing them to be rebased. Well, I guess if problems occur due to running on merged, then eventually you still need to fix them, but I feel like sometimes an author may want to make sure the PR works by itself, before rebasing and dealing with the new issues. I don't have strong opinions on this though...

> Running on head commit increases the chance that CI run passes standalone but fails once the changes are merged.

This wasn't a problem when we had bors, and probably won't be one once we gain access to merge queues, so I'm wondering if we could still apply the change you are proposing, but making it a temporary one?

wrwg:
> One concern I have is that this might make it harder to work on bigger PRs by stripping away the ability to test them in CI without forcing them to be rebased. Well, I guess if problems occur due to running on merged, then eventually you still need to fix them, but I feel like sometimes an author may want to make sure the PR works by itself, before rebasing and dealing with the new issues. I don't have strong opinions on this though...
> 
> > Running on head commit increases the chance that CI run passes standalone but fails once the changes are merged.
> 
> This wasn't a problem when we had bors, and probably won't be one once we gain access to merge queues, so I'm wondering if we could still apply the change you are proposing, but making it a temporary one?

I you want to test a PR at tip, why not do this locally. You can just replicate the commands CI does in the repo root:

```
cargo x lint
cargo xclippy --workspace --all-targets
cargo xfmt --check
cargo test --workspace
```

The flows have become sufficiently less complex in the new repo. (BTW on my private 16 core Ryzen it also finishes in a few minutes).

The current setting has already created headache, at least for the prover team. Een with bors, you have those PRs which unexpectedly fail on /land. It is unnatural and inconsistent with the "integration" in the CI. I wish we could get rid of it ASAP and not wait for queues, but I'm open also to other reasonings.

sblackshear (approved):

wrwg:
Post land note: inspired by the PR discussion with @vgao1996 , this PR now also added `./scripts/check_pr.sh` which allows to run the various CI steps locally. This was ported from the prover and adopted to the whole repo.

PR #25, opened 2022-04-14
Point to a shared Move Discord

davidiw:


README.md
```diff
@@ -11,7 +11,7 @@ This repository is the official home of the Move virtual machine, bytecode verif
 
 ## Community
 
-* Join us on the [Move Discord](https://discord.gg/M95qX3KnG8).
+* Join us on the [Move Discord](https://discord.gg/vTepKPdF).
```
wrwg, line 14: Is that the public or contributor-only channel? (If such a thing exists?)

davidiw: This will invite them to the server. The server contains channels for both.

wrwg (approved):

Issue #13, opened 2022-04-13
[Feature Request]  Read the script signature in session 


areshand:
# 🚀 Feature Request

## Motivation

To correctly execute a script, we need to have the signature of the script to (1) validate the script has the correct signature and (2) to determine how to add a signer into the args.

Now, there is already support for the script_function (some previous discussion happens here https://github.com/diem/move/issues/209). However, we still need a "load_script" function to provide similar functionalities.

areshand:
Hi @tnowacki, didn't find there is support for load_script. do you plan to add util function to view the signature of script? cc: @zekun000  @msmouse

tnowacki:
Ah yeah, can do. Should be easy to add

areshand:
Hi @tnowacki when would you have time to make this change?

tnowacki:
PR up

PR #35, opened 2022-04-15
[move-vm] Add load_script to session API

tnowacki:
## Motivation

- allow for easy adapter rules, as was similarly intended with the load_function API
- Fix some vector operations

## Test Plan

- new test


tnowacki:
Fixes #13

davidiw (approved):
Thanks for moving fast on this.

PR #38, opened 2022-04-16
[move-vm] Removed most RCs from runtime

tnowacki:
Migrating from other repo

## Motivation

- RCs are no longer necessary due to Move's reference safety checks
- Some RCs are still needed to update clean/dirty status or hold onto external memory safely
- I feel this greatly simplifies the layout of values in value_impl.rs
- It does however make `write_ref` unsafe, as it could lead to memory issues if the reference safety checks are wrong or not run
- I do not have a reference heavy bench mark to test if this makes any perf difference. My guess is it won't, but might help with memory usage

## Test Plan

- ran tests
- I'm a bit nervous that our tests might not be exhaustive enough though....

### DIFF
--- a/language/extensions/async/move-async-vm/src/async_vm.rs
+++ b/language/extensions/async/move-async-vm/src/async_vm.rs
@@ -15,6 +15,7 @@ use move_core_types::{
     identifier::Identifier,
     language_storage::{ModuleId, StructTag, TypeTag},
     resolver::MoveResolver,
+    value::MoveTypeLayout,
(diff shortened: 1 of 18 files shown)


wrwg:
I ported one of my comments on your original PR to issue  #40 . The other ones can be still found in the older PR. Basically, those comment were around my concern using unsafe APIs throughout our code (in the VM and native functions). Perhaps this can be avoided, and that is what #40 explores.

Issue #46, opened 2022-04-17
[Bug] move-dissassembler panicked over out-of-bound error


whst:
# 🐛 Bug
If a generic function contains a generic borrow of a struct that takes N type parameters Ty1, Ty2, ..., TyN, then all the type parameters have to appear as the first N type parameters of the function. Otherwise, the move-dissasembler cannot generate the string represenation of the borrowed type in `struct_type_info()` function in disaseembler.rs.

## To reproduce
Here is an MWE that demonstrates the problem.

After compilation, the `FooAX` module can be dissembled by move-dissasembler.
```
address 0x1 {
module FooAX {
  struct C<T: copy + store + drop> has key { f: T }
  fun foo<A: copy + store + drop, X>(x: X): X acquires C { 
      let _ = borrow_global<C<A>>(@0x1); 
      x
  }
}
}
```

But move-disassembler failed to show the bytecode of the almost identical module `FooXA`:
```
address 0x1 {
module FooXA {
  struct C<T: copy + store + drop> has key { f: T }
  fun foo<X, A: copy + store + drop>(x: X): X acquires C {  // The order changes
      let _ = borrow_global<C<A>>(@0x1); 
      x
  }
}
}
```
The error info is:
```
thread 'main' panicked at 'Unable to dissassemble: Type parameter index 1 out of bounds while disassembling type signature', language/tools/move-disassembler/src/main.rs:128:58
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

## Expected Behavior

move-dissembler should be able to handle both.

## Additional context
After some inspection of the source code, I found that when `disassemble_instruction()` sees a [BorrowGlobalGeneric](https://github.com/move-language/move/blob/6a8d297f6df87e8f1efa01a286c076abe07b6ec9/language/tools/move-disassembler/src/disassembler.rs#L651), it uses `struct_type_info()`, where `signature` has index not within the length of `struct_source_map.type_parameters`, which causes the error.

PR #49, opened 2022-04-18
Fix the bug that move-disassemble using wrong type_param_context

whst:
## Motivation
Fixed the bug mentioned in #46.

Basically, `struct_type_info()` shouldn't use the source map of the struct. Instead, it should use the source map of the enclosing function.
For example,
```rust
address 0x1 {
module Foo {
  struct S<A: copy + store + drop, B: copy + store + drop> has key, drop { fa: A, fb: B }

  fun foo<X: copy+drop+store, Y: copy+drop+store>(x: X, y: Y): S<Y, X> {
    S { fa: y, fb: x }  // PackGeneric instruction has signature [1,0]
  }
}
}
```
Here the [1 and 0] actually refer to [Y and X], not [B and A].
Current implementation will generate `PackGeneric[0](S<B, A>)`, which is wrong.
This PR fix this and now the disassembled instruction will be `PackGeneric[0](S<Y, X>)`.

This is also the reason of the out-of-bound error mentioned in #46, where the enclosing function contains more generic paramters than the struct, and some instructions' signature contains index greater than the length of the wrongly used type_parameters of struct's source_map, causing the panick.

### DIFF
--- a/language/tools/move-disassembler/src/disassembler.rs
+++ b/language/tools/move-disassembler/src/disassembler.rs
@@ -265,18 +265,13 @@ impl<'a> Disassembler<'a> {
         &self,
         struct_idx: StructDefinitionIndex,
         signature: &Signature,
+        type_param_context: &[SourceName],
     ) -> Result<(String, String)> {
         let struct_definition = self.get_struct_def(struct_idx)?;
-        let struct_source_map = self
-            .source_mapper
-            .source_map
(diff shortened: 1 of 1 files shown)


tnowacki (approved):

tnowacki:
Simple fix, thanks!
And I wouldn't be surprised if there were some more spots with similar bugs in this code, so be on the look out!

PR #51, opened 2022-04-18
[chore] update PR template to match new organization

damirka:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -18,11 +18,4 @@ The project is undergoing daily changes. Pull Requests will be reviewed and resp
 
 ## Related PRs
```
wrwg, line 19: The related PR section is always never used. I think we should drop it.

wrwg (approved):
This PR itself uses the empty template ;)

.github/PULL_REQUEST_TEMPLATE.md
damirka: Agreed. Will update in a sec.

Issue #53, opened 2022-04-18
[account address] Cleanup multiple to string & from string implementations


gregnazario:
I made the previous change to make 0x more prevalent in the account address outputs, but it's clear to me that there are too many ways to display the account addresses and it adds a lot of confusion.

I'm going to clean it up so that there are less implementations and that it's straightforward at that point.

gregnazario:
cc @areshand I'm working on this right now

gregnazario:
Seems to me there are currently 3 use cases:

for 0x00000000abcd

(to & from hex_literal and also display/debug) To & From Hex -> 0xabcd
(short_str_lossless) To short str as identifier -> abcd
(a currently non existent function) And for some time fuzzy matching back to address without 0x -> abcd

One I want to remove which is confusing is (to_hex, from_hex), which doesn't take a 0x, and has to be exactly the right length number of bytes

@tnowacki would you agree with me?  So, I can move forward on this?

Also, what's your opinion on lowercase vs uppercase, since `short_str_lossless` is lowercase and `display` is uppercase

tnowacki:
Thanks for looking at this! It has always annoyed me, which is partially why the source language's parsing/rendering is its own thing :P 

I think I agree with the setup you have described with to/from hex and short_str_lossless and maybe this "fuzzy" matching back. But it will be easier for me to know for sure when I see the code.

And I would love to do uppercase by default if we don't already! I think both @sblackshear  and myself tried to do that on separate occasions but got frustrated with the amount of breakages

PR #54, opened 2022-04-19
[account-address] Make display and hex outputs consistent

gregnazario:
## Motivation

Now, all default outputs should be `0xCAPS`.  Additionally, there is now a `from_hex_fuzzy()` which handles both `0x` and non-`0x` inputs for the time being.

This is built as follows:
1. `from_hex` -> Only hex characters, exact length
2. `from_hex_literal` -> Only 0x with hex characters, with length extension
3. `from_hex_fuzzy` -> Match either condition
4. `short_str_lossless` -> With 0x, but drop leading 0s
5. Drop the `to_hex` and `to_hex_literal` since we can now use `Display` or `Debug` using the proper formatter.  Additionally to remove any confusion around it.  All of are now consistent as upper case.

https://github.com/move-language/move/issues/53

### DIFF
--- a/language/tools/move-package/src/source_package/manifest_parser.rs
+++ b/language/tools/move-package/src/source_package/manifest_parser.rs
@@ -3,7 +3,7 @@
 
 use crate::{source_package::parsed_manifest as PM, Architecture};
 use anyhow::{bail, format_err, Context, Result};
-use move_core_types::account_address::{AccountAddress, AccountAddressParseError};
+use move_core_types::account_address::AccountAddress;
 use move_symbol_pool::symbol::Symbol;
 use std::{
     collections::{BTreeMap, BTreeSet},
@@ -213,10 +213,9 @@ pub fn parse_addresses(tval: TV) -> Result<PM::AddressDeclarations> {
                         } else if addresses
                             .insert(
                                 ident,
-                                Some(parse_address_literal(entry_str).context(format!(
-                                    "Invalid address '{}' encountered.",
-                                    entry_str
-                                ))?),
+                                Some(AccountAddress::from_hex_fuzzy(entry_str).context(
+                                    format!("Invalid address '{}' encountered.", entry_str)
+                                )?),
                             )
                             .is_some()
                         {
@@ -254,7 +253,7 @@ pub fn parse_dev_addresses(tval: TV) -> Result<PM::DevAddressDeclarations> {
                         } else if addresses
                             .insert(
                                 ident,
-                                parse_address_literal(entry_str).context(format!(
+                                AccountAddress::from_hex_fuzzy(entry_str).context(format!(
                                     "Invalid address '{}' encountered.",
                                     entry_str
                                 ))?,
@@ -279,14 +278,6 @@ pub fn parse_dev_addresses(tval: TV) -> Result<PM::DevAddressDeclarations> {
             x.type_str()
         ),
     }
-}
-
-// Safely parses address for both the 0x and non prefixed hex format.
-fn parse_address_literal(address_str: &str) -> Result<AccountAddress, AccountAddressParseError> {
-    if !address_str.starts_with("0x") {
-        return AccountAddress::from_hex(address_str);
-    }
-    AccountAddress::from_hex_literal(address_str)
 }
 
 fn parse_dependency(tval: TV) -> Result<PM::Dependency> {
--- a/language/evm/move-to-yul/src/generator.rs
+++ b/language/evm/move-to-yul/src/generator.rs
@@ -278,7 +278,7 @@ impl Generator {
                 emit!(ctx.writer, "let $arg{} := ", idx);
                 match arg {
                     MoveValue::Address(addr) => {
-                        emitln!(ctx.writer, "{}", addr.to_hex_literal());
+                        emitln!(ctx.writer, "{}", addr);
                     }
                     _ => unreachable!(
                         "only address literals are allowed as test arguments currently"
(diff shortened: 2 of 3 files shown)


gregnazario:
@tnowacki lmk how this looks

tnowacki:
Overall it looks fine to me! but I imagine there might be some changes if you have tests to fix (they are all red, not sure if related, looks like it might not be). Just ping me again when things are passing :)

gregnazario:
@tnowacki it looks like there is a compilation issue in addition to the tests failing, though the compilation issue doesn't look like it's due to me

language/move-ir-compiler/move-ir-to-bytecode/syntax/src/syntax.rs
```diff
@@ -187,7 +187,7 @@ fn parse_account_address(
             message: "expected Tok::AccountAddressValue".to_string(),
         });
     }
-    let addr = AccountAddress::from_hex_literal(tokens.content())
+    let addr = AccountAddress::from_hex_fuzzy(tokens.content())
```
tnowacki, line 190: This should always expect the `0x`

language/testing-infra/transactional-test-runner/src/tasks.rs
```diff
@@ -23,21 +23,9 @@ pub enum RawAddress {
...
-        )
-    })?;
-
-    Ok(AccountAddress::new(array))
-}
-
 impl RawAddress {
     pub fn parse(s: &str) -> Result<Self> {
-        if let Ok(addr) = parse_address_literal(s) {
+        if let Ok(addr) = AccountAddress::from_hex_fuzzy(s) {
```
tnowacki, line 28: This isn't the same behavior. WIthout the `0x` parses from a decimal value, and that should be the behavior for this, and all CLI tools, to match the source language

language/tools/move-cli/src/sandbox/commands/run.rs
```diff
@@ -66,7 +66,7 @@ move run` must be applied to a module inside `storage/`",
 
     let signer_addresses = signers
         .iter()
-        .map(|s| AccountAddress::from_hex_literal(s))
+        .map(|s| AccountAddress::from_hex_fuzzy(s))
```
tnowacki, line 69: maybe we should use the source language parsing here?

tnowacki: And I was surprised to not a see a change for other arg parsing. Have you looked at that?

language/tools/move-package/src/source_package/manifest_parser.rs
```diff
@@ -213,7 +213,10 @@ pub fn parse_addresses(tval: TV) -> Result<PM::AddressDeclarations> {
                         } else if addresses
                             .insert(
                                 ident,
-                                Some(parse_address_literal(entry_str).context("Invalid address")?),
+                                Some(
+                                    AccountAddress::from_hex_fuzzy(entry_str)
```
tnowacki, line 217: I think these should not be fuzzy. We should force 0x for package addresses

tnowacki: Didn't we add a test for that on the last PR?

gregnazario: The original one `parse_address_literal` was fuzzy, so I'm unsure there.  I can switch it back.

language/tools/move-cli/src/sandbox/commands/run.rs
gregnazario: I'm trying to scope down my stuff to AccountAddresses for now, we can come back and do other args later.

language/testing-infra/transactional-test-runner/src/tasks.rs
gregnazario: `parse_address_literal` was already doing fuzzy matching.  So, if that's the case I'm not making this any more wrong, but it was wrong in the first place

tnowacki: This the wrong sort of fuzzy parsing though. It should be following the source languages convention: `0x` is hex, otherwise decimal

tnowacki: But then again.... this is mostly for identifiers, so maybe we should just force the 0x

Should probably reserve the dec/hex parsing for args

language/tools/move-package/src/source_package/manifest_parser.rs
tnowacki: Let's make sure we add back the dropped in tests from #59, those should test this change here one way or another

mkurnikov:
Does this fix https://github.com/diem/move/issues/141 ?

tnowacki:
> Does this fix [diem/move#141](https://github.com/diem/move/issues/141) ?

It should at least make it easier to fix :P, unsure right now if it fixes it or not

tnowacki (changes requested):
Let's switch the calls in `manifest_parser` to `from_hex_literal`. Requesting changes to make sure we get those tests back from the revert. 

Thanks for all of this! It should have been done long ago

gregnazario:
> Let's switch the calls in `manifest_parser` to `from_hex_literal`. Requesting changes to make sure we get those tests back from the revert.
> 
> Thanks for all of this! It should have been done long ago

Sorry, been very distracted on other things, I'll probably come back on a pass through this in a week or two.

tnowacki:
@gregnazario, ping me when you want another review! (looks like you are still working through some changes)

PR #55, opened 2022-04-19
[repo] Remove workspace hack.

wrwg:
Workspace hack is a [build optimizer](https://docs.rs/cargo-hakari/latest/cargo_hakari/about/index.html) which avoids building different feature combinations for different build commands in one workspace. As the link shows, it has advantages for certain build scenarios.

However, it also has disadvantages when building crosscuts through our repo. For example, even if one doesn't ever need the EVM feature of move-cli, workspace-hack will trigger to build it anyway.

This PR removes workspace-hack. This should lead to faster build time if one does repeated builds of one particular crate (the usual dev scenario), and *may* result in slower builds if one often switches the cargo build target (more unusual dev scenario). For building the full workspace, no difference in timing is expected, as with one build command, cargo always builds the sum of all features. This has been verified on a Ryzen 5950x (16 core):

```
With WH:
real    1m16.463s
user    11m14.493s
sys     1m8.965s

Without WH:
real    1m12.165s
user    11m8.295s
sys     1m10.649s
```


damirka (approved):
LGTM!

wrwg:
Update: to make this work, I needed to remove some code from `devtools/x`. There was a hardwired dependency from the hakari package and config in x.toml which only works if there is in fact workspace-hack package. The code around this dep is now removed.

wrwg:
This is now ready to merge. If they are no further concerns I go ahead based on @damirka's approval.

PR #57, opened 2022-04-19
[evm] make evm features non-default

vgao1996:
This makes the EVM features non-default so that one doesn't have to install the ethereum toolchain if one doesn't need it. A few CI tests are added to ensure these features don'e get broken.

This also renames the `evm-arch` feature to `evm-backend`, just to make it consistent with what we use in other crates.


damirka (approved):
Thanks for doing it!

PR #59, opened 2022-04-20
Revert "[account-address] Input and output with a 0x for hex output (…

gregnazario:
…#7)"

This reverts commit ba7c06401ea94a2f712962e00dcf27aebc02cfc4.

This change causes issues with module publishing and transactions, as it doesn't properly convert all hex versions.  I'll fix it in this PR https://github.com/move-language/move/pull/54 but for now I need to revert it

@tnowacki

### DIFF
--- a/language/move-core/types/src/account_address.rs
+++ b/language/move-core/types/src/account_address.rs
@@ -68,7 +68,7 @@ impl AccountAddress {
 
         let hex_len = literal.len() - 2;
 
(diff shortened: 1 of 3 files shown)


gregnazario:
@tnowacki I'd like to push this through, as the previous PR needs more testing, and has caused other issues as a result.  I'll come back to fix all of them together

tnowacki (approved):

PR #60, opened 2022-04-20
[evm][hardhat-move] print out result as soon as each build task finishes

vgao1996:
Right now, when you compile Move contracts in hardhat, no results are shown until all build tasks have finished, which may confuse the user. This PR changes it so that the result is printed out once each individual build task finishes, notifying the user of the progress.


junkil-park (approved):
LGTM. Thanks!

PR #74, opened 2022-04-24
[type] Add alias to StructTag for compatibility with old serialized json

jolestar:
## Motivation

Add an alias to StructTag for compatibility with old serialized json.

When the TypeTag add serde(rename = "u8"), this break the compatibility with old version, so add a alias = "U8" for compatibility.

I try to fix this in https://github.com/diem/move/pull/106, but I find it fixed by https://github.com/diem/move/pull/127, so I closed my PR.

```
 #[serde(rename(serialize = "u8", deserialize = "U8"))]
    U8,
```

But I find it was changed back to lowercase in https://github.com/diem/move/pull/153.

I think the right approach to keep compatibility is to add the `alias` other than adding the `deserialize`.

@areshand please help to review this.

(Write your answer here.)
Yes

## Test Plan

Add a unit test.

### DIFF
--- a/language/move-core/types/src/language_storage.rs
+++ b/language/move-core/types/src/language_storage.rs
@@ -18,21 +18,22 @@ pub const CORE_CODE_ADDRESS: AccountAddress = AccountAddress::ONE;
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Hash, Eq, Clone, PartialOrd, Ord)]
 pub enum TypeTag {
-    #[serde(rename(serialize = "bool", deserialize = "bool"))]
+    // alias for compatibility with old json serialized data.
+    #[serde(rename = "bool", alias = "Bool")]
     Bool,
-    #[serde(rename(serialize = "u8", deserialize = "u8"))]
+    #[serde(rename = "u8", alias = "U8")]
     U8,
-    #[serde(rename(serialize = "u64", deserialize = "u64"))]
+    #[serde(rename = "u64", alias = "U64")]
     U64,
-    #[serde(rename(serialize = "u128", deserialize = "u128"))]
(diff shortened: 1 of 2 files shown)


language/move-core/types/src/language_storage.rs
```diff
@@ -41,8 +42,8 @@ pub struct StructTag {
     pub address: AccountAddress,
     pub module: Identifier,
     pub name: Identifier,
-    // TODO: rename to "type_args" (or better "ty_args"?)
-    #[serde(rename(serialize = "type_args", deserialize = "type_args"))]
+    // alias for compatibility with old json serialized data.
```
tnowacki, line 45: Can you add back the TODO?

tnowacki (commented):
I don't really know why this was last changed so, 
@areshand and @sblackshear from  https://github.com/diem/move/pull/153

language/move-core/types/src/language_storage.rs
areshand: The reason this changed was we use **serde-reflection** to generate rust code from the YAML file generated from **serde**. If the YAML file use lowercaswe, serde-reflection will fail since the lowcase is reserved for keywords. We previously though we could just alias or deserialized to serialize to lowercase while deserialized back to uppercase. But it doesn't work as expected since serde-reflection doesn't  know how serde want to deserialize. We just endup hack our serde-reflection to make the rust generation work. 

In terms of @jolestar's change, It lgtm and it shouldn't break our serde-reflection as well since the generated YAML file doesn't change.

jolestar: @tnowacki Add the TODO back. I mistakenly thought adding `serde(rename)` finished the renaming.

tnowacki (approved):

PR #79, opened 2022-04-25
[chore] Replace mirai annotations with normal asserts

tnowacki:
- mirai is not being run. Replaced the annotations with the corresponding assert/debug_assert
- The mirai-only annotations are now debug_asserts, because why not

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1832,12 +1832,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "mirai-annotations"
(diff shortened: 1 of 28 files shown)


bmwill (approved):

bmwill:
I guess mirai is now dead?

tnowacki:
> I guess mirai is now dead?

I think it is still being worked on? But it hasn't been in usage in the Move repo for a long time

PR #83, opened 2022-04-27
Fix error messages mangled by unicode

cowlicks:
This fixes a small bug I found while looking looking for a "[trojan source](https://trojansource.codes/trojan-source.pdf)" using bidirectional override unicode codepoints. Fortunately, Move is not vulnerable because the it's parser only allows printable ASCII and newlines.

The bug occurs when some code contains the codepoint U+202E - right to left override.
The error does not escape the codepoint,  all the text following the error message is reversed when rendered to a terminal.

So this code:
```move
script {
    use Std::Debug;
    // this is actually:
    // fun hello{U+202E}world(signer:account) {
    fun hello‮world(account:signer) {
        Debug::print(&account)
    }
}
```

When run gives:
```
$ move sandbox run sources/debug_script.move --signers 0xf -v

error[E01001]: invalid character
  ┌─ ./sources/debug_script.move:3:14
  │
5 │     fun hello‮world(account:signer) {
  │               Invalid character '‮' found when reading file. Only ASCII printable characters, tabs (\t), and line endings (\n) are permitted.

```

With this PR we get:
```
$ ../move/target/debug/move sandbox run sources/debug_script.move --signers 0xf -v

error[E01001]: invalid character
  ┌─ ./sources/debug_script.move:3:14
  │
5 │     fun hello‮world(account:signer) {
  │               Invalid character '\u{202e}' found when reading file. Only ASCII printable characters, tabs (\t), and line endings (\n) are permitted.
```

I also noticed what looked like a similiar bug in `language/move-ir-compiler/move-ir-to-bytecode/src/parser.rs` so I added a fix and test for that too. But it wasn't immediately obvious how to invoke the move-ir-to-bytecode crate to reproduce the issue.

Note the source code shown in the error is also rendered in a mangled way, but I wasn't sure if that counts as a bug or not.

yes

## Test Plan

Can someone explain how I could reproduce the issue in the `move-ir-to-bytecode` parser? Otherwise the unit tests added cover the changes.


language/move-compiler/src/parser/comments.rs
```diff
@@ -34,3 +34,20 @@ pub fn verify_string(file_hash: FileHash, string: &str) -> Result<(), Diagnostic
...
+        let result = super::verify_string(
+            super::FileHash::new("unused"),
+            checkmark,
+            );
+        let diagnostics = result.unwrap_err();
+        let err_msg = format!("{:?}", diagnostics);
+        assert!(err_msg.contains(&checkmark.escape_default().to_string()));
+        assert!(!err_msg.contains(&checkmark));
+    }
+}
```
tnowacki, line 53: Could you add a test for this instead in `language/move-compiler/tests/move_check/parser/`

language/move-ir-compiler/move-ir-to-bytecode/src/parser.rs
```diff
@@ -105,4 +105,14 @@ mod tests {
             good_chars.pop();
         }
     }
+
+    #[test]
```
tnowacki, line 109: This one is fine, I don't care about the ir parser as much :P

tnowacki (changes requested):
Thanks for the fix! Just a small change to make sure we have end-to-end testing for how this is handled

language/move-compiler/src/parser/comments.rs
cowlicks: Test moved, and rebased into this commit.

cowlicks:
Updated the test and rebased into the move-compiler commit as per [CONTRIBUTING.md#bisect-able-history](https://github.com/move-language/move/blob/main/CONTRIBUTING.md#bisect-able-history)

language/move-compiler/tests/move_check/parser/invalid_character_comment.exp
```diff
@@ -1,6 +1,6 @@
-error[E01001]: invalid character
+error[E01001]:invalid character
   ┌─ tests/move_check/parser/invalid_character_comment.move:3:44
   │
 3 │ // Non-ASCII characters in comments (e.g., ф) are also not allowed.
-  │                                            ^ Invalid character 'ф' found when reading file. Only ASCII printable characters, tabs (\t), and line endings (\n) are permitted.
+  │                                            ^ Invalid character '\u{444}' found when reading file. Only ASCII printable characters, tabs (\t), and line endings (\n) are permitted.
```
tnowacki, line 5: I now have a much better understanding of whats going on here, thanks for the test.

Would it be possible to display the original character if it is printable? Not sure if there is an easy way to check that or not.

If not, this at least fixes the bug, so we can land as is

language/move-ir-compiler/move-ir-to-bytecode/src/parser.rs
tnowacki, line 109: Gah! I'm super sorry. I was reminded that we actually have these `.exp` tests for the IR now too. We should make this one to follow suit. 

It should go in 
`language/move-ir-compiler/transactional-tests/tests/parsing`

wrwg:
Is this still active?

PR #85, opened 2022-04-27
[move-package] Run compiler once

tnowacki:
- Removed compiling each package separately. It is now run once for the root
- Caching is a bit too hard, and likely not worth it. It can be added back later.

## Motivation

- The package system was rather difficult to work with for transitive dependencies, specifically if you want to extract information from the compiler with a custom driver
- This change means ALL packages (root + transitive deps) are compiled at once, with no caching.
- The caching could be brought back, but right now my guess is that the perf improvements would be marginal.

## Test Plan

- Updated tests (though I didn't actually see any tests that tested the caching behavior? I think it was broken anyway...)


language/tools/move-cli/src/package/cli.rs
```diff
@@ -496,57 +489,38 @@ pub fn run_move_unit_tests(
                 .collect::<HashMap<_, _>>()
         })
         .collect();
+    let root_package = Some(resolution_graph.root_package.package.name);
     let build_plan = BuildPlan::create(resolution_graph)?;
     // Compile the package now. We need to treat the root package differently since we need to
```
awelc, line 494: I think that this comment is largely out of date since the code does not specialize towards the root package anymore.

language/tools/move-package/src/compilation/compiled_package.rs
```diff
@@ -463,142 +464,72 @@ impl CompiledPackage {
...
-            dep_package.gather_module_resolution_metadata(
-                &resolved_package.renaming,
-                &mut module_resolution_metadata,
-            )
-        }
-        // gather from compiled units (which do not include deps)
-        for (mod_id, name) in resolutions.into_iter().flatten() {
-            let _old_value = module_resolution_metadata.insert(mod_id, name);
-            debug_assert!(_old_value.is_none());
+        let paths = {
```
awelc, line 509: Am I reading it right that we will now have all ASTs for all (dependent) packages generated with a single compiler invocation without any additional modifications?

awelc (approved):
Looks good to me overall!

language/tools/move-package/src/compilation/compiled_package.rs
tnowacki: Correct. And each definition (module or script) will have the package name tagged on top of it, in the field `package_name: Option<Symbol>`

awelc (approved):

Manahip:
Ah I guess I can answer my own question. Yes `CompiledPackage.root_compiled_units` are implicitly topologically sorted by dependency right now, thanks to https://github.com/move-language/move/blob/main/language/move-compiler/src/to_bytecode/translate.rs#L134

PR #98, opened 2022-05-04
import Diem Framework for testing and researching

vgao1996:
This PR gets the Diem Framework imported into the Move repo so that we can use it for future testing and researching.

More specifically, this
1. Imports the Move packages: `DPNFramework`, `DiemCoreFramework` and `DiemExperimental`
2. Imports a few crates that are needed for running unit tests (native function impl + crypto + df-cli). The code has been changed slightly to make it work within the new repo.
3. Adds two scripts `build_all.sh` and `test_all.sh`, which can be used to build and test all the three packages respectively. These scripts are also used by the newly added CI tests.

Here's an overview of the directory layout:
```
language/documentation/examples
  - diem-framework
    - crates
      - cli
      - crypto
      - crypto-derive
      - natives
    - move-packages
      - core
      - DPN
      - experimental
    - build_all.sh
    - test_all.sh
```


wrwg (commented):
This is awesome! We need it for improving the Move Prover, and testing the Move compiler.

I wonder whether we have a chance to eventually remove the Rust crates which come with it. I assume that is not the case because likely some tests depend on the specific Diem bindings. Even more so a reason why we need more power in CI.

vgao1996:
@tnowacki @sblackshear Let me know if this looks reasonable to you!

vgao1996:
Bump for a bit more attention

wrwg (approved):

PR #99, opened 2022-05-04
[evm] Contract modules and storage parameters (1st step)

### LINKED ISSUE #91: [evm] attributes related to EVM should have indicative names
Right now the EVM specific attributes have general names like `#[storage]`, `#[event(...)]`, `#[create(...)]`, ...

We should probably include a prefix to indicate that these are EVM specific attributes. e.g. `#[evm_storage]` or `#[evm(storage)]`.

wrwg:
This implements the first step towards passing references to contract storage to callables instead of requiring to use `borrow_global<T>(self())`. The intended functionality is not yet implemented but this prepares it.

- Now contracts are associated with modules. One has to use `#[evm_contract]` as a module attribute to identify an EVM contract. Only callables inside of that particular module are recognized to belong to the EVM contract. This also addresses and closes #91.
- As a result, one package can contain multiple contracts. However, the compiler will issue a warning if there is more than one contract and ignore all but the first because the remaining tool chain isn't yet prepared for dealing with multiple contracts.
- The `#[storage]` attribute now indicates that callables should take a reference to this struct. This attribute is optional, so old-style contracts which use `move_to` and `borrow` are still supported. Because `#[storage]` is not fully implemented yet, it has been removed where it was used, as all existing tests still use old-style storage.
- There was some heavier refactoring needed to be able to emit code for `move_to` and other storage related operations. Because the new regime requires to call `move_to`, `borrow` etc. outside of the `FunctionGenerator`, those functions had to be moved on `Generator` level. This is only partly used until now and will be fully leveraged in the 2nd step.

Nits:
- Compiler output of the package system has been redirected to stderr instead of stdout. This was done because hardhat drops stdout somehow, and seems cleaner anyway.

## Test Plan

Existing tests

### DIFF
--- a/language/evm/examples/sources/ERC1155.move
+++ b/language/evm/examples/sources/ERC1155.move
@@ -1,4 +1,4 @@
-#[contract]
+#[evm_contract]
 /// An implementation of the ERC-1155 Multi Token Standard.
 module Evm::ERC1155 {
     use Evm::Evm::{sender, self, sign, emit, isContract};
--- a/language/evm/examples/sources/ERC165.move
+++ b/language/evm/examples/sources/ERC165.move
@@ -1,4 +1,4 @@
-#[contract]
+#[evm_contract]
 /// An implementation of the ERC-165.
 module Evm::ERC165 {
     use Evm::IERC165;
--- a/language/evm/examples/sources/ERC20.move
+++ b/language/evm/examples/sources/ERC20.move
@@ -1,4 +1,4 @@
-#[contract]
+#[evm_contract]
 /// An implementation of the ERC-20 Token Standard.
 module Evm::ERC20 {
     use Evm::Evm::{sender, self, sign, emit};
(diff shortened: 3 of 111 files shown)


vgao1996:
> This was done because hardhat drops stdout somehow, and seems cleaner anyway.

@wrwg I took a look at the plugin code and can confirm that it does print out both stdout and stderr, but only if there's an error. Was this the thing that confused you? But you did mention you fixed the problem by changing the compiler output to stdout so I'm not sure if this is the real cause...

wrwg:
> > This was done because hardhat drops stdout somehow, and seems cleaner anyway.
> 
> @wrwg I took a look at the plugin code and can confirm that it does print out both stdout and stderr, but only if there's an error. Was this the thing that confused you? But you did mention you fixed the problem by changing the compiler output to stdout so I'm not sure if this is the real cause...

There was an error. It printed the fact that there is an error to stderr which showed, but the error messages where printed to stdout which did not show.

rahxephon89 (approved):

Issue #84, opened 2022-04-27
[Bug] Struct name identifier in unpack statement does not preserve location in translate AST


awelc:
# 🐛 Bug

Consider the following unpack statement:
```rust
let SomeStruct { some_field: value } = s;
```
In the statement above, the `SomeStruct` identifier does not have its own identity (e.g., with respect to its location in the source file) but rather is an alias to the struct definition identifier. This will cause problems when trying to build use-def information for identifiers in the source code (e.g., for the purpose of displaying this information in an IDE).

A simple solution is to stick the location information into the `Unpack` structure, but:
- a more general solution may exist (e.g., store "full" identifier info (`Name`) along with the alias to the struct definition)
- there may be other places where this kind of aliasing causes "use" identifiers to loose their identity

PR #102, opened 2022-05-05
[compiler] Added support for preserving locations of aliased identifiers

awelc:
## Motivation

This is a fix for a problem described in https://github.com/move-language/move/issues/84. In short, we want to preserve locations of identifiers that are aliases to modules or module-level definitions. It fixes the problem more generally (e.g., also for `Pack` and other AST nodes).

Yes


awelc:
Based on the wrong branch...

PR #103, opened 2022-05-05
[compiler] Added support for preserving locations of aliased identifiers

awelc:
## Motivation

This is a fix for a problem described in https://github.com/move-language/move/issues/84. In short, we want to preserve locations of identifiers that are aliases to modules or module-level definitions. It fixes the problem more generally (e.g., also for `Pack` and other AST nodes).

Yes

### DIFF
--- a/language/move-compiler/src/expansion/aliases.rs
+++ b/language/move-compiler/src/expansion/aliases.rs
@@ -161,20 +161,23 @@ impl AliasMap {
                     self.unused[*depth].modules.remove(n);
                 }
                 *depth_opt = None;
-                // What we are trying to do here is to preserve the
-                // location of the module name rather than replacing
-                // it with the location of the alias. The n parameter
-                // represents JUST the module name, so we leave
-                // locations related to module's address intact as
-                // replacing them with the location of the module name
-                // does not seem right
-                Some(sp(
-                    n.loc,
+                // We are preserving the name's original location,
+                // rather than referring to where the alias was
+                // defined. The name represents JUST the module name,
+                // though, so we do not change location of the address
+                // as we don't have this information.
+                // TODO maybe we should also keep the alias reference
+                // (or its location)?
+                let sp!(
+                    _,
                     ModuleIdent_ {
-                        address: ident.value.address,
-                        module: ModuleName(sp(n.loc, ident.value.module.value())),
-                    },
-                ))
+                        address,
+                        module: ModuleName(sp!(_, module))
+                    }
+                ) = ident;
+                let address = *address;
+                let module = ModuleName(sp(n.loc, *module));
+                Some(sp(n.loc, ModuleIdent_ { address, module }))
             }
         }
     }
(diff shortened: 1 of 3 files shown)


language/move-compiler/src/expansion/aliases.rs
```diff
@@ -152,28 +153,51 @@ impl AliasMap {
         self.unused.len()
     }
 
-    pub fn module_alias_get(&mut self, n: &Name) -> Option<&ModuleIdent> {
+    pub fn module_alias_get(&mut self, n: &Name) -> Option<ModuleIdent> {
         match self.modules.get_mut(n) {
             None => None,
             Some((depth_opt, ident)) => {
```
tnowacki, line 159: nits: Pattern match here:
```
Some((depth_opt, sp!(_, ModuleIdent_ {
              address,
              module: ModuleName(sp!(_, module)),
})
```
or 
```
Some((depth_opt, ident)) => {
   let  sp!(_, ModuleIdent_ { address, module: ModuleName(sp!(_, module))) }) = ident;
```
It is more consistent with what I try to do in the compiler. I find I get lost in long accessor chains

language/move-compiler/src/expansion/aliases.rs
```diff
@@ -152,28 +153,51 @@ impl AliasMap {
...
+                // locations related to module's address intact as
+                // replacing them with the location of the module name
+                // does not seem right
+                Some(sp(
+                    n.loc,
+                    ModuleIdent_ {
+                        address: ident.value.address,
+                        module: ModuleName(sp(n.loc, ident.value.module.value())),
+                    },
+                ))
```
tnowacki, line 177: Other nit, regardless of the pattern matching. I tend to prefer let binding and then name punning. It is usually the same number of lines or less.
So 
```
let address = ...;
let module = ...;
Some(sp(n.loc, ModuleIdent_ { address,  module }))
```

language/move-compiler/src/expansion/aliases.rs
```diff
@@ -152,28 +153,51 @@ impl AliasMap {
...
+                // What we are trying to do here is to preserve the
+                // location of the member name rather than replacing
+                // it with the location of the alias. The n parameter
+                // represents JUST the member name, so even if it's
+                // preceded by the module name, we don't know the
+                // location of this name - hence leaving ModuleIdent
+                // location intact.
+                Some((
+                    sp(ident_name.0.loc, ident_name.0.value),
+                    sp(n.loc, ident_name.1.value),
```
tnowacki, line 199: same nits as above, extra extra important here. I really tend to dislike the `.0` and `.1` type accessors unless I'm projecting out in a `.map` or `.filter` or something similar.

language/move-compiler/src/expansion/translate.rs
```diff
@@ -437,7 +437,9 @@ fn module_(
     for member in members {
         match member {
             P::ModuleMember::Use(_) => unreachable!(),
-            P::ModuleMember::Friend(f) => friend(context, &mut friends, f),
+            P::ModuleMember::Friend(f) => {
+                friend(context, &mut friends, f);
```
tnowacki, line 441: spurious change?

language/move-compiler/src/expansion/aliases.rs
```diff
@@ -152,28 +153,51 @@ impl AliasMap {
...
                 }
                 *depth_opt = None;
-                Some(ident)
+                // What we are trying to do here is to preserve the
+                // location of the module name rather than replacing
+                // it with the location of the alias. The n parameter
+                // represents JUST the module name, so we leave
+                // locations related to module's address intact as
+                // replacing them with the location of the module name
+                // does not seem right
```
tnowacki, line 170: Maybe just 
```
// We are preserving the names original location, rather than referring to where the alias was defined. 
// TODO maybe we should also keep the aliases loc? Or some reference to the alias?
```
I like the addition of a comment, but it feels just a tad long

language/move-compiler/src/expansion/aliases.rs
```diff
@@ -152,28 +153,51 @@ impl AliasMap {
...
+                ))
             }
         }
     }
 
-    pub fn member_alias_get(&mut self, n: &Name) -> Option<&(ModuleIdent, Name)> {
+    pub fn member_alias_get(&mut self, n: &Name) -> Option<(ModuleIdent, Name)> {
         match self.members.get_mut(n) {
             None => None,
             Some((depth_opt, ident_name)) => {
```
tnowacki, line 185: Here you could do `Some((depth_opt, (sp!(_, mname), sp!(_, mem_name))) =>`

tnowacki (approved):
Thanks for the changes for the nits! And thanks for the fix!

PR #104, opened 2022-05-05
[move-cli] Adds "--only-deps" argument to the move package build to allow skipping…

vladimirovmm:
Adds ``--only-deps`` argument to the ``move package build`` to allow skipping project source files in the compilation process.
This parameter will be used for autocomplete in IDE. It allows you to upload all dependencies to the ``build`` folder, even if there are errors in project. 
This is also useful for dependency testing.


awelc:
@vladimirovmm, from what I understand existing language servers for Move have been built on top of custom tool chains (treesitter and Dove). Do you plan to use the Move compiler as a foundation for the language server? Perhaps we should sync up on this idea?

awelc:
Actually, on closer inspection, I think I got it slightly wrong - Dove is used for local building whereas Pontem's VSCode server is using Move compiler for diagnostics (but not for anything else I think). Still, perhaps we should talk if you want to pursue this direction.

borispovod:
Hey @awelc ,

The current functional indeed for cases when IDE needed to index dependencies before build of package, so user can just fetch dependencies and continue develop his smart contracts.

Currently we support Intellij IDE plugin, there's not need for LSP as it based on Intellij framework. 

We are planning to continue work in LSP in the near future, yet not right now, hope we will have enough capacity in our team little later.

borispovod:
@awelc any update on it?

awelc:
> @awelc any update on it?

If you are asking about the PR, @tnowacki should be involved in this.

If it's about a follow-up on the LSP support, I am thinking about adding some additional features  to "our" language server, but if your plans on this front are not yet crystalized, there is probably no huge need for coordination until they do.

tnowacki (approved):
Looks like there is some side discussion on IDE features here, but as far as the flag + implementation, it seems reasonable to me!

tnowacki:
Actually though, before this lands, could you add a test for the flag?

tnowacki (changes requested):
Missed the lack of new tests. Just something basic even would be great. Requesting changes just to make sure it isn't missed!

language/tools/move-cli/src/sandbox/commands/test.rs
```diff
@@ -334,7 +334,7 @@ pub fn run_one(
     }
 
     let expected_output = fs::read_to_string(exp_path).unwrap_or_else(|_| "".to_string());
-    if expected_output != output {
+    if expected_output.trim() != output.trim() {
```
tnowacki, line 337: Why was this change needed?

vladimirovmm: @tnowacki When I was creating the test, I copied the result from the console along with not-display symbol that the editor does not display and could not delete it.
`format_diff(...)` also didn't display the difference.
Therefore, to simplify writing tests, I added `.trim()`

vladimirovmm: @tnowacki Remove it?

tnowacki: Sorry missed the mention!

I'd prefer to have it removed. We have `UB=1` or `UPBL=1` to update the baseline for these tests to prevent any copy paste issues (and to make your life easier :) )

vladimirovmm: @tnowacki Fixed

borispovod:
Hey guys, any update on it?

PR #106, opened 2022-05-05
[evm][move package] make `--arch ethereum` present in all builds, but return an error if feature `evm-backend` isn't enabled

vgao1996:
Previously, the `--arch ethereum` option is only present when feature `evm-backend` is enabled, which ended up causing some confusion. This PR makes the option always available, but return an error if `evm-backend` isn't enabled when  invoked.

### DIFF
--- a/language/tools/move-cli/src/package/cli.rs
+++ b/language/tools/move-cli/src/package/cli.rs
@@ -304,9 +304,12 @@ pub fn handle_package_commands(
                     config.compile_package(&rerooted_path, &mut std::io::stderr())?;
                 }
 
-                #[cfg(feature = "evm-backend")]
                 Architecture::Ethereum => {
+                    #[cfg(feature = "evm-backend")]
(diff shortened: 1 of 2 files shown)


language/tools/move-cli/src/package/cli.rs
```diff
@@ -304,9 +304,12 @@ pub fn handle_package_commands(
                     config.compile_package(&rerooted_path, &mut std::io::stderr())?;
                 }
 
-                #[cfg(feature = "evm-backend")]
                 Architecture::Ethereum => {
+                    #[cfg(feature = "evm-backend")]
                     config.compile_package_evm(&rerooted_path, &mut std::io::stderr())?;
+
+                    #[cfg(not(feature = "evm-backend"))]
+                    bail!("The Ethereum architecture is not supported because move-cli was not compiled with feature flag `evm-backend`.");
```
wrwg, line 312: Great!

wrwg (approved):

Issue #95, opened 2022-05-03
[Feature Request][Table] Lifetime support? Let table take K instead of &K?


msmouse:
# 🚀 Feature Request

I'm not quite sure what I'm asking in practice. Maybe taking the key by value (and require K: copy).

## Motivation
This doesn't compile because "Invalid return. Local variable 'bucket_idx' is still being borrowed.".
A similar thing can be done easily in Rust.
```
module AptosFramework::BucketedVector {
    use AptosFramework::Table;
    use Std::Vector;

    struct BucketedVector<V> has store {
        bucket_size: u64,
        buckets: Table::Table<u64, vector<V>>
    }

    public fun borrow<T: store>(vec: &BucketedVector<T>, idx: u64): &T {
        let bucket_idx = idx / vec.bucket_size;
        let item_idx = idx % vec.bucket_size;

        let bucket = Table::borrow(&vec.buckets, &bucket_idx);
        Vector::borrow(bucket, item_idx)
    }
}
```

## Pitch

**Describe the solution you'd like**
Make it work so we can implement a BucketedVector.

**Describe alternatives you've considered**
Let user code manage the buckets directly.

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))
I'm not able to.

wrwg:
Bummer. This is a result of the over-approximation of the borrowing graph in Move. For the call `let r = Table::borrow(&vec.buckets, &bucket_idx)`, Move assumes that `r` is borrowing from *both* input parameters, because in contrast to Rust, there are no borrow annotations which could tell it that it borrows only from the first parameter.

I don't know whether there is a possible workaround by hacking the compiler to make it aware of more fine-grained borrow information for certain native functions,

Otherwise, the only option is to make table keys `K:copy`.

This is just another instance of that Move's borrow semantics is unfortunately not always working as expected, because of its simplifications compared to Rust.

tnowacki:
A quick response, and I can give a more detailed one later (traveling today). 

- Lifetime support is not easy to add in the current system. There is too much ambiguity in the borrow graph in relation to equivalent references. 
- @sblackshear and I have been working a long time on a new borrow checker formalization (and in turn implementation).  This set-based approach will be more amenable to lifetime support. 
- I can try to add this new set-based approach in the next month or so. (hopefully)
[...]

PR #109, opened 2022-05-05
[table-extenstion] Add trait bounds (copy + drop) to Table key

lightmark:
## Motivation

According to https://github.com/move-language/move/issues/95, Move lacks of lifetime support so we have to find a way to work around this. Otherwise, there is no way to implement clean APIs for many struct built on top of `Table` such as `IterableTable`.

This PR add trait bounds (copy + drop) to the `key`

Yes

## Test Plan

cargo test

### DIFF
--- a/language/extensions/move-table-extension/sources/Table.move
+++ b/language/extensions/move-table-extension/sources/Table.move
@@ -9,13 +9,13 @@ module Extensions::Table {
     const ENOT_EMPTY: u64 = 102;
 
     /// Type of tables
-    struct Table<phantom K, phantom V> has store {
+    struct Table<phantom K: copy + drop, phantom V> has store {
         handle: u128,
         length: u64,
     }
 
     /// Create a new Table.
-    public fun new<K, V: store>(): Table<K, V> {
+    public fun new<K: copy + drop, V: store>(): Table<K, V> {
         Table{
             handle: new_table_handle(),
             length: 0,
@@ -23,7 +23,7 @@ module Extensions::Table {
     }
 
     /// Destroy a table. The table must be empty to succeed.
-    public fun destroy_empty<K, V>(table: Table<K, V>) {
+    public fun destroy_empty<K: copy + drop, V>(table: Table<K, V>) {
         assert!(table.length == 0, Errors::invalid_state(ENOT_EMPTY));
         destroy_empty_box<K, V, Box<V>>(&table);
         drop_unchecked_box<K, V, Box<V>>(table)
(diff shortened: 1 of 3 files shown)


language/extensions/move-table-extension/sources/Table.move
```diff
@@ -43,12 +43,22 @@ module Extensions::Table {
         &borrow_box<K, V, Box<V>>(table, key).val
     }
 
+    /// Same as `borrow` but pass in a copyable key instead.
+    public fun borrow_by_key<K: copy, V>(table: &Table<K, V>, key: K): &V {
```
msmouse, line 47: I'm not really a fan of this name..

language/extensions/move-table-extension/sources/Table.move
```diff
@@ -43,12 +43,22 @@ module Extensions::Table {
...
+        &borrow_box_by_key<K, V, Box<V>>(table, key).val
```
msmouse, line 48: doesn't this work `&native_borrow_box<K, V, Box<V>>(table, &key).val`?

Do you need to introduce `borrow_box_by_key`?

msmouse: K. So native functions are checked as well. You won. 😂 
Do some code reuse on the native side at least, though.

lightmark: It doesn't what because it surfaces the same problem we try to solve here.

language/extensions/move-table-extension/sources/Table.move
lightmark: me, either.

language/extensions/move-table-extension/src/lib.rs
```diff
@@ -420,6 +414,34 @@ fn native_borrow_box(
         smallvec![val],
     ))
 }
+fn native_borrow_box(
```
msmouse, line 417: nit: blank line

language/extensions/move-table-extension/sources/Table.move
wrwg: I think we should switch the entire API to use copyable keys everywhere. If we leave it like suggested here, it will create a bunch of confusion, AFAICT this API is rarely used for now, so we should be able to do this.

The original functionality of being able to use arbitrary non-copyable data as keys would then need to be achieved by `borrow(table, sha256(&non_copyable_key))`.

jolestar: > I think we should switch the entire API to use copyable keys everywhere. If we leave it like suggested here, it will create a bunch of confusion, AFAICT this API is rarely used for now, so we should be able to do this.
> 
> The original functionality of being able to use arbitrary non-copyable data as keys would then need to be achieved by `borrow(table, sha256(&non_copyable_key))`.

I totally support adding `copy` to the key type.

wrwg: > > I think we should switch the entire API to use copyable keys everywhere. If we leave it like suggested here, it will create a bunch of confusion, AFAICT this API is rarely used for now, so we should be able to do this.
> > The original functionality of being able to use arbitrary non-copyable data as keys would then need to be achieved by `borrow(table, sha256(&non_copyable_key))`.
> 
> I totally support adding `copy` to the key type.

It's sad that we have to do this because of Move borrow semantics restriction. From a language design viewpoint, we are hitting a problem of lacking compositionality. Basically, looking at `borrow(table, sha256(&non_copyable_key))`, we are not able to define a new function `borrow'(table, &k) == borrow(table, sha256(k))`. This is a serious language restriction which we should address, at least for library builders (this kind of developers could use attributes, no general lifetime annotations needed). However, we won't be able to address this in the same time frame as we need tables ready. 

@sblackshear @tnowacki @davidiw

lightmark: @tnowacki I thought you were thinking about implementing lifetime in `move` long time ago?

davidiw: Actually, this is a great point, I'd be curious from @msmouse 's perspective on the cost associated with the following:
* copyable
* reference
* reference + bcs + sha256 (would this satisfy the borrow checker requirements?)

zekun000: we probably still want to have the by reference API after borrow checker upgrades?

tnowacki (commented):
I'll discuss lifetime support on the issue #95

language/extensions/move-table-extension/sources/Table.move
wrwg: We don't know when, if, and how the borrow checker will be upgraded. If it is we can introduce the reference API again. In the meantime, it just creates confusion for Move users, as it is not fully usable.

wrwg: Let me also add that the decision to use `&K` was not unanimous. Some folks preferred the `K has copy` approach from the beginning.

msmouse: I like the route where we use `K has copy` in all the APIs, because as a beginner Mover those APIs will hint more explicitly that the size of the key matters. And smaller keys doesn't need to go through sha256 to waste CPU and be inflated in size (@davidiw did you imply a solution where we always sha256 the keys?)

wrwg:
I suggest to conduct an out-of-band meeting to discuss this early next week for interested parties. Perhaps hold on merging or working on the alternative, making keys everywhere copyable, until we discussed. Anybody wanting to join, say so on Discord and I setup an invite.

lightmark:
Updated, changing key to be copyable everywhere

lightmark:
@wrwg @davidiw plz have a look, thanks

davidiw (approved):

wrwg (approved):
Awesome, thank you.

PR #110, opened 2022-05-06
[evm] Contract modules and storage parameters (part 2)

wrwg:
This completes the support of the storage struct references as parameters for callable functions.

- If `#[storage]` is given, the `#[create]` function must return an instance of it. The Yul generator initializes the storage from this value in the constructor dispatcher.
- A `#[callable]` function can take a mutable or immutable reference to the storage as parameter. This parameter is initialized in the function dispatcher.

A test has been added, and the `ERC20Mock.move` adapted to the new model. The diff on this example shows how to use it and in which ways it is better.


rahxephon89 (approved):

junkil-park (approved):
It indeed looks much nicer!

PR #90, opened 2022-05-01
[move-compiler] Support CRLF as newline for Move files

yubing744:
## Motivation

Fix issue: https://github.com/starcoinorg/starcoin/issues/3130

### Have you read the [Contributing Guidelines on pull requests](https://github.com/move-language/move/blob/main/CONTRIBUTING.md#pull-requests)?

Yes

### move-ir-compiler
* test case
language/move-ir-compiler/transactional-tests/tests/parsing/crlf.mvir

* run test
``
  cd language/move-ir-compiler/transactional-tests && cargo test
``

### move-compiler
* test case
language/move-compiler/tests/move_check/parser/newline_crlf.move

* run test
``
  cd language/move-compiler && cargo test
``

PR #116, opened 2022-05-08
[move-compiler] fix invalid character cr when match tab cr lf

yubing744:
## Motivation

Fix bug for https://github.com/move-language/move/pull/90

image

YES

### move-compiler
* test case
language/move-compiler/tests/move_check/parser/newline_crlf_start_with_tab_or_space.move

* run test
``
cd language/move-compiler && cargo test
``


language/move-compiler/src/parser/lexer.rs
```diff
@@ -224,8 +224,7 @@ impl<'input> Lexer<'input> {
         // a multi-line or single-line comment.
         loop {
             // Trim only the whitespace characters we recognize: newline(\n|\r\n), tab, and space.
-            text = text.trim_start_matches("\r\n");
-            text = text.trim_start_matches(|c: char| matches!(c, '\n' | '\t' | ' '));
+            text = text.trim_start_matches(|c: char| matches!(c, '\r' | '\n' | '\t' | ' '));
```
wrwg, line 227: As far as I see this would again allow us to accept arbitrary CR in the source, which we do not want, Your trimming algorithm need to be smarter and handwritten. Something like:

```rust
fn move_trim(s: &str) -> String {
  let mut i = 0;
  while i < s.len() {
    match s[i] {
      `\t' | '\r' |  '\n` => i += 1,
      `\r' if i > 0 && s[i-1] == `\n` => i += 1,
      _ => break
    }
  }
  s[i..].to_string()
} 
```

yubing744: @wrwg If verify_string has ensured that CRLF is present at the same time, then trim_whitespace_and_comments will not have a separate CR here.

yubing744:
@wrwg cargo x lint can skip some move file check?

language/move-compiler/src/parser/lexer.rs
wrwg: Hey Owen, can you use a link to the source in your github clone instead of a screenshot? This would make it easier to verify this.

You may not be aware of this, but a few years ago, Diem had a significant problem with external auditors pointing out security issues with the lexer, the code you are modifying. This was back then a public relations disaster. That is why we are today very, very careful about which characters we allow. An auditor may come along trim your code and don't see your verify_string. So for double extra safety, make the trim function secure to only trim CRLF and not standalone CR. Thank you.

yubing744: Ok, thank you very much for the tip. I will pay particular attention to the security of the submitted PR later.

yubing744: I'll replace the screenshots later

yubing744:
@wrwg I added a trim_whitespace function that needs to be reviewed.

language/move-compiler/src/parser/lexer.rs
```diff
@@ -638,3 +637,68 @@ fn get_name_token(name: &str) -> Tok {
...
+        assert_eq!(trim_whitespace("\r\n\nxxx"), "xxx");
+
+        assert_eq!(trim_whitespace(" \r\n\r\n"), "");
+        assert_eq!(trim_whitespace("\r\n \t\n"), "");
+
+        assert_eq!(trim_whitespace(" \r\n\r\nxxx"), "xxx");
+        assert_eq!(trim_whitespace("\r\n \t\nxxx"), "xxx");
+
+        assert_eq!(trim_whitespace(" \r\n\r\nxxx\n"), "xxx\n");
+        assert_eq!(trim_whitespace("\r\n \t\nxxx\r\n"), "xxx\r\n");
```
jolestar, line 702: Why is the end `\r\n` not trimmed?

language/move-compiler/src/parser/lexer.rs
```diff
@@ -638,3 +637,68 @@ fn get_name_token(name: &str) -> Tok {
...
+
+    &text[pos..]
+}
+
+#[cfg(test)]
+mod tests {
+    use super::trim_whitespace;
+
+    #[test]
+    fn test_trim_whitespace() {
```
jolestar, line 670: Need a test case for single `\r` which not be trimmed.

language/move-compiler/src/parser/lexer.rs
```diff
@@ -638,3 +637,68 @@ fn get_name_token(name: &str) -> Tok {
...
+fn trim_whitespace(text: &str) -> &str {
+    let mut pos = 0;
+    let mut iter = text.chars();
+
+    while let Some(chr) = iter.next() {
+        match chr {
+            ' ' | '\t' | '\n' => pos += 1,
+            '\r' => {
+                let nch = iter.next();
+                if nch.is_some() && nch.unwrap() == '\n' {
```
jolestar, line 651: Should rewrite to

```
if let Some('\n') = nch {
}
```
, is more simple?

language/move-compiler/src/parser/lexer.rs
yubing744: Only trim the start characters, I change the function name.

language/move-compiler/src/parser/lexer.rs
yubing744: OK,I will add a test case

language/move-compiler/src/parser/lexer.rs
yubing744: Good

yubing744: @jolestar updated!

language/move-compiler/src/parser/lexer.rs
```diff
@@ -638,3 +637,72 @@ fn get_name_token(name: &str) -> Tok {
...
+            ' ' | '\t' | '\n' => pos += 1,
+            '\r' => {
+                if let Some('\n') = iter.next() {
+                    pos += 2;
+                    continue;
+                }
+
+                break;
+            }
+            _ => break,
```
tnowacki, line 657: This is likely fine, but you can also consider: 
```
match chr {
    ' ' | '\t' | '\n' => pos += 1,
    '\r' if matches!(iter.next(), Some('\n')) => pos += 2,
    _ => break,
};
```
But I know that while this is easier for me to read, it might not be for everyone else :)

yubing744: That's great, it's very concise.

yubing744: @tnowacki I updated the PR and passed the CI review.

tnowacki (approved):

PR #119, opened 2022-05-09
[transactional-tests][breaking] New argument parsing

tnowacki:
- New argument parsing for value args
- Breaking for old tests, but now consistent with source language syntax
- Some additional minor tweaks for the transactional test framework 

This will break the syntax for existing transactional tests, most notably around addresses. In the past `0x42` was an address value, but now it is following the source languages syntax so that same argument will need a `@`, e.g. `@0x42` 

## Motivation

- This allows custom parsable values from the transactional test framework
- In the future, it should also allow for custom CLI args for anyone wrapping the Move CLI 

## Follow ups

- Use this parser for the CLI
- Mark the MoveValue parser as deprecated 

## Test Plan

- Some small tests
- Updated tests


language/move-command-line-common/src/address.rs
```diff
@@ -0,0 +1,165 @@
...
+            Some((n, format)) => Ok(NumericalAddress {
+                bytes: AccountAddress::new(n),
+                format,
+            }),
+            None =>
+            // TODO the kind of error is in an unstable nightly API
+            // But currently the only way this should fail is if the number is too long
+            {
+                Err(format!(
+                    "Invalid address literal. The numeric value is too large. The maximum size is \
```
sblackshear, line 74: Nit: printing `s` might be useful.

sblackshear (approved):
Nice! An improved version of this has been sorely needed. 

My other wishlist item for the future is a better `TypeTag` parser...

tnowacki:
> My other wishlist item for the future is a better TypeTag parser...

This IS a better TypeTag parser. Or at the very least you can convert ParsedType into TypeTag

PR #126, opened 2022-05-12
[package manager] Improve unresolved addresses error message

sblackshear:
Fixing an unresolved address issue is easy (requires a small addition to Move.toml), but that may not be obvious to package system novices. Suggest the resolution in the error message.

### DIFF
--- a/language/tools/move-package/src/resolution/resolution_graph.rs
+++ b/language/tools/move-package/src/resolution/resolution_graph.rs
@@ -169,8 +169,11 @@ impl ResolvingGraph {
 
         if !unresolved_addresses.is_empty() {
             bail!(
-                "Unresolved addresses found: [\n{}\n]",
-                unresolved_addresses.join("\n")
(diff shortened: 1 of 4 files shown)


awelc (approved):

language/tools/move-package/src/resolution/resolution_graph.rs
```diff
@@ -169,8 +169,9 @@ impl ResolvingGraph {
 
         if !unresolved_addresses.is_empty() {
             bail!(
-                "Unresolved addresses found: [\n{}\n]",
-                unresolved_addresses.join("\n")
+                "Unresolved addresses found: [\n{}\n]\nTo fix this, add an entry for each unresolved address to the [addresses] section of {}/Move.toml: e.g.,\n[addresses]\nStd = \"0x1\"",
```
tnowacki, line 172: Nit, can you fix the string so it fits in the column length

tnowacki (approved):

Issue #124, opened 2022-05-12
[vectors] stdlib tests fail on CI


wrwg:
Possibly because Move unit test when embedded in Rust tests don't lead to a test failure, we have a few vector test failures which sneaked into the main branch. From running move-stdlib tests:

```
[ FAIL    ] 0x1::VectorTests::borrow_out_of_range
[ FAIL    ] 0x1::VectorTests::destroy_non_empty
[ FAIL    ] 0x1::VectorTests::pop_out_of_range
[ FAIL    ] 0x1::VectorTests::swap_empty
[ FAIL    ] 0x1::VectorTests::swap_out_of_range
```

Those failures do *not* need lead to Rust test failure, and therefore passed by CI.

The failures have this funny shape:

```
┌── borrow_out_of_range ──────
│ error[E11001]: test failure
│    ┌─ /home/wrwg/gh/move/language/move-stdlib/./tests/VectorTests.move:97:9
│    │
│ 94 │     fun borrow_out_of_range() {
│    │         ------------------- In this function in 0x1::VectorTests
│    ·
│ 97 │         V::borrow(&v, 1);
│    │         ^^^^^^^^^^^^^^^^ Test did not abort with expected code. Expected test to abort with 1 but instead it aborted with 1 here
│ 
│ 
└──────────────────
```

So 1 was expected as an abort code but 1 was found...

Perhaps the failures are related to the new vector operations introduced in #37.

I'm rolling up PR #125 to fix the issue that Move unit tests don't produce a test failure, but will have  to disable some tests temporarily.

wrwg:
@tnowacki suspecting the vector opcodes for the failing tests, but the cause may also be older.

PR #127, opened 2022-05-12
[move-stdlib] Fix vector tests

tnowacki:
- Fixed vector tests failing, re-enabled tests
- The tests exited with VECTOR_OPERATION_ERROR instead of ABORTED
- We should be able to specify the difference in the future

## Motivation

#124

## Test Plan

- re-enabled tests


wrwg (approved):

tnowacki:
This is a temporary fix, I'd like to see something more robust in the future, but I want the tests online for now.
See #128 for future improvements

PR #147, opened 2022-05-17
[language server] Partial implementation of go-to-def and go-to-references

awelc:
## Motivation

IDE features supported in the language server are at the moment somewhat limited. This PR implements partial support for the go-to-def and go-to-references features. 

The main purpose here is to support building symbols information for the whole language, in particular information allowing us to connect identifier uses with their definitions.

There is still a fair amount of work to have these features supported smoothly in the IDE frontend, but let's take this process step-wise and start with reviewing the symbolication analysis itself. Missing pieces include (but are not limited to):
- making symbolicator run asynchronously and repeatedly on source code changes
- at this point, the IDE support works under the assumption that it's opened in the folder containing the Move manifest file - we should include support for discovering directory structure upon opening a random Move source file
- symbolication will succeed only if the Move code opened in the IDE actually builds and at this point there is not indication whether it does or does not build - we should pass compiler diagnostics back to the IDE to deliver a visual clue to  the developer that something is wrong with the code

As a result of the missing features described above, the code in this PR is not quite ready for wider deployment, say, to the VSCode plugin. That being said, it already works pretty well for code browsing and could be user-tested this way. In order to do it, one has to set `symbols::DEFS_AND_REFS_SUPPORT` constant to `true`, rebuild the language server, and open the IDE in a directory containing the Move manifest file.

Yes

## Test Plan

Tests for symbolicating all individual language features are part of this PR. We also make standard library a dependency of test Move packages to make sure that the symbolicator can process larger amounts of fairly complicated Move code.


language/move-analyzer/src/symbols.rs
```diff
@@ -0,0 +1,2184 @@
...
+};
+use move_ir_types::location::*;
+use move_package::compilation::build_plan::BuildPlan;
+use move_symbol_pool::Symbol;
+
+/// Enabling/disabling the language server reporting readiness to support go-to-def and
+/// go-to-references to the IDE.
+pub const DEFS_AND_REFS_SUPPORT: bool = false;
+
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
```
tnowacki, line 84: Is there a reason not to give this `Copy`? Doesn't look like it to me

tnowacki: Same for similar types below

language/move-analyzer/src/symbols.rs
```diff
@@ -0,0 +1,2184 @@
...
+            col_end,
+        };
+
+        match references.get_mut(&def_loc) {
+            Some(v) => {
+                v.insert(use_loc);
+            }
+            None => {
+                let mut s = BTreeSet::new();
+                s.insert(use_loc);
```
tnowacki, line 205: `BTreeSet::from([use_loc])` might be a bit nicer :)

language/move-analyzer/src/symbols.rs
```diff
@@ -0,0 +1,2184 @@
...
+                references.insert(def_loc, s);
+            }
+        }
```
tnowacki, line 208: references.entry(def_loc).or_insert_with(BTreeSet::new).insert(use_loc)

language/move-analyzer/src/symbols.rs
```diff
@@ -0,0 +1,2184 @@
...
+#[derive(Debug)]
+struct UseDefMap(BTreeMap<u32, BTreeSet<UseDef>>);
+
+impl UseDefMap {
+    fn new() -> Self {
+        Self(BTreeMap::new())
+    }
+
+    fn insert(&mut self, key: u32, val: UseDef) {
+        match self.0.get_mut(&key) {
```
tnowacki, line 242: similar entry trick from above

language/move-analyzer/src/symbols.rs
```diff
@@ -0,0 +1,2184 @@
...
+                &pos,
+                &mut references,
+                module_def,
+                &files,
+                &file_id_mapping,
+            );
+            mod_outer_defs.insert(*module_ident, defs);
+            mod_use_defs.insert(*module_ident, symbols);
+            match source_files.get(&pos.file_hash()) {
+                Some((fpath, _)) => {
```
tnowacki, line 316: I like trying to reduce horizontal indentation when possible, but not necessary
```
let ((fpath, _)) = match source_files.get(&pos.file_hash()) { SOme((fpath, _)) => fpath, None => continue }
```

language/move-analyzer/src/symbols.rs
```diff
@@ -0,0 +1,2184 @@
...
+        let mut use_def_map = UseDefMap::new();
+
+        for (pos, name, def) in &mod_def.structs {
+            match Self::get_start_loc(&pos, files, file_id_mapping) {
+                Some(name_start) => {
+                    let mut field_defs = vec![];
+                    if let StructFields::Defined(fields) = &def.fields {
+                        for (fpos, fname, _) in fields {
+                            match Self::get_start_loc(&fpos, files, file_id_mapping) {
+                                Some(start) => {
```
tnowacki, line 383: Same sort of things as above, trying to do a match where you do `let data = match ... { ... None => continue }` is going to make this a lot more readable

language/move-analyzer/src/symbols.rs
```diff
@@ -0,0 +1,2184 @@
...
+        }
+
+        let fhash = loc.file_hash();
+        (
+            ModuleDefs {
+                fhash,
+                structs,
+                constants,
+                functions,
+            },
```
tnowacki, line 491: and a mega nit, I tend to not like it when my tuples span multiple lines. I think it just reads funny. But its a personal preference. 

I would do 
```
let module_defs =  ModuleDefs {
                fhash,
                structs,
                constants,
                functions,
            };
(module_defs,   use_def_map)
```
to avoid that (ignoring my horrible indentation)

tnowacki (commented):
I think this should be enough to start on comments, rest of things look fine I think, and a bit more mechanical
Big ones just being that:
- avoiding horizontal indentation in spots can be helpful
- .entry is cool
- giving these types `Copy` might help remove some cruft in spots

language/move-analyzer/src/symbols.rs
awelc: I did not want to add trait derivations unless the value of the type needs to be copied but does not implement the `Copy` trait and the compiler barfs.

language/move-analyzer/src/symbols.rs
awelc: This is pretty cool!

language/tools/move-package/src/resolution/resolution_graph.rs
```diff
@@ -769,3 +770,22 @@ impl ResolvedPackage {
...
+impl ResolvedGraph {
+    pub fn file_sources(&self) -> BTreeMap<FileHash, (Symbol, String)> {
+        self.package_table
+            .iter()
+            .flat_map(|(_, rpkg)| {
+                rpkg.get_sources(&self.build_options)
+                    .unwrap()
+                    .iter()
+                    .map(|fname| {
+                        let contents = read_to_string(Path::new(fname.as_str())).unwrap();
```
tnowacki, line 783: nit, bit more standard to `use std::fs` and then `fs::read_to_string`, but not a big deal either way

language/tools/move-package/src/resolution/resolution_graph.rs
```diff
@@ -769,3 +770,22 @@ impl ResolvedPackage {
...
+                        let fhash = FileHash::new(&contents);
+                        (fhash, (*fname, contents))
+                    })
+                    .collect::<BTreeMap<_, _>>()
```
tnowacki, line 787: Do you need this inner collect? Just curious, what error do you get without it?

language/tools/move-package/src/resolution/resolution_graph.rs
```diff
@@ -769,3 +770,22 @@ impl ResolvedPackage {
         }
     }
 }
+
+impl ResolvedGraph {
+    pub fn file_sources(&self) -> BTreeMap<FileHash, (Symbol, String)> {
```
tnowacki, line 775: Nit, probably fine to move this into the same `impl ResolvedGraph` block above

language/move-analyzer/src/symbols.rs
```diff
@@ -0,0 +1,2164 @@
...
+        let mut file_id_mapping = BTreeMap::new();
+        let mut file_name_mapping = BTreeMap::new();
+        for (fhash, (fname, source)) in source_files {
+            let id = files.add(*fname, source.clone());
+            file_id_mapping.insert(*fhash, id);
+            file_name_mapping.insert(*fhash, *fname);
+        }
+
+        let build_plan = BuildPlan::create(resolution_graph)?;
+        let mut typed_ast = vec![];
```
tnowacki, line 269:
```
let mut typed_ast = None;
...
typed_ast = Some(typed_program.clone());
...
let modules = typed_ast.unwrap()
```
Might be a bit more idiomatic, and then you don't need the `== 1`

language/move-analyzer/src/symbols.rs
```diff
@@ -0,0 +1,2164 @@
...
+pub fn on_use_request(
+    context: &Context,
+    symbols: &Symbols,
+    use_fpath: &str,
+    use_line: u32,
+    use_col: u32,
+    id: RequestId,
+    use_def_action: impl Fn(&UseDef) -> Option<serde_json::Value>,
+) {
+    let mut result = Option::<serde_json::Value>::None;
```
tnowacki, line 1374: Second use of `Option::None` I've seen. Whats the issue you ran into when you just used `None`?

tnowacki (approved):
I have no real issues with this and will approve to unblock you! 

Left some more Rust style related comments.
The general logic looks fine and straight forward. 

I am wondering though if it would have been better to use `Loc`s everywhere. And then if you really needed the LSP based query, to do another pass that converted the Loc based maps to this more custom logic. 
The reason I'm thinking about this is that this definition mapping might be useful for other scenarios, but now the logic is rather custom for just LSP and not that portable.
But this isn't something that needs to be solved in this PR!

language/tools/move-package/src/resolution/resolution_graph.rs
awelc: I am pretty sure this was inspired by other code pieces in the same file, rather than a result of a deep thought :-) There is an error, though, if collect is not there:
```
cannot return value referencing temporary value
```

In fact, adding the collect call is what is suggested by flycheck

language/move-analyzer/src/symbols.rs
awelc: No error... Might have been a copypasta from the use in the previous method when you do need to specify the type.

PR #150, opened 2022-05-19
[move prover] add unchecked subtraction

whst:
## Motivation

Just like we can ignore addition overflow, it is more convenient to ignore subtraction underflow in some cases.

Yes.

## Test Plan

### DIFF
--- a/language/move-model/src/pragmas.rs
+++ b/language/move-model/src/pragmas.rs
@@ -50,6 +50,9 @@ pub const ALWAYS_ABORTS_TEST_PRAGMA: &str = "always_aborts_test";
 /// for overflow.
 pub const ADDITION_OVERFLOW_UNCHECKED_PRAGMA: &str = "addition_overflow_unchecked";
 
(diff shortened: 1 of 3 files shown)


wrwg:
Can you please explain closer why you want to do this? It is problematic to just change the semantics of the underlying Move program.

If you want to prevent propagating an abort condition to the caller you can also do:

```
spec some_fun {
   pragma opaque; // Only use the spec at caller side, not the function implementation
   aborts_if [concrete] <condition under which this implementation aborts>;
   aborts_if [abstract] true; // on caller side assume this never aborts
}
```

There should be some examples of this in the Diem framework specs, e.g. here: https://github.com/move-language/move/blob/8958b40e97a4f94a588e5365fdb90ec18f00d888/language/documentation/examples/diem-framework/move-packages/DPN/sources/DiemConfig.move#L355

wrwg:
Closing as we aren't moving forward with this. Feel free to reopen if this is the wrong call.

PR #151, opened 2022-05-19
[move-compiler] Allow 'vector' as a module name

tnowacki:
## Motivation

- old rule overly protective. 

## Test Plan

- updated tests


damirka (approved):
Looks good!

PR #152, opened 2022-05-19
[move-ir] De-tokenize primitive types

tnowacki:
- primitive type names are no longer tokens. They use the name token like other identifiers

## Test Plan

- internal change


vgao1996 (approved):
LGTM. One question though: what are the long-term benefits you think this would this buy us? Being able to use these previously reserved words at different places depending on contexts?

Issue #58, opened 2022-04-19
[move package test] panics when cache is hot


vgao1996:
Right now if you run `move package test` inside a package that depends on MoveStdlib and MoveStdlib only, for the first time it'll succeed:
```
BUILDING MoveStdlib
BUILDING Foo
Running Move unit tests
[ PASS    ] 0x1::A::one_plus_one
[ PASS    ] 0x1::A::should_panic
[ PASS    ] 0x1::A::test_with_arg
Test result: OK. Total tests: 3; passed: 3; failed: 0
```
However if you run this command for the second time, if complains about not being able to find a source file:
```
CACHED MoveStdlib
BUILDING Foo
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: No such file or directory './build/MoveStdlib/sources/ASCIITests.move'', language/tools/move-cli/src/package/cli.rs:510:83
```

Issue #107, opened 2022-05-05
Tracking Issue for Hardening the Toolings Related to Move on EVM


vgao1996:
- [ ] Make `--arch ethereum` available in all builds #106
- [ ] Unify `move --ethereum package build` and `move package test --evm`
- [ ] Update the documentation for `move-cli`
- [ ] Remove the `build/evm` directory before starting a new build #105
- [ ] Potential bug: cli tests `rebuild_after_touching_manifest` may fail (due to a preexisting build directory?)

PR #153, opened 2022-05-20
rename move-stdlib tests to match module names

jamesmeijers:
## Motivation

#58 #107 
The test files ASCIIStringTests.move and OptionsTests.move were misnamed, their filename did not match the the module name within. This caused bug #58 . This is a quick fix, in the long run I don't think the system should not be assuming file names based on module names, even if naming conventions are suggested. 

Yes

## Test Plan

I tested it locally to ensure the bug was fixed.


tnowacki:
@vgao1996 can you confirm this fixes #58 ?

vgao1996 (approved):
@tnowacki This fixes the specific case mentioned in #58, but as @jamesmeijers mentioned, does NOT address the general issue. I'll take a look at the package system and try to correct it there.

PR #156, opened 2022-05-20
[RFC][codemod] Make module names lower case

tnowacki:
- Made all stdlib module lower snake case
- I probably missed some cases, but wanted to put up the initial PR 

## Motivation

- As discussed in the planning meeting, the belief is that doing this will add some texture to the files, where you can easily differentiate between a module and the types in that module. We have seen some repeated paper cuts of people doing things like `use Addr::MyType;` and then being confused when they get an error for undefined type `MyType` because they needed `use Addr::MyType::MyType`
- There are other ways to solve the described problem above, but I think its a good indication of lack of visual texture in the current style


tnowacki:
Mostly looking for sort of pre-approval for this before proceeding with fixing all the tests :)

jolestar (commented):
Maybe we should support the dir module and the file module like Rust, so we can reorg the stdlib by dir, like `Std::Hash` => `std::util::hash`, and keep `Hash.move` and `util/hash.move` both for compatibility. This feature also can help to organize the Move source files, when a project has many modules.

CapCap:
I agree with @jolestar ; as is this is significantly more confusing, and I believe it removes texture vs adding to it. Most languages (rust included) use capitals to help denote Classes, Modules, Constants, etc; having things lowercased like this violates the expectations set forth by most languages and conventions that exist.

tnowacki:
> Maybe we should support the dir module and the file module like Rust, so we can reorg the stdlib by dir, like Std::Hash => std::util::hash, and keep Hash.move and util/hash.move both for compatibility. This feature also can help to organize the Move source files, when a project has many modules.

We don't have nested named address spaces, so this is not something we can support right now. At least by this design. But it might be worthwhile to split various modules in to separates spaces! Not everything needs to be in `Std` :)

tnowacki:
>Many languages use capitals to help denote Classes, Modules, Constants, etc; having things lowercased like this violates the expectations set forth by most languages and conventions that exist.

I hear you, and languages, specifically OCaml, influenced our previous standard here. 

But we have found over time is that it is confusing to have things like `Coin` as a module and `Coin` as a type to share the same name. This can be particularly confusing if you have an error that says "Unbound type 'Coin'", but you see `Coin` functions, e.g. `Coin::zero()`  all over a file.

There are a few different ways we could change the coding standards here to help with this. One would be to go back to the ML/OCaml style of doing `t` or `T` as the main type in a module, so `Coin::T` instead of `Coin::Coin`. We found this though to clash with our more Java/C++/Rust style for generics.

All of these things (and some things in the works for doing receiver style syntax or other type associated functions ala Rust `impl` blocks) led us to choosing the Rust style of lower snake case for module names. 

TL;DR we want to differentiate visually between module and type names. Module names never exist on their own (they are always used to access something, `m::foo` or `m::Coin`) and they were the best candidate for change.

jolestar:
> We don't have nested named address spaces, so this is not something we can support right now. At least by this design. But it might be worthwhile to split various modules in to separates spaces! Not everything needs to be in `Std` :)

I think we do not need the nested named address, just need the nested module.

First, I can not find a graceful way to do this upgrade. If we keep both `Hash.move` and `hash.move` in std, we will encounter the git or os case insensitive problem.

Second, Some projects have encountered the source files organization problem, like https://github.com/Elements-Studio/starswap-core/tree/master/sources.

Finally, I totally support changing the module name to lowercase, and I still looking for a compatibility solution.

sblackshear (approved):
We discussed this in the Move community meeting and all of the stakeholders expressed broad support. Let's go for it!

One question: is their a plan for fixing the docs as well? I am sure they are full of uppercase names.

language/move-prover/boogie-backend/src/lib.rs
```diff
@@ -37,8 +37,8 @@ const VECTOR_SMT_ARRAY_EXT_THEORY: &[u8] =
 const MULTISET_ARRAY_THEORY: &[u8] = include_bytes!("prelude/multiset-array-theory.bpl");
 const TABLE_ARRAY_THEORY: &[u8] = include_bytes!("prelude/table-array-theory.bpl");
 
-const BCS_MODULE: &str = "0x1::BCS";
-const EVENT_MODULE: &str = "0x1::Event";
+const BCS_MODULE: &str = "0x1::bcs";
```
wrwg, line 40: As you are already here perhaps add a TODO or issue that the absolute addresses need to go away :-/ That's halfway implemented but hasn't gone fully through yet.

language/move-prover/boogie-backend/src/prelude/native.bpl
```diff
@@ -62,27 +62,27 @@ function {:inline} $EmptyVec{{S}}(): Vec ({{T}}) {
     EmptyVec()
 }
 
-procedure {:inline 1} $1_Vector_empty{{S}}() returns (v: Vec ({{T}})) {
+procedure {:inline 1} $1_vector_empty{{S}}() returns (v: Vec ({{T}})) {
```
wrwg, line 65: This here would be then also fixed. I did it for the 'Extensions' alias but not for the 'Std' alias.

wrwg (approved):
Aptos happy to see this forthcoming (@davidiw FYI).

jolestar (approved):

PR #158, opened 2022-05-22
[hardhat-move] Make the plugin more sophisticated

zjhmale:
## Motivation

the `hardhat-move` plugin seems not well constructed (everything is in a single file) before and only support `ethereum` architecture, so I think a more matured hardhat plugin should be there for more developers to join and growth the whole ecosystem.

## Future plan

there still many TODOs need to be done for this plugin:

* a downloader to download `move` executables for each concrete version specified in `hardhat.config.js` after there official releases on Github
* `test` task for move contracts build by `move` and `async-move` architecture
* file cache system for eliminating redundant compiling, contracts will only be compiled when it is a brand new contract or the file content has been changed since last build
* unit test and integration test for this plugin
* maybe it's better to move this plugin out from the move monorepo as a standalone repo for further development and maintainence


language/evm/hardhat-move/src/compile.ts
```diff
@@ -0,0 +1,253 @@
...
+            failedToBuildAll = true;
+        }
+    }
+
+    if (failedToBuildAll) {
+        // TODO: terminate gracefully
+        throw new Error("Failed to build one or more Move packages");
+    }
+
+    if (arch != ARCH_ETHEREUM) return;
```
vgao1996, line 221: If I'm understanding correctly, hardhat is an Ethereum-centered dev environment. Based on this, I'm not sure it makes sense to have arch as a configurable option -- other architectures seem to be out of the scope of this plugin. Could I ask what motivated you to add this?

vgao1996 (commented):
Thanks for putting this up! I haven't had the time to study the test failure, but I'm in favor of the improvements you suggested in general, with a few exceptions (see the individual comments below).

Regarding the future plans you mentioned:
- Executable downloader
  - This is something we should eventually have, but is currently blocked by Move's lack of proper versioning and release schedule.
- Support for non-ethereum architectures
  - I'm not sure about this, see the individual comment below.
- Avoid unneeded recompilation
  - This has been partially implemented in https://github.com/move-language/move/pull/97
- Unit test and integration test for this plugin
  - Functionality wise, this is already covered by a CI test using this plugin to run tests in hardhat-examples, but besides those, I agree that we should have tests that specifically belong to the plugin.
- Separate repo for the plugin
  - Might be possible in the future, but hard to pull off right now.

language/evm/hardhat-move/src/index.ts
```diff
@@ -1,392 +1,99 @@
...
+            {
+                arch: arch,
+                movePath: movePath,
+                artifacts: artifacts,
+                config: config
+            }
+        );
+    });
+
+subtask(TASK_COMPILE_MOVE_RUN_BINARY)
```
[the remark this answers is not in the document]
zjhmale: I think it is only for the purpose of maintenance and reviewing, it will be more clear to split a "big" task into separate sub tasks from my perspective

zjhmale:
will look at the test failures and fix them

language/evm/hardhat-move/src/index.ts
```diff
@@ -1,392 +1,99 @@
...
+    TASK_COMPILE_MOVE_RUN_BINARY,
+} from "./task-names";
+import { compile } from "./compile"
+import { MoveBuild } from "./types";
+import { DEFAULT_ARCH } from "./constants"
+import { locateMoveExecutablePath } from "./executable"
+
+extendConfig((config, userConfig) => {
+    let defaultConfig: Partial<MoveBuild> = {};
+    if (userConfig.move) {
```
zjhmale, line 16: could be a more concise solution to handle empty config

language/evm/hardhat-move/src/compile.ts
```diff
@@ -0,0 +1,253 @@
...
+        let manifestPath = Path.join(path.toString(), "Move.toml");
+        let manifestStats: Fs.Stats = await Fs.promises.stat(manifestPath);
+
+        return manifestStats.isFile();
+    }
+
+    return false;
+}
+
+export async function listMovePackages(contractsPath: Fs.PathLike): Promise<Array<String>> {
```
zjhmale, line 26: make sure if multiple move packages in a single workspace is still supported

zjhmale:
Hi @vgao1996, I have updated this PR, pls review again. Now the hardhat plugin focus on Ethereum architecture only. And I tried to add unit test for this plugin beside the integration tests in `hardhat-example`, since we currently use some experimental API like `fs.promise` which did not comply with `ts-node` properly, it seems impossible to run test with mocha and chai with it, so I'll leave it right now and may fix it in further update.

vgao1996:
@zjhmale thanks for making the update and my apologies for the slow review. I've been quite busy this week but will have a closer look at this in the next few days.

PR #163, opened 2022-05-24
[language server] Async symbolication on file saves

awelc:
## Motivation

This is a refinement for the recently added (https://github.com/move-language/move/pull/147) symbolication support in the language server. The purpose of the original PR (https://github.com/move-language/move/pull/147) was to introduce (and test) the symbolication algorithm itself.

The goal of this PR is to move the needle in terms of user experience - in the original PR, symbolication was triggered only once at the language server start. In this PR we trigger symbolication asynchronously (in a separate thread) each time a file is saved so that the most recent information about symbols is available to the IDE.

One reason to use only a single symbolicator thread is to throttle the number of times symbolication (which is relatively expensive) is triggered. At this point it's not a big issue as it's triggered of file save operation, but could become one when we manage to get the compiler operate on in-memory files, and when we will be triggering symbolication on cadence closer to key strokes than to file saves.

Yes

### DIFF
--- a/language/move-analyzer/src/symbols.rs
+++ b/language/move-analyzer/src/symbols.rs
@@ -49,14 +49,25 @@
 use crate::context::Context;
 use anyhow::Result;
 use codespan_reporting::files::{Files, SimpleFiles};
+<<<<<<< HEAD
 use im::ordmap::OrdMap;
+=======
+>>>>>>> 8a48fe92a (Added async symbolication on file saves)
 use lsp_server::{Request, RequestId};
 use lsp_types::{GotoDefinitionParams, Location, Position, Range, ReferenceParams};
 use std::{
     cmp,
+<<<<<<< HEAD
     collections::{BTreeMap, BTreeSet},
     fs,
     path::{Path, PathBuf},
+=======
+    collections::{BTreeMap, BTreeSet, VecDeque},
+    fs,
+    path::{Path, PathBuf},
+    sync::{Arc, Condvar, Mutex},
+    thread,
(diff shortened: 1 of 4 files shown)


language/move-analyzer/src/symbols.rs
```diff
@@ -168,6 +174,91 @@ pub struct Symbols {
...
+                    // hold the lock only as long as it takes to get the data, rather than through
+                    // the whole symbolication process (hence a separate scope here)
+                    let mut symbolicate = mtx.lock().unwrap();
+                    if *symbolicate == RunnerState::Quit {
+                        break;
+                    }
+                    if *symbolicate == RunnerState::Run {
+                        *symbolicate = RunnerState::Wait;
+                        true
+                    } else {
```
tnowacki, line 216: nit 
```
match *symbolicate {
  RunnerState::Quit => break,
  RunnerState::Run => { *symbolicate = RunnerState::Wait; true }
  RunnerState::Wait > ...,
}
```

language/move-analyzer/src/symbols.rs
```diff
@@ -168,6 +174,91 @@ pub struct Symbols {
...
+                        // wait for next request
+                        symbolicate = cvar.wait(symbolicate).unwrap();
+                        if *symbolicate == RunnerState::Quit {
+                            break;
+                        }
+                        if *symbolicate == RunnerState::Run {
+                            *symbolicate = RunnerState::Wait;
+                            true
+                        } else {
```
tnowacki, line 225: same here, should just match

tnowacki (approved):
Some nits, but otherwise looks fine to me!

PR #164, opened 2022-05-25
[scripts] Git installation in dev_setup.sh

villesundell:
## Motivation

Although most of the development setups have Git installed (since development happens in a Git repository), this assumption does not hold for all the cases: for example in the case of Docker's `ADD .` where local files are copied into a container without Git.

However, Git is required by the Move CLI (when using packages with Git dependencies), although this is undocumented.

Hence, the development setup script should have an option for installing it (although it's disabled by default).

Summary: **This PR adds -g switch to dev_setup.sh for installing Git.**

Yes.

## Test Plan

```
docker build -t move/cli -f docker/move-cli/Dockerfile .
```


sblackshear (approved):
Thanks!!

PR #165, opened 2022-05-26
[language server] Added support for displaying compiler diagnostics

awelc:
## Motivation

This another refinement of recently added support for more advanced language server features (https://github.com/move-language/move/pull/147). This PR adds support for passing compiler diagnostics from the language server to the IDE (so that it is more clear why advanced features may not work or work in a limited fashion when the compilation fails).

This PR also fixes an ongoing problem with auto-completion where the loop where lexing problems could lead to a loop collecting identifiers from a file to run forever. 

Finally, supporting lexer diagnostics, particularly in the limited way they were supported, does not make much sense now that we have compiler diagnostics so the lexer ones have been removed.

Yes

### DIFF
--- a/language/move-compiler/src/diagnostics/mod.rs
+++ b/language/move-compiler/src/diagnostics/mod.rs
@@ -145,7 +145,7 @@ fn convert_loc(file_mapping: &FileMapping, loc: Loc) -> (FileId, Range<usize>) {
     (id, range)
 }
 
-pub fn render_diagnostic(
+fn render_diagnostic(
     file_mapping: &FileMapping,
     diag: Diagnostic,
 ) -> csr::diagnostic::Diagnostic<FileId> {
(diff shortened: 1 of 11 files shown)


language/move-compiler/src/diagnostics/mod.rs
```diff
@@ -145,7 +145,7 @@ fn convert_loc(file_mapping: &FileMapping, loc: Loc) -> (FileId, Range<usize>) {
     (id, range)
 }
 
-fn render_diagnostic(
+pub fn render_diagnostic(
```
tnowacki, line 148: Why was this made public?

awelc: I was using it in the language server before I discovered `into_codespan_format` and forgot to change it back. Fixed now!

emmazzz (approved):

PR #173, opened 2022-05-30
[prover] Implementation of Tables

wrwg:
This PR provides the full implementation of tables in the Move Prover.

- Adds a new source `Table.spec.move` which contains specification related declarations. The capability to separate specs into different files has been added a while ago but never used. In this spec, most table functions are declared intrinsic.
- Extends the Boogie prelude by a theory for tables and the native implementation of the Table module. The most important aspect here is that we use an encoding of keys `enc: K -> int`, which is axiomatized such that
  `equal(k1, k2) <=> enc(k1) == enc(k2)`.  This way we solve the problem that structured keys do not have extensional equality. Moreover, the int encoding comes handy to implement the memory model.
- Extends the memory model impl of `&mut` to tables. We need to support borrow paths with mixed field, vector, and table edges. Fortunately, since table keys are encoded as int, the vector treatment could be generalized for tables.
- Adds tests in `verify_table.move` in the prover test suite.
- Extends the boogie wrapper to display tables on verification errors.
- Extends the model to support address aliases. Specifically, the model is now capable of determining where `Std` and `Extensions` aliases are mapped to. This allows to generate boogie code with the correct (absolute) addresses.
- A few more drive-by changes.

## Motivation

Make prover ready for Move applications which use tables.

Yes

## Test Plan

Added new verification tests.


jolestar:
@whst please help to review

wrwg:
@junkil-park @rahxephon89 @DavidLDill I'm off the grid for quite a while in 24 hours from now, so if we want to land this before, not much time left.

language/move-prover/tests/sources/functional/verify_table.move
```diff
@@ -0,0 +1,150 @@
...
+        aborts_if k1 == k2;
+    }
+
+    fun remove(): Table<u8, u64> {
+        let t = add();
+        Table::remove(&mut t, 2);
+        t
+    }
+    spec remove {
+        ensures spec_contains(result, 1) && spec_contains(result, 3);
```
junkil-park, line 50: I am curious if `!spec_contains(result, 3)` will be provable? I can figure it by testing it after this PR lands.

junkil-park:
Great to see the table support in Prover!

junkil-park (approved):

language/move-prover/tests/sources/functional/verify_table.move
wrwg: Yes it should be.

PR #177, opened 2022-06-01
[move-package] Change all yaml files to be TOML

gregnazario:
## Motivation

Let's stick with 1 parser for Move.  I picked TOML, since it was the one that was already being manually manipulated for the Move package system.

yes

## Test Plan

Ran tests, though the solidity ones didn't like me

## Related issue
https://github.com/move-language/move/issues/176

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1172,12 +1172,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "31ad93652f40969dead8d4bf897a41e9462095152eb21c56e5830537e41179dd"
 
 [[package]]
-name = "dtoa"
-version = "0.4.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "88d7ed2934d741c6b37e33e3832298e8850b53fd2d2bea03873375596c7cea4e"
-
-[[package]]
 name = "duct"
 version = "0.13.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


wrwg (approved):
Agreed sticking to one format is much better,

Since the build tree is generated, this should not introduce a breaking change, so lets land it.

villesundell:
Hello! :wave: The problem doesn't seem to be the EVM tests alone, since the CLI Docker image test (last in the list) also fails with "`Error: values must be emitted before tables`" (https://github.com/move-language/move/runs/6695042916). The same test seems to be successful in other PRs, such as this one: https://github.com/move-language/move/runs/6683370096.

gregnazario:
> Hello! 👋 The problem doesn't seem to be the EVM tests alone, since the CLI Docker image test (last in the list) also fails with "`Error: values must be emitted before tables`" (https://github.com/move-language/move/runs/6695042916). The same test seems to be successful in other PRs, such as this one: https://github.com/move-language/move/runs/6683370096. (UPDATE: I was able to reproduce this issue by checking out gregnazario:toml-everywhere and following the [quickstart sequence](https://github.com/move-language/move#quickstart)).

I'll take a look at it, it's just very strange that changing these formats seems to have caused a different issue.

wrwg:
> Error: values must be emitted before table

This is a serious issue with toml. May need to move fields which represent collections at the end of the underlying Rust struct definition.

gregnazario:
Closing, this won't be changed in time.

PR #178, opened 2022-06-03
[Move] Make compilation kill the process on failure/warning optional

awelc:
## Motivation

At this point, building a package containing either errors or just warnings results in the whole process being killed, which makes it impossible for adapters to choose to publish modules with warnings only. This PR addresses the problem by making  process exiting upon encountering failure/warning optional.

Yes

### DIFF
--- a/language/benchmarks/src/move_vm.rs
+++ b/language/benchmarks/src/move_vm.rs
@@ -41,7 +41,7 @@ fn compile_modules() -> Vec<CompiledModule> {
         vec![],
         move_stdlib::move_stdlib_named_addresses(),
     )
-    .build_and_report()
+    .build_and_report(true)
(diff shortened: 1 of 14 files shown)


awelc:
> This is the point of the `*_and_report` API end points. The non-exiting versions exist already without the suffix, e.g. `check()` and `build()`.

Makes sense. Also, upon closer inspection, I think going through the path of modifying `report_diagnostics` signature caused the changes to spread too far.

> Is there something else missing here?

Diagnostics reporting is a bit opaque and seems to need an extra entry point to avoid process exiting after reporting. 

It seemed cleaner to start a new PR rather than continuing with this one, so here is another tack: https://github.com/move-language/move/pull/184 (and I am closing this one).

PR #182, opened 2022-06-05
[file-format] Deprecate script visibility for an entry point modifier

tnowacki:
- Replaces Script visibility with an is_entry marker for functions
- For backwards compatibility, Script visibility now means public + is_entry
- The visibility is removed from V5+

## Motivation

- We want to separate out the notion of entry points from visibility
- This gives adapters more flexibility 
- Additionally, it allows module writers to have more expressiveness. You can even declare a private entry function, which enables new expressiveness, as you can ensure the logic is only performed at the start of the transaction (assuming the module itself does not call the function)

## Test Plan

- Updated tests

## Followup 

- Add an entry modifier to the source language and deprecate `public` script


wrwg (approved):
Glad to see that!

PR #187, opened 2022-06-07
Ethereum ABI for Move

rahxephon89:
## Motivation

Initial step to support generating Ethereum ABI for Move

Yes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2427,6 +2427,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "move-ethereum-abi"
+version = "0.1.0"
+dependencies = [
+ "serde 1.0.130",
+ "serde_json",
+]
+
+[[package]]
 name = "move-explain"
 version = "0.1.0"
 dependencies = [
@@ -2798,6 +2806,7 @@ dependencies = [
  "move-command-line-common",
  "move-compiler",
  "move-core-types",
+ "move-ethereum-abi",
  "move-ir-types",
  "move-model",
  "move-prover-boogie-backend",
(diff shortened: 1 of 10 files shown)


runtian-zhou:
Just to add a bit more context here. The goal is to refactor the ethereum abi generator such that the same generator could be used for regular Move modules instead of just Move modules that are intended to be compiled to EVM. Once we have such generator, we can attach the generated ethereum abi as part of the Move module's metadata section and use the generated abi to serve a ethereum style query of global states and events. This would allow us to have a truly ERC compatible smart contract written and deployed in the Move world. cc @jolestar

language/evm/move-abi/src/abi_move_type.rs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) The Diem Core Contributors
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+
+use crate::abi_signature_type::ABIJsonSignature;
+
+#[derive(Serialize, Deserialize)]
```
runtian-zhou, line 10: Maybe add the key that we intend to use in the metadata section here as well.

runtian-zhou (commented):
I think `move-ethereum-abi` may be a better crate name. `move-abi` to me is a more generic name and could lead to confusion here.

runtian-zhou:
I'm a bit concerned whether the added `serde` flag could affect how we serialize/deserialize data. I'm fine landing the PR as is but hopefully we can have some round trip tests to make sure the serialization and deserialization is done properly.

runtian-zhou (approved):
Looks good to me!

PR #189, opened 2022-06-08
[Move analyzer] Added type-on-hover

awelc:
## Motivation

This is another refinement for the recently added (https://github.com/move-language/move/pull/147) symbolication support in the language server. This PR adds the type-on-hover functionality.

The basic idea is to store type information with the symbolication data and be able to use it when request for a type of a given identifier is issued.

We cannot use `Type` for this exclusively as it does not directly represent function signatures, so we special-case the latter. The debugging print for `Type` wasn't producing the best output for displaying info in the IDE so we implemented the `Display` trait for `Type_` to handle this.

Yes

### DIFF
--- a/language/move-analyzer/src/symbols.rs
+++ b/language/move-analyzer/src/symbols.rs
@@ -56,11 +56,14 @@ use codespan_reporting::files::SimpleFiles;
 use crossbeam::channel::Sender;
 use im::ordmap::OrdMap;
 use lsp_server::{Request, RequestId};
-use lsp_types::{Diagnostic, GotoDefinitionParams, Location, Position, Range, ReferenceParams};
+use lsp_types::{
+    Diagnostic, GotoDefinitionParams, Hover, HoverContents, HoverParams, LanguageString, Location,
+    MarkedString, Position, Range, ReferenceParams,
+};
 use std::{
     cmp,
     collections::{BTreeMap, BTreeSet, HashMap},
-    fs,
+    fmt, fs,
     path::{Path, PathBuf},
     sync::{Arc, Condvar, Mutex},
     thread,
@@ -71,9 +74,9 @@ use url::Url;
 use move_command_line_common::files::FileHash;
 use move_compiler::{
     expansion::ast::{Fields, ModuleIdent, ModuleIdent_},
-    naming::ast::{StructDefinition, StructFields, TParam, Type, TypeName_, Type_},
+    naming::ast::{addr_to_string, StructDefinition, StructFields, TParam, Type, TypeName_, Type_},
     parser::ast::StructName,
-    shared::Identifier,
+    shared::{format_comma, Identifier},
     typing::ast::{
         BuiltinFunction_, Exp, ExpListItem, Function, FunctionBody_, LValue, LValueList, LValue_,
         ModuleCall, ModuleDefinition, SequenceItem, SequenceItem_, UnannotatedExp_,
@@ -108,15 +111,32 @@ struct UseLoc {
     col_end: u32,
 }
 
+/// Information about a type of an identifier. The reason we need an additional enum is that there
+/// is not direct representation of a function type in the Type enum.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum IdentType {
+    RegularType(Type),
+    FunctionType(
+        ModuleIdent_, /* defining module */
+        Symbol,       /* name */
+        Vec<Type>,    /* type args */
+        Vec<Type>,    /* args */
+        Type,         /* ret */
+        Vec<Type>,    /* acquires */
+    ),
+}
+
 /// Information about both the use identifier (source file is specified wherever an instance of this
 /// struct is used) and the definition identifier
-#[derive(Debug, Clone, Eq, Copy)]
+#[derive(Debug, Clone, Eq)]
 pub struct UseDef {
     /// Column where the (use) identifier location starts on a given line (use this field for
     /// sorting uses on the line)
     col_start: u32,
     /// Column where the (use) identifier location ends on a given line
     col_end: u32,
+    /// Type of the (use) identifier
+    use_type: IdentType,
     /// Location of the definition
     def_loc: DefLoc,
 }
@@ -159,7 +179,7 @@ pub struct Symbolicator {
     /// Contains type params where relevant (e.g. when processing function definition)
     type_params: BTreeMap<Symbol, DefLoc>,
     /// Current processed module (always set before module processing starts)
-    current_mod: Option<ModuleIdent_>,
+    current_mod: Option<ModuleIdent>,
 }
 
 /// Maps a line number to a list of use-def pairs on a given line (use-def set is sorted by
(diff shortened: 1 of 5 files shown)


language/move-compiler/src/naming/ast.rs
```diff
@@ -563,6 +563,80 @@ impl fmt::Display for TypeName_ {
...
+                            "{}::{}::{}<{}>",
+                            addr,
+                            module_ident.module.value(),
+                            struct_name,
+                            format_comma(ss)
+                        )
+                    }
+                }
+            },
+            Type_::Var(tv) => write!(f, "var type ({})", tv.0),
```
tnowacki, line 625: Do you expect this to appear on hover?

language/move-compiler/src/naming/ast.rs
```diff
@@ -563,6 +563,80 @@ impl fmt::Display for TypeName_ {
     }
 }
 
+impl fmt::Display for Type_ {
+    fn fmt(&self, f: &mut fmt::Formatter) -> std::fmt::Result {
```
tnowacki, line 567: Type_ intentionally shouldn't have display. There is a lot of contextual information in here that you probably don't want to display, Like the abilities for a type parameter :) 

If you want this, I would rework `core::error_format_impl_`

awelc: I am on the phone right now but I am trying to get as much info as I can so that I can start working on it at an arbitrary time. What would be your preferred solution then? I can always just turn it into type_to_string function (and perhaps move it to symbolicator itself)?

tnowacki: I think you probably want to re-use as much as you can from `core::error_format`? Or you really want/need printing of types that differs from the compiler?

awelc: I will take a look at `core::error_format` when I am back at my laptop, but since we don't want this to be an implementation of the `Display` trait, I am not sure if it's possible to package it as part of `core::error_format` or in some different way and just reuse parts of `core::error_format`

tnowacki: Not sure why you need Display, but if you really do, you can always do something like `MyTypeWrapper(Type)`and `impl Display for MyTypeWrapper { ... /* use core::error_format */ ... }`

The reason `Type` doesn't implement display is to avoid accidents (as you need the substitution map to display non expanded types)

awelc: > Not sure why you need Display, but if you really do, you can always do something like `MyTypeWrapper(Type)`and `impl Display for MyTypeWrapper { ... /* use core::error_format */ ... }`

I don't really need it, I simply though it would be the cleanest solution considering that some other AST nodes implement `Display` as well. I will simply turn this into `type_to_string`.

tnowacki: Do we need another type_to_string though? That is what `error_format` is. The name is simply because it assumes that string will be used in an error

awelc: OK, now that I had a chance to actually take a look at the code, I understand the whole deal with `error_format` is, and I am not sure if I want to keep the exact format for type-on-hover (I also don't want to re-work `error_format` as this would have a further reaching consequences for how compiler errors are displayed). An example of what I think is unnecessary in the `error_format` output for the type-on-hover application is how it displays `TypeName_::ModuleType` - while I understand that one would find it in some cases useful to have both numeric address and the name (e.g., `(some_module=0xCAFE)::SomeStruct`) it feels a bit too verbose (and not super intuitive for me) for type-on-hover  where `some_module::SomeStruct` feels totally sufficient. Admittedly I was at least a bit inspired by how IntelliJ plugin displays these.

That being said the differences are fairly minor, so there are two solutions, and it would be great to agree on one (I don't have a super strong preference) before I do another edit:
- use `error_format` and live with some potential minor clutter
- have a customized format for type-on-hover (we don't have to add anything to the compiler core for it - it can become a part of displaying `IdentType` in the symbolicator itself)

language/move-compiler/src/naming/ast.rs
awelc: I do not but since I was writing a generic(ish) implementation of the `Display` trait I wanted to handle it at least semi-gracefully. If we specialize it more towards typ-on-hover, I'd probably just make it into "invalid type".

language/move-compiler/src/naming/ast.rs
tnowacki: The verbosity is necessary for errors since you can construct a state where it says `a::m::T` is not compatible with `a::m::T`, and it is because the addresses are different, e.g. `(a=0x1)::m::T` vs `(a=0x2)::m::T`. 

Happy to make the error_format a bit more robust/configurable, but it also might make sense for you to do your own thing on the symbols side

awelc: I went the route of doing this on the symbols side. When thinking about it some more, I figured it would actually make sense to make the IDE display more susceptible to user feedback by keeping it separate from compiler error messages (put a comment along those lines in the code as well).

language/move-analyzer/src/symbols.rs
```diff
@@ -191,6 +212,125 @@ pub struct SymbolicatorRunner {
...
+                // verbosity that is not needed here.
+                //
+                // It also seems like a reasonable idea to be able to tune user experience in the
+                // IDE independently on how compiler error messages are generated.
+                write!(f, "{}", type_to_ide_string(t))
+            }
+            Self::FunctionType(mod_ident, name, type_args, args, ret, acquires) => {
+                let type_args_str = if !type_args.is_empty() {
+                    let mut s = "<".to_string();
+                    s.push_str(type_list_to_ide_string(type_args).as_str());
```
tnowacki, line 231: You can also do `s.push_str(&type_list_to_ide_string(type_args))` or `s += &type_list_to_ide_string(type_args)` if you want. But whatever floats ya boat :)

language/move-analyzer/src/symbols.rs
```diff
@@ -191,6 +212,125 @@ pub struct SymbolicatorRunner {
...
+            "&{} {}",
+            if *mut_ { "mut" } else { "" },
+            type_to_ide_string(s)
+        ),
+        Type_::Param(tp) => {
+            if tp.abilities.is_empty() {
+                format!("{}", tp.user_specified_name)
+            } else {
+                format!(
+                    "{}: {}",
```
tnowacki, line 278: Do you really want to display these? Feel a bit strange

language/move-analyzer/src/symbols.rs
```diff
@@ -191,6 +212,125 @@ pub struct SymbolicatorRunner {
...
+                    if ss.is_empty() {
+                        "".to_string()
+                    } else {
+                        format!("<{}>", type_list_to_ide_string(ss))
+                    }
+                )
+            }
+        },
+        Type_::Anything => "_".to_string(),
+        _ => "invalid type".to_string(),
```
tnowacki, line 314: Should you panic here? or something maybe indicate to the user its an internal error?

tnowacki: And a nit, I try to avoid `_` patterns here. It would probably be helpful to list out which cases you don't expect

tnowacki (approved):

language/move-analyzer/src/symbols.rs
awelc: I was wondering about that myself, particularly that we don't do it for struct types (for clarity but also because in some cases we simply don't know). I will remove it for now - we can always add it back if IDE extension users complain.

language/move-analyzer/src/symbols.rs
awelc: > Should you panic here? or something maybe indicate to the user its an internal error?

The whole language server is implemented such that it avoids panicking as it would exit the server process. Feels like a better user experience to have some functionality missing (ideally being explicit about it) than dealing with the server crashes with the IDE does not handle well.

tnowacki: I think at the very least, it should be a bit more clear to a dev that this is _not_ an intended type to have displayed

awelc: > I think at the very least, it should be a bit more clear to a dev that this is _not_ an intended type to have displayed

I have changed it to:
```
        Type_::Var(_) => "invalid type (var)".to_string(),
        Type_::UnresolvedError => "invalid type (unresolved)".to_string(),
```

PR #190, opened 2022-06-09
[move-vm] Fix compatibility issues for new entry modifier

tnowacki:
- Fix issues where rules were removed for V4

## Motivation

- Preserving these is a pain, but makes upgrades easier
- The changes in #182 made it so you could get inconsistent results on V4 versions of the file format. Where the old software would say no, and the new software would say yes. 

## Test Plan

- Bit of a pain, had to manually construct V4 modules

### DIFF
--- a/language/move-bytecode-verifier/src/dependencies.rs
+++ b/language/move-bytecode-verifier/src/dependencies.rs
@@ -4,29 +4,34 @@
 
 //! This module contains verification of usage of dependencies for modules and scripts.
 use move_binary_format::{
-    access::ModuleAccess,
+    access::{ModuleAccess, ScriptAccess},
     binary_views::BinaryIndexedView,
     errors::{verification_error, Location, PartialVMError, PartialVMResult, VMResult},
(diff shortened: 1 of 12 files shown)


language/move-bytecode-verifier/src/dependencies.rs
```diff
@@ -473,3 +500,72 @@ fn compare_structs(
...
+        match (current_is_entry, script_functions.contains(&fhandle_idx)) {
+            (true, true) => (),
+            (_, true) => {
+                return Err(PartialVMError::new(
+                    StatusCode::CALLED_SCRIPT_VISIBLE_FROM_NON_SCRIPT_VISIBLE,
+                )
+                .at_code_offset(fdef_idx, idx)
+                .with_message(
+                    "script-visible functions can only be called from scripts or other \
+                    script-visibile functions"
```
sblackshear, line 563:
```suggestion
                    script-visible functions"
```

sblackshear (approved):

PR #191, opened 2022-06-10
[book] Switch modules to lower snake case

tnowacki:
- Updated stdlib usage
- Updated examples

## Motivation

- docs are good

## Test Plan

- eyes


damirka (approved):
Lgtm!

PR #192, opened 2022-06-10
[book] Update script visibility docs

tnowacki:
- Updated docs to reflect new `entry` modifier

## Motivation

- docs are good

## Test Plan

- eyes

### DIFF
--- a/language/documentation/book/src/functions.md
+++ b/language/documentation/book/src/functions.md
@@ -43,7 +43,7 @@ script {
 }
 ```
 
-To allow access from other modules or from scripts, the function must be declared `public`, `public(friend)`, or `public(script)`.
+To allow access from other modules or from scripts, the function must be declared `public` or `public(friend)`.
 
 #### `public` visibility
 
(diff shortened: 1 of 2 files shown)


language/documentation/book/src/functions.md
```diff
@@ -111,39 +111,70 @@ script {
...
+}
+
+script {
+    fun calls_m_foo(): u64 {
+        0x42::m::foo() // valid!
+    }
+}
+```
+
+Even internal functions can be marked as `entry`! This let's you guarantee that the function is called only at the beginning of execution (assuming you do not call it elsewhere in your module)
```
emmazzz, line 151: "let's" -> "lets"

language/documentation/book/src/functions.md
```diff
@@ -43,7 +43,7 @@ script {
 }
 ```
 
-To allow access from other modules or from scripts, the function must be declared `public`, `public(friend)`, or `public(script)`.
+To allow access from other modules or from scripts, the function must be declared `public` or `public(friend)`.
```
emmazzz, line 46: `public entry` also can be called from other modules right?

emmazzz (approved):

language/documentation/book/src/functions.md
tnowacki: `entry` is not a part of visibility, so would be more confusing to mention it than to not. 

You can have `public entry` and `public(friend) entry` and even (private/internal) `entry`

PR #194, opened 2022-06-13
[async-move] Return extension context in async move execution api.

runtian-zhou:
## Motivation

Return the native context in async move's api for better composibility. The goal here is to integrate table extension to async move and the current api is not capable of doing that as native context gets consumed.

Yes

## Test Plan

`cargo test`


rahxephon89 (approved):

wrwg:
Sorry for not reviewing, but I'm traveling in Europe until end of June with spotty connection.

PR #195, opened 2022-06-14
[move-lang] Fix ControlFlowGraph trait

fEst1ck:
Fix the signature of `next_block` and add documentation.

## Motivation

The `next_block` method returns the block id of the next block to traverse. The current return type however is `CodeOffset`, which is rather confusing.

Yes.

## Test Plan

Since `pub type BlockId = CodeOffset;`, this change won't affect execution.


tnowacki (approved):
Thanks!

Issue #196, opened 2022-06-14
[Bug]panic in debug mode when invoking status_type() if status is VMStatus::Error{ StatusCode::OUT_OF_GAS }


JerryKwan:
# 🐛 Bug

panic occurred when invoking status_type() in debug mode if status is VMStatus::Error{StatusCode::OUT_OF_GAS}
The backtrace is something like the following

## To reproduce

**Code snippet to reproduce**
```rust
use crate::{
    vm_status::{VMStatus, StatusCode},
};

#[test]
fn test_stats_code() {
    let vm_status = VMStatus::Error(StatusCode::OUT_OF_GAS);
    let code = vm_status.status_code();
    assert_eq!(code, StatusCode::OUT_OF_GAS);
}
```

**Stack trace/error message**
```
ERROR - details: panicked at 'assertion failed: code.status_type() != StatusType::Execution', /root/.cargo/git/checkouts/move-ae0d26a86327ae96/4735fa5/language/move-core/types/src/vm_status.rs:132:17
details: panicked at 'assertion failed: code.status_type() != StatusType::Execution', /root/.cargo/git/checkouts/move-ae0d26a86327ae96/4735fa5/language/move-core/types/src/vm_status.rs:132:17
2022-06-13T07:19:57.761619499-07:00 ERROR - backtrace:    0:     0x5603d1dbd07a - backtrace::backtrace::libunwind::trace::h75698078811e3944
                               at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.65/src/backtrace/libunwind.rs:93:5
                           backtrace::backtrace::trace_unsynchronized::hd57af2b933e855e0
                               at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.65/src/backtrace/mod.rs:66:5
   1:     0x5603d1dbcff2 - backtrace::backtrace::trace::hf602886d933bcd64
                               at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.65/src/backtrace/mod.rs:53:14
   2:     0x5603d1e2a7c6 - backtrace::capture::Backtrace::create::h5d0bd96752a1217b
                               at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.65/src/capture.rs:176:9
   3:     0x5603d1e2a6fd - backtrace::capture::Backtrace::new::h72069796affebd67
[...]

PR #197, opened 2022-06-14
fix the wrong logic in VMStatus::status_code()

JerryKwan:
fix the wrong logic in VMStatus::status_code()

## Motivation

(Write your answer here.)

## Test Plan

unit test case is attached

### DIFF
--- a/language/move-core/types/src/vm_status.rs
+++ b/language/move-core/types/src/vm_status.rs
@@ -130,7 +130,7 @@ impl VMStatus {
                 let code = *code;
                 debug_assert!(code != StatusCode::EXECUTED);
                 debug_assert!(code != StatusCode::ABORTED);
-                debug_assert!(code.status_type() != StatusType::Execution);
+                debug_assert!(code.status_type() == StatusType::Execution);
                 code
             }
         }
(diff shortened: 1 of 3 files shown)


language/move-core/types/src/vm_status.rs
```diff
@@ -130,7 +130,7 @@ impl VMStatus {
                 let code = *code;
                 debug_assert!(code != StatusCode::EXECUTED);
                 debug_assert!(code != StatusCode::ABORTED);
-                debug_assert!(code.status_type() != StatusType::Execution);
+                debug_assert!(code.status_type() == StatusType::Execution);
```
tnowacki, line 133: This debug assert is correct, because it should be an `ExecutionFailure` if it has this status type. Can you repro the failure?

JerryKwan: @tnowacki 
Please see https://github.com/move-language/move/issues/196
And regarding the place where VMStatus::Error(StatusCode::OUT_OF_GAS) generated, GasStatus::deduct_gas() will return PartialVMError::new(StatusCode::OUT_OF_GAS), and PartialVMError::finish(self, location: Location) -> VMError, and VMError::into_vm_status will return VMStatus::Error(major_status) other than VMStatus::ExecutionFailure if location is Location::Undefined. But most of the time VMRuntime::execute_function_impl() use Location::Undefined
Wish this can help

JerryKwan:
@tnowacki How about we temporarily comment out the status_type check in VMStatus::status_code()?

tnowacki:
Thanks!

PR #199, opened 2022-06-14
move-bytecode-verifier: remove the Clone bound for the error type in …

bmwill:
…the TransferFunctions trait

Error types generally do not implement clone.

### DIFF
--- a/language/move-bytecode-verifier/src/absint.rs
+++ b/language/move-bytecode-verifier/src/absint.rs
@@ -50,7 +50,7 @@ pub type InvariantMap<State, AnalysisError> =
 /// Auxiliary data can be stored in self.
 pub trait TransferFunctions {
     type State: AbstractDomain;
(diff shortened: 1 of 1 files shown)


tnowacki:
Do we need any constraint at all? Why was it there?

bmwill:
I don't really know why it was originally there. I just know that at least bounding it to the Error type makes more sense than clone

tnowacki:
I kind of feel like it shouldn't be bounded by anything if it works? If I want to make my Error type `()`, thats my own prerogative :P (does `()` implement `Error`??)

I think the `Clone` was probably there from the days of join errors *shudders*. Not really important but it is curious. Maybe @sblackshear remembers?

bmwill:
That's fine, i'll just remove the bound completely.

tnowacki (approved):

PR #200, opened 2022-06-15
[Move] Fixed symbolication of modules stored in the same file

awelc:
## Motivation

Symbolication support for the language server incorrectly assumed that each module is stored in a separate file. This PR fixes this problem.

Yes


awelc:
> I don't really know whats going on, but lgtm :)

A bit of extra explanation then. Previously I assumed that there can only be one module per file which translated to an implementation with a 1:1 mapping between module id and the file path. Consequently there was no way to have two modules (and their symbols) associated with a single file path - information about references, types, etc. was being reported for only one module in a multi-module file. This PR fixes that by "flattening" all the symbols for multiple modules from the same file into a single date structure (key-d on file path) rather than going through a dance of first mapping file path to module id and then module id to the  module symbols.

PR #206, opened 2022-06-17
[ethereum_abi_annotator] Implement an adapter to convert between ethe…

runtian-zhou:
…reum encoded data and move encoded data.

## Motivation

Implement an adapter to allow interacting with Move world with ethereum style api. There are two major components:
1. An event annotator. The goal here is to convert the event emitted by MoveVM into log format used in the ethereum ecosystem, including both the encoding and the topic hash.
2. A call data annotator. The goal is to convert an eth call data into parameters into invokations into MoveVM's session and also a function to encode the Move value into ethereum's encoding.

I'll add an example later on how to use those apis. The apis needs a lot more tuning on error propagation. This PR just serves as a proof of concept for now.

Yes

## Test Plan

Will add an example demonstrating how the annotator should work.

### DIFF
--- a/language/evm/move-ethereum-annotator/src/events.rs
+++ b/language/evm/move-ethereum-annotator/src/events.rs
@@ -1,0 +1,95 @@
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::{
+    util::{convert_move_value, ModuleCache},
+    LRU_CACHE_SIZE,
+};
+use anyhow::{anyhow, bail, Result};
+use ethabi::{encode, RawLog};
+use ethereum_types::H256;
+use lru::LruCache;
+use move_core_types::value::MoveValue;
+use move_core_types::{
+    effects::Event, language_storage::TypeTag, resolver::MoveResolver, value::MoveTypeLayout,
+};
+use move_vm_runtime::move_vm::MoveVM;
+
+pub struct MoveEventAnnotator<R: MoveResolver> {
+    resolver: R,
+    abi_cache: ModuleCache<R>,
+    layout_cache: LruCache<TypeTag, MoveTypeLayout>,
+}
+
+impl<R: MoveResolver + Clone> MoveEventAnnotator<R> {
+    pub fn new(resolver: R) -> Self {
+        Self {
+            resolver: resolver.clone(),
+            abi_cache: ModuleCache::new(resolver),
+            layout_cache: LruCache::new(LRU_CACHE_SIZE),
+        }
+    }
+
+    pub fn annotate_move_event(&mut self, event: Event) -> Result<RawLog> {
+        let struct_tag = {
+            use TypeTag::*;
+            match &event.2 {
+                Struct(s) => s,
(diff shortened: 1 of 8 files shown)


language/evm/move-ethereum-annotator/src/events.rs
```diff
@@ -0,0 +1,95 @@
...
+                Struct(s) => s,
+                Bool | U8 | U64 | U128 | Address | Signer | Vector(_) => {
+                    bail!("Unsupported primitive types")
+                }
+            }
+        };
+
+        let move_value = {
+            if !self.layout_cache.contains(&event.2) {
+                let move_vm = MoveVM::new(vec![].into_iter()).unwrap();
```
jolestar, line 46: Run a VM instance to get MoveTypeLayout may be too heavy?

language/evm/move-ethereum-annotator/src/util.rs
```diff
@@ -0,0 +1,180 @@
...
+    value::MoveValue,
+};
+use move_ethereum_abi::abi_move_type::{ABIMoveSignature, ABI_ETHER_MOVE_KEY};
+use std::{collections::HashMap, convert::TryFrom, str::FromStr};
+
+pub(crate) struct ModuleCache<R: ModuleResolver> {
+    resolver: R,
+    cache: LruCache<ModuleId, CacheEntry>,
+    // TODO: Review the selector map's value type. Will function name be sufficient here or could there be hash collision issues?
+    selector_cache: LruCache<[u8; 4], String>,
```
jolestar, line 24: I think we should consider the hash collision issues. If we define selector_cache as LruCache<[u8; 4], Vec<String>> can resolve this?

jolestar (commented):
I like this feature. Can you give an example of how we can use it?

language/evm/move-ethereum-annotator/src/events.rs
runtian-zhou: Yea this is a bit unfortunate. I need sth that could convert `StructTag` to `MoveTypeLayout`. I picked MoveVM jsut because it's capabale of doing that. I think this should be fine as the only thing here is the resolver.

language/evm/move-ethereum-annotator/src/util.rs
runtian-zhou: How a `Vec` here could help the issue? The attack vector I have on top of my mind is following: 
module A has an api that has signature `Foo(arg1, arg2)` and module B has an api that has signature `Bar(arg1, arg2)` where the selector of `Foo` and `Bar` collides (it's easy to come up with sth like that as we only take the top 4 bytes for the selector). Then during the transaction, when we invoke `A::Foo` it could resolve to `A::Bar` because of the collision. Maybe a selector cache per module will resolve this issue (but could explode the cache size potentially).

jolestar: if use Vec<String> to keep the function name, if the resolved result `A::Bar` does not exist in the A module, we should try to fetch the module again, and mapping short_signature with `["Foo", "Bar"].

Maybe a selector cache per module is more suitable.

PR #207, opened 2022-06-17
[deprecate] Move non-std stdlib files to nursery

tnowacki:
- As mentioned in the weekly meeting, moving some files to the nursery as a temporary solution for deprecating non-std stdlib modules

## Motivation

- make the stdlib more standard and more universal across deployments

## Test Plan

- ran tests


sblackshear (approved):
Approving now that we've gotten anyone on the same page in the community meeting.

Issue #204, opened 2022-06-17
[Feature Request] Publish move-analyzer binaries for different platforms


yubing744:
# 🚀 Feature Request

Publish move-analyzer binaries for different platforms, including MacOS, Windows, and Linux.

## Motivation

After installing the move-analyzer plug-in, it prompts that the move-analyzer binary cannot be found. I hope the plug-in can be downloaded automatically or provide a download page.

tnowacki:
@awelc, putting this on your radar.

yubing744:
@awelc Can you help create a Release that generates the move-analyzer binary?

Example:  https://github.com/yubing744/starcoin-move/releases/tag/v0.1.2

PR #208, opened 2022-06-18
[move-analyzer] Automatically upload the move binary program after the Release is generated

yubing744:
## Motivation

Fix https://github.com/move-language/move/issues/204

## Test Plan

I tested it successfully in my own repository 

Release：https://github.com/yubing744/starcoin-move/releases/tag/v0.1.2
Pipeline:  https://github.com/yubing744/starcoin-move/actions/runs/2516531587


yubing744:
@awelc Can help review this PR.

.github/workflows/release_asset.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: install rust toolchain
+        uses: actions-rs/toolchain@v1
+      - name: build for ${{ matrix.platform }}
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --release
+
+      - name: build move release asset
+        run: bash ./scripts/release.sh ${{ matrix.platform }}
```
bmwill, line 32: You shouldn't need to specify `bash` here
```suggestion
        run: ./scripts/release.sh ${{ matrix.platform }}
```

.github/workflows/release_asset.yml
```diff
@@ -0,0 +1,51 @@
...
+        if: ${{ github.event_name == 'release'}}
+        id: upload-release-asset
+        uses: actions/upload-release-asset@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          upload_url: ${{ github.event.release.upload_url }}
+          asset_path: ./move-${{ matrix.platform }}.zip
+          asset_name: move-${{ matrix.platform }}.zip
+          asset_content_type: application/zip
```
bmwill, line 51: Would you elaborate, maybe even add a comment to the file, about the purpose of each of these different jobs?

My best guess is that one is for when a "release" even occurs (by creating a new tag) and the other is for manually triggering the job. Is that correct?

scripts/release.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+# Copyright (c) The Diem Core Contributors
```
bmwill, line 2: idk what the consensus was on the copyright headers but i don't think you need the Diem ones anymore?

scripts/release.sh
```diff
@@ -0,0 +1,20 @@
...
+  cp -v target/release/$BIN move-artifacts/
+done
+
+cp -v README.md move-artifacts/
+
+if [ "$1" == "windows-latest" ]; then
+  7z a -r move-$1.zip move-artifacts
+else
+  zip -r move-$1.zip move-artifacts
+fi
```
bmwill, line 20: This script doesn't actually do any of the building of the assets which feels a little bit odd. I'd probably add a `cargo build --release --bin <bin> --bin <bin>` to this script and then remove the build from the CI job. That way you can actually run this script without assuming that someone already did the `cargo build` step previously.

.github/workflows/release_asset.yml
```diff
@@ -0,0 +1,51 @@
...
+    name: Build release asset
+    continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - ubuntu-latest
+          - ubuntu-18.04
+          - macos-latest
+          - windows-latest
```
bmwill, line 18: We do want move to work properly on windows as well as these other platforms, do we currently have a CI job that actually builds and runs tests on all of these platforms?

.github/workflows/release_asset.yml
yubing744: Got it, I'll remove it

.github/workflows/release_asset.yml
yubing744: ok, i'll write a code comment

scripts/release.sh
yubing744: The CI check prompt requires a copyright notice, I copied a copy from git-checks.sh

https://github.com/move-language/move/blob/548f632ca3ace2707c59c96dda198a9a7bead592/scripts/git-checks.sh#L2

scripts/release.sh
yubing744: I'll change the name of the script and add some code comments

.github/workflows/release_asset.yml
yubing744: I'm not sure, but the move-analyzer VSCode plugin requires a Windows version binary.

yubing744:
＠bmwill　I updated the PR code

.github/workflows/release_asset.yml
vgao1996: > do we currently have a CI job that actually builds and runs tests on all of these platforms?

AFAIK ubuntu is the ONLY platform on which we run our CI tests. I guess it would be tricky to have a mac os or windows based testing platform even if this is something we should ideally have.

yubing744: I can add some CI tests, on mac and windows platforms

bmwill (approved):
Changes generally look good to me, still some uncertainty around the CI jobs for non-linux platforms. I'll approve but leave it to @awelc to give the final say

awelc:
@bmwill, thank you for taking this one on!

.github/workflows/release_asset.yml
yubing744: On the windows platform, there is a problem with removing bash, I added it back through this PR

PR #211, opened 2022-06-19
[move-to-yul] Extend move-to-yul compiler to support async move contracts

rahxephon89:
## Motivation

This PR proposes a possible way to compile async move contracts into Yul code:

- #[actor] corresponds to #[evm_contract]
- #[init] corresponds to #[create]
- #[message] corresponds to #[callable]
- #[state] corresponds to #[storage]
- calling of send_<foo>(actor: address, args) translates into emit of an `ethereum` event Foo(actor, message_hash, args) where message_hash currently follows the code [here](https://github.com/move-language/move/blob/341adf2224cb3d9331648285b61c0d299931d717/language/move-compiler/src/attr_derivation/async_deriver.rs#L289)

Yes

## Test Plan

hardhat test: AsyncEvent
baseline test: AccountStateMachine.move


language/evm/move-to-yul/src/events.rs
```diff
@@ -186,6 +186,163 @@ impl EventSignature {
     }
 }
 
+/// Generate emit functions for send_<name>
+pub(crate) fn define_emit_fun_for_send(
```
runtian-zhou, line 190: Would be nice to gate those functionalities under a `async_move` compilation flag.

language/evm/move-to-yul/src/events.rs
```diff
@@ -186,6 +186,163 @@ impl EventSignature {
...
+        let topic_0_hash = format!("0x{:x}", Keccak256::digest(event_sig_str.as_bytes()));
+
+        emitln!(ctx.writer, "let {} := {}", topic_0_var, topic_0_hash);
+        let message_str = format!(
+            "{}::{}",
+            fun.module_env.self_address(),
+            fun.get_full_name_str().replace("send_", "")
+        );
+
+        let hash_bytes = Sha3_256::digest(message_str.as_bytes());
```
runtian-zhou, line 228: Do you mind explain a bit how this part of code works? Specifically how will the event be encodeded? Maybe some comments here would be great.

language/move-compiler/src/attr_derivation/async_deriver.rs
```diff
@@ -287,6 +312,7 @@ fn message_hash(
     };
     let addr_str = format!("0x{:X}", account_addr);
     let hash_str = format!("{}::{}::{}", addr_str, mod_def.name, fun_def.name);
+    println!("hash_str:{}", hash_str);
```
runtian-zhou, line 315: Remove the print?

language/move-compiler/src/attr_derivation/async_deriver.rs
```diff
@@ -117,12 +131,23 @@ fn derive_module_for_async(
                     sign,
                     send_call,
                 ));
+
+                let mut struct_name = fun_def.name.to_string();
+                struct_name.replace_range(0..1, &struct_name[0..1].to_uppercase());
+                new_structs.push(new_struct(
```
runtian-zhou, line 137: IIRC for every send message we will create a struct here. Am I understanding it correctly?

rahxephon89: Yes, the struct is annotated with #[event] so that the move-to-yul compiler can generate code to emit the corresponding event.

language/evm/move-to-yul/src/events.rs
rahxephon89: Added a gate to generate this function only when the contract is annotated with #[actor]

language/evm/move-to-yul/src/events.rs
```diff
@@ -186,6 +186,168 @@ impl EventSignature {
...
+        let signature_types = &event_sig.para_types;
+        let topic_0_var = ctx.make_local_name(target, local_name_idx);
+        local_name_idx += 1;
+        let event_sig_str = event_sig.to_string();
+        let topic_0_hash = format!("0x{:x}", Keccak256::digest(event_sig_str.as_bytes()));
+
+        emitln!(ctx.writer, "let {} := {}", topic_0_var, topic_0_hash);
+
+        // compute the message hash
+        // message_hash is the first 8 bytes of Sha3_256::digest(address::module_name::foo)
```
runtian-zhou, line 226: I think you should use `message_hash` function in `language/extensions/async/move-async-vm/src/actor_metadata.rs`

language/evm/hardhat-examples/test/AsyncEvent.test.js
```diff
@@ -0,0 +1,22 @@
...
+const make_test = function (contract_name) {
+  return function () {
+    before(async function () {
+      this.Event = await ethers.getContractFactory(contract_name);
+      this.event = await this.Event.deploy();
+      await this.event.deployed();
+    });
+    it("AsyncEvent_xfer_deposit", async function () {
+      const tx = this.event.xfer(this.event.address,42);
+      await expect(tx).to.emit(this.event, 'Xfer_deposit').withArgs(this.event.address, 0x0b32e0fa, 42, this.event.address, 0);
```
runtian-zhou, line 13: This would be a great starter test!

language/evm/move-to-yul/src/events.rs
```diff
@@ -186,6 +186,168 @@ impl EventSignature {
     }
 }
 
+/// Generate event emit functions for send_<name> in async move contracts
+/// example: the function send_foo(actor: address, args) will emit an ethereum event
+/// Foo(actor, message_hash, args) where encoding follows the ethereum standard
+pub(crate) fn define_emit_fun_for_send(
```
runtian-zhou, line 192: This logics makes sense to me for now. I was wondering if we should add a sequence number to the event to help perform deduplication on the listener side. Maybe we can revisit the definition once we start to implement the listener code.

language/evm/move-to-yul/src/events.rs
```diff
@@ -186,6 +186,168 @@ impl EventSignature {
...
+        for (i, (_, solidity_ty, move_ty, indexed_flag, _)) in signature_types.iter().enumerate() {
+            let mut var = if i == 0 {
+                params_vec.get(0).unwrap().to_string()
+            } else if i == 1 {
+                message_hash_var.clone()
+            } else {
+                params_vec.get(i - 1).unwrap().to_string()
+            };
+
+            if *indexed_flag {
```
runtian-zhou, line 258: Can you give me an example when this could be indexed? IIUC we don't need to index any of the extra arguments for param.

language/evm/move-to-yul/src/events.rs
```diff
@@ -186,6 +186,168 @@ impl EventSignature {
...
+            message_hash_str
+        );
+
+        let mut indexed_paras = vec![];
+        let mut indexed_vars = vec![topic_0_var];
+        let mut unindexed_paras = vec![];
+        let mut unindexed_vars = vec![];
+
+        for (i, (_, solidity_ty, move_ty, indexed_flag, _)) in signature_types.iter().enumerate() {
+            let mut var = if i == 0 {
```
runtian-zhou, line 250: Do we index message hash and actor address? I don't see a clear reason why we should/should not index them but would be great if we could clarify the default behavior there and adda TODO to revisit the decision later.

language/evm/move-to-yul/src/functions.rs
```diff
@@ -43,11 +43,12 @@ impl<'a> FunctionGenerator<'a> {
         let fun = &ctx.env.get_function(fun_id.to_qualified_id());
         // TODO: change back to is_native_or_intrinsic if we decide to implement
         // intrinsic functions (such as reverse, contains, is_empty) in the Vector module as well
-        if fun.is_native() {
+        if fun.is_native() || fun.get_full_name_str().contains("send_") {
```
runtian-zhou, line 46: It would be nice to create a compilation flag to gate this feature IMO so that this if would only be active when a rust feature flag is set, similar to how we gate the table extension

runtian-zhou (commented):
Overall looks good to me! Do you mind add those TODOs for the decision we might revisit later and we can land this PR IMO.

language/evm/move-to-yul/src/events.rs
rahxephon89: I am not sure what you mean here but we can add indexed to parameters if there is a need.

language/evm/move-to-yul/src/functions.rs
rahxephon89: I added a gate when _function_ is called so that only when #[actor] is specified, the feature is activated.

language/evm/move-to-yul/src/events.rs
runtian-zhou: IIUC the arguments here would be receipient address, function hash, and the rest of the arguments. Are we going to index any of those parameters? Can module author specific which arguments to index?

runtian-zhou (approved):
LGTM!

PR #220, opened 2022-06-23
[Move] Added Move analyzer's support for multi-root workspaces

awelc:
## Motivation

This is another language server enhancement, this time related to supporting multi-root workspaces. This PR not only adds support for opening multiple directories within a workspace but also adds support for manifest file searching in case a single Move source file is open. 

The implementation is fairly straightforward - instead of replacing existing symbolication information, augment it with additional symbols coming from additional files/directories being opened.

This PR also implements discovery of manifest files in case a Move source file is opened rather than a package directory. Previously, if the former has happened, the new features would not be available as the language server did not know where to find Move.toml file needed for compilation.

Yes


emmazzz (approved):
LGTM

PR #224, opened 2022-06-27
[move-analyzer] Fix bug for move-analyzer not found

yubing744:
## Motivation

The previous default value configuration, even if I configure the PATH, it will prompt that the executable program cannot be found [here](https://github.com/move-language/move/blob/548f632ca3ace2707c59c96dda198a9a7bead592/language/move-analyzer/editors/code/src/context.ts#L22).

YES

## Test Plan

I manually tested three scenarios:
1. If move-analyzer is not installed, an error will be prompted
2. Through cargo install, the plugin can be loaded normally
3. Configure the move-analyzer path through the plug-in, and the plug-in can be loaded normally


awelc (approved):
LGTM!

PR #229, opened 2022-06-28
[Move analyzer] Added support for more verbose compiler diagnostics

awelc:
## Motivation

This PR augments compiler diagnostics passed to the IDE with additional information. Previously we were only passing the main error message as it seemed like a good idea to keep the information displayed on-hover brief. However, in the case of some error messages, the main message is not quite enough to fully understand what's going on, hence adding additional info.

Yes

### DIFF
--- a/language/move-analyzer/src/diagnostics.rs
+++ b/language/move-analyzer/src/diagnostics.rs
@@ -3,11 +3,12 @@
 
 use crate::utils::get_loc;
 use codespan_reporting::{diagnostic::Severity, files::SimpleFiles};
-use lsp_types::{Diagnostic, DiagnosticSeverity, Range};
+use lsp_types::{Diagnostic, DiagnosticRelatedInformation, DiagnosticSeverity, Location, Range};
 use move_command_line_common::files::FileHash;
 use move_ir_types::location::Loc;
 use move_symbol_pool::Symbol;
 use std::collections::{BTreeMap, HashMap};
+use url::Url;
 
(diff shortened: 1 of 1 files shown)


language/move-analyzer/src/diagnostics.rs
```diff
@@ -23,11 +24,36 @@ pub fn lsp_diagnostics(
...
+                    None
+                } else {
+                    let mut related_info = vec![];
+                    for (lloc, lmsg) in labels {
+                        if let Some(lstart) =
+                            get_loc(&lloc.file_hash(), lloc.start(), files, file_id_mapping)
+                        {
+                            if let Some(lend) =
+                                get_loc(&lloc.file_hash(), lloc.end(), files, file_id_mapping)
+                            {
```
tnowacki, line 42: A bit tricky to make "cleaner", but I think using `?` here would help a lot 
```
Some(labels.into_iter()
    .filter_map((lloc, lmsg) {
        let lstart = get_loc(&lloc.file_hash(), lloc.start(), files, file_id_mapping)?;
        let lend = get_loc(&lloc.file_hash(), lloc.end(), files, file_id_mapping)?;
        ...
        Some(DiagnosticRelatedInformation { ... })
     })
     .collect())
```

tnowacki (approved):

language/move-analyzer/src/diagnostics.rs
awelc: Nice trick with `?` and `filtered_map`!

PR #232, opened 2022-06-29
Aw/missing manifest fix

awelc:
## Motivation

This PR modifies the time when a manifest file is being located by the language server. Previously finding manifest file was attempted at upon two different events:
- when the first root directory was opened in the IDE
- when a Move source file was opened in the IDE

The first one is not only redundant (as opening a directory without opening a Move source file is basically a no-op) but also may lead to a confusion - for example if a user opens a root directory of a repo containing multiple Move packages/projects, they will get an error as the manifest file in this case will not be found in a parent directory of the repository root (and even if some manifest file is found, it will be most likely an incorrect one). This very problem was reported by @emmazzz 

Yes

### DIFF
--- a/language/move-analyzer/src/symbols.rs
+++ b/language/move-analyzer/src/symbols.rs
@@ -431,6 +431,7 @@ impl SymbolicatorRunner {
         cvar.notify_one();
     }
 
+    /// Finds manifest file in a subdirectory of a Move source file passed as argument
     fn root_dir(starting_path: &Path) -> Option<PathBuf> {
         eprintln!("Looking for manifest file");
         let mut current_path_opt = Some(starting_path);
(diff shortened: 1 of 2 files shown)


language/move-analyzer/src/symbols.rs
```diff
@@ -431,16 +431,21 @@ impl SymbolicatorRunner {
         cvar.notify_one();
     }
 
+    /// Finds manifest file in a subdirectory of a Move source file passed as argument
     fn root_dir(starting_path: &Path) -> Option<PathBuf> {
+        eprintln!("Looking for manifest file");
```
emmazzz, line 436: Maybe remove these debug prints?

emmazzz (approved):

language/move-analyzer/src/symbols.rs
awelc: Good point! I thought I'd leave the for future debugging in the field but since these are printed every time a file is changed, it's a bit much.

PR #237, opened 2022-07-04
[stdlib] UTF8 String type

wrwg:
This introduces a new type `string::String` which represents a valid UTF8 string. The type is intended to replace `ascii::string`.

The UTF8 enforcement is based on Rust's and correct only modulo Rust. When creating a string with `string::utf8(bytes)`
the Rust `str` implementation will check the correctness of the encoding. On deconstructing or composing strings, the Rust function `str::is_valid_char_boundary` is used to allow composition only on utf8 boundaries. The String type does not allow direct mutation of bytes, therefore utf8 correctness is not violated in other ways.

Currently, the only way to denote a string is as in `string::utf8(b"hello")`. If someone would want to create unicode characters, they would have to use the `\x1F` escape in the Move string syntax. We may want to consider introducing string literal `u"foo"` which can directly be used to denote strings, and which also supports `\u{digits}` style of unicode escapes.


jolestar:
Is it possible to use `string::String` as the transaction argument?

language/move-stdlib/sources/string.move
```diff
@@ -0,0 +1,99 @@
...
+        let end = sub_string(s, at, l);
+        append(&mut front, o);
+        append(&mut front, end);
+        *s = front;
+    }
+
+    /// Returns a sub-string using the given byte indices, where is the first position and j is the start of the first
+    /// byte not included (or the length of the string). The indices must be at valid utf8 char boundaries,
+    /// guaranteeing that the result is valid utf8.
+    public fun sub_string(s: &String, i: u64, j: u64): String {
```
areshand, line 78: should the i and j be the i-th and j-th character instead of i and j th bytes?

wrwg:
> Is it possible to use `string::String` as the transaction argument?

This would be desirable, however, the transaction dispatcher then needs to check the utf8 correctness.

language/move-stdlib/sources/string.move
```diff
@@ -0,0 +1,99 @@
...
+        assert!(at <= vector::length(bytes) && internal_is_char_boundary(bytes, at), EINVALID_INDEX);
+        let l = length(s);
+        let front = sub_string(s, 0, at);
+        let end = sub_string(s, at, l);
+        append(&mut front, o);
+        append(&mut front, end);
+        *s = front;
+    }
+
+    /// Returns a sub-string using the given byte indices, where is the first position and j is the start of the first
```
jiangying000, line 75:
```suggestion
    /// Returns a sub-string using the given byte indices, where i is the first position and j is the start of the first
```

language/move-stdlib/src/natives/string.rs
```diff
@@ -0,0 +1,98 @@
...
+    mut args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(args.len() == 3);
+    let j = pop_arg!(args, u64) as usize;
+    let i = pop_arg!(args, u64) as usize;
+    let s_arg = pop_arg!(args, VectorRef);
+    let s_ref = s_arg.as_bytes_ref();
+    unsafe {
+        let s_str = std::str::from_utf8_unchecked(s_ref.as_slice());
+        let v = Value::vector_u8((&s_str[i..j]).as_bytes().iter().cloned());
```
davidiw, line 70: this is interesting, we've defined s_str as a string but we can now index into it?

language/move-stdlib/src/natives/string.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+}
+
+pub fn native_index_of(
+    context: &mut NativeContext,
+    _ty_args: Vec<Type>,
+    mut args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(args.len() == 2);
+    unsafe {
```
davidiw, line 83: could we add a note about what is unsafe and why we're doing it?

language/move-stdlib/src/natives/string.rs
```diff
@@ -0,0 +1,98 @@
...
+    context: &mut NativeContext,
+    _ty_args: Vec<Type>,
+    mut args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(args.len() == 3);
+    let j = pop_arg!(args, u64) as usize;
+    let i = pop_arg!(args, u64) as usize;
+    let s_arg = pop_arg!(args, VectorRef);
+    let s_ref = s_arg.as_bytes_ref();
+    unsafe {
```
davidiw, line 68: could we add a note about what is unsafe and why we're doing it?

language/move-stdlib/src/natives/string.rs
```diff
@@ -0,0 +1,98 @@
...
+pub fn native_is_char_boundary(
+    context: &mut NativeContext,
+    _ty_args: Vec<Type>,
+    mut args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(args.len() == 2);
+    let i = pop_arg!(args, u64);
+    let s_arg = pop_arg!(args, VectorRef);
+    let s_ref = s_arg.as_bytes_ref();
+    unsafe {
```
davidiw, line 50: could we add a note about what is unsafe and why we're doing it?

davidiw (commented):
Seems fine for me, will let it sit for a few another day or so for more feedback.

language/move-stdlib/sources/string.move
```diff
@@ -0,0 +1,99 @@
...
+
+    /// An invalid UTF8 encoding.
+    const EINVALID_UTF8: u64 = 1;
+
+    /// Index out of range.
+    const EINVALID_INDEX: u64 = 2;
+
+    /// A `String` holds an immutable sequence of bytes which is guaranteed to be in utf8 format.
+    struct String has copy, drop, store {
+        bytes: vector<u8>,
```
tnowacki, line 14: Totally fine with this for now, but is this BCS compatible with BCS utf8 strings?

One benefit of doing a native string is that BCS will verify that the string is utf8, which is really nice to have for transaction/`entry` function arguments.

language/move-stdlib/src/natives/string.rs
wrwg: Rust slice `s[i..j]` is available on Vec, &[], String, and &str.

language/move-stdlib/src/natives/string.rs
wrwg: Done. Also narrow the unsafe down to the actual unsafe function.

language/move-stdlib/sources/string.move
```diff
@@ -0,0 +1,99 @@
...
+    use std::vector;
+    use std::option::{Self, Option};
+
+    /// An invalid UTF8 encoding.
+    const EINVALID_UTF8: u64 = 1;
+
+    /// Index out of range.
+    const EINVALID_INDEX: u64 = 2;
+
+    /// A `String` holds an immutable sequence of bytes which is guaranteed to be in utf8 format.
```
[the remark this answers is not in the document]
wrwg: Good catch. Yes, I started with the idea of those being immutable, but then started to understand the Rust string model, which guarantees preservation of utf8 by only allowing slicing, insertion, and removal on char boundaries. Interesting enough, the approach of using byte indices everwhere but then have this concept of char boundaries is pretty practical. The check for char boundary can be done locally to the byte index, with some more forward and backward context, I guess.

language/move-stdlib/sources/string.move
```diff
@@ -0,0 +1,99 @@
...
+    public fun try_utf8(bytes: vector<u8>): Option<String> {
+        if (internal_check_utf8(&bytes)) {
+            option::some(String{bytes})
+        } else {
+            option::none()
+        }
+    }
+
+    /// Creates a new string from a sequence of bytes. Performs no checks.
+    public fun utf8_unchecked(bytes: vector<u8>): String {
```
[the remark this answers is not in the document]
wrwg: Removed.

language/move-stdlib/sources/string.move
wrwg: Totally understand the benefit of a native string type. But is is also very costly to implement :-/ You mentioned before when we discussed impl of u256 that it is difficult to add a new type to the system, and sure it is. 

The deserialization problem is the only one I see for strings which a native type could do better. Memory representation should be good (and actually being able to "move " `vector<u8>` into a string is awesome for perf).

But we need to solve the problem of invariants on transaction parameters anyway, and String is only one special case of those.  I think we need some kind of allow list on structs for transactions, and then custom validators for them.

tnowacki: > But we need to solve the problem of invariants on transaction parameters anyway, and String is only one special case of those. I think we need some kind of allow list on structs for transactions, and then custom validators for them.

It is indeed something we need to solve. And we should. 

My hope was that if the BCS serialization format would be the same for this struct  and the BCS strings, that we could drop in replace this string struct with a native type over time.

language/move-core/types/src/unit_tests/value_test.rs
```diff
@@ -48,6 +48,24 @@ fn struct_deserialization() {
...
+    let val = MoveValue::Vector(vec![
+        MoveValue::U8(1),
+        MoveValue::U8(22),
+        MoveValue::U8(13),
+        MoveValue::U8(99),
+    ]);
+    let s1 = MoveValue::Struct(MoveStruct::Runtime(vec![val.clone()]))
+        .simple_serialize()
+        .unwrap();
+    let s2 = val.simple_serialize().unwrap();
```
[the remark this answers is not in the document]
tnowacki: I just wanted to make sure we could drop in replace a native string type without breaking old serialization formats, which I think we can do. It's less about structs having no overhead (which is actually tested all over the place, I abuse the hell out of it in the transactional tests lol), and more about making sure the format for BCS strings is what we expect it to be. As you are saying the tests are equivalent (especially if you have one that links vector<u8> with utf8 strings)

wrwg:
Ping on this PR @tnowacki , @sblackshear @davidiw . Either let me know what else to consider here or let us land this.

language/move-vm/types/src/lib.rs
```diff
@@ -4,6 +4,7 @@
 
 #![forbid(unsafe_code)]
 
+extern crate core;
```
tnowacki, line 7: What is this added for?

language/move-stdlib/src/natives/string.rs
```diff
@@ -0,0 +1,98 @@
...
+    context: &mut NativeContext,
+    _ty_args: Vec<Type>,
+    mut args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(args.len() == 1);
+    let s_arg = pop_arg!(args, VectorRef);
+    let s_ref = s_arg.as_bytes_ref();
+    let ok = std::str::from_utf8(s_ref.as_slice()).is_ok();
+    // TODO: extensible native cost tables
+    let cost = native_gas(context.cost_table(), NativeCostIndex::PUSH_BACK, 1);
```
tnowacki, line 37: Can we just add new indices for the string ops here?

language/move-core/types/src/unit_tests/value_test.rs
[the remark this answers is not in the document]
tnowacki: Ah yeah thanks for the added string test!

tnowacki (approved):
Thanks for taking this on! Can't wait to play with it on our end and deprecate our own UTF8 wrapper :) 

Just a question about the gas for this, but not blocking

tnowacki:
A note on syntax:
> Currently, the only way to denote a string is as in string::utf8(b"hello"). If someone would want to create unicode characters, they would have to use the \x1F escape in the Move string syntax. We may want to consider introducing string literal u"foo" which can directly be used to denote strings, and which also supports \u{digits} style of unicode escapes.

I would suggest, no `u` prefix, and add `\u` escapes to bytestrings too. For example
`"hello \u{1f30d}" : std::string::String`
`b"hello \u{1f30d}": vector<u8>`

language/move-stdlib/src/natives/string.rs
wrwg: @vgao1996 is currently working on extending and consolidating the gas model. The problem here is that we can extend the indices for natives within the Move repo, but not outside.

language/move-vm/types/src/lib.rs
wrwg: Removed

PR #240, opened 2022-07-05
Flatten package commands

tnowacki:
- Made all package commands base-line
- Should make usage of the move cli a bit more natural

## Motivation

- Splitting into separate command structure like this will make the reusing of the commands a lot easier
- And with other CLI tools re-using the commands, they can easy add extra/custom args with `#[clap(flatten)]` and a separate struct

## Test Plan

- updated tests


awelc (approved):
LGTM

PR #242, opened 2022-07-05
[move-cli] Fix 'new' command

tnowacki:
- Fix tabbing and new lines on command
- Fix behavior when no path is specified

## Test Plan

- New test


awelc (approved):
LGTM

PR #243, opened 2022-07-06
Write Test Output To Any Writer Over stdout

oxade:
## Motivation
When embedding tests, we want to capture the output. This PR allows specifying output modes besides stdout

Yes
## Test Plan
Standard tests


language/tools/move-cli/src/base/test.rs
```diff
@@ -136,12 +137,13 @@ pub enum UnitTestResult {
     Failure,
 }
 
-pub fn run_move_unit_tests(
+pub fn run_move_unit_tests<W: Write + Send>(
```
tnowacki, line 140: Do you need `Send` here?

tnowacki (approved):

PR #245, opened 2022-07-06
[unit-test] Allow extensions used for unit testing to live outside of Move repo

wrwg:
Until now, extensions leveraged by the Move UT infra had to live in the Move repo.

This PR adds a hook which can be used as in `move_unit_tests::extensions::set_extension_hook(fun)`,
where `fun` is populating a `NativeContextExtensions` for the according custom environment. How
to use this is documented in the code at `extensions.rs` and also in an integrated unit test.

### DIFF
--- a/language/tools/move-unit-test/src/extensions.rs
+++ b/language/tools/move-unit-test/src/extensions.rs
@@ -7,7 +7,8 @@
 //! to be usable.
 
 use move_vm_runtime::native_extensions::NativeContextExtensions;
-use std::fmt::Write;
+use once_cell::sync::Lazy;
+use std::{fmt::Write, sync::Mutex};
 
 #[cfg(feature = "table-extension")]
 use itertools::Itertools;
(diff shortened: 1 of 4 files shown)


language/tools/move-unit-test/src/extensions.rs
```diff
@@ -73,3 +94,26 @@ fn print_table_extension<W: Write>(w: &mut W, extensions: &mut NativeContextExte
 
 #[cfg(feature = "table-extension")]
 static DUMMY_RESOLVER: Lazy<BlankStorage> = Lazy::new(|| BlankStorage);
+
+
+#[cfg(test)]
+mod tests {
+    use better_any::{Tid, TidAble};
```
davidiw, line 101: I don't see this being used?

davidiw (approved):

language/tools/move-unit-test/src/extensions.rs
wrwg: Not sure what you mean. The test is executed. On needs to have both imports for `#[derive(Tid)]`.

davidiw: indeed... sometimes I wish I could delete comments, but at least I can resolve this one :)

PR #258, opened 2022-07-11
WIP: cfg flag to skip bytecode verif

oxade:
## Motivation
Note: I'm open to a better way to achieve this. This was just the easiest way I found.
Need to bypass verifier in some instances for instruction calibration.
Currently hand-writing bytecode to allow for calibrating VM costs, however verifier is preventing certain states.

Yes

## Test Plan

Checking that verifier does not cause restrictions


sblackshear:
Can you share the non-verifying code that you are trying to run, and perhaps we can suggest alternative ways of writing the same code that will verify? The VM assumes the bytecode verifier has run, so the meaning (either in terms of results or execution time) of running non-verifiable code is not defined. I would not want to conclude anything about performance by running such code.

PR #218, opened 2022-06-23
[move-analyzer] Support document symbol PSL

yubing744:
## Motivation

[starcoin-ide](https://github.com/starcoinorg/starcoin-ide) expects that users can select a single test function on the interface, and then execute the test. In order to realize this function, I need to extract the test function entry in the move file. I found that move-analyzer has parsed the AST of Move, and found the textDocument/DocumentSymbol interface of the LSP protocol, which can get all the function information in the move file, but move-analyzer has not implemented this interface yet, so I submitted this PR.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/move-language/move/blob/main/CONTRIBUTING.md#pull-requests)?

YES

## Test Plan

I added a use case to the move-analyzer plugin to test the textDocument/DocumentSymbol protocol.
Test file:  https://github.com/yubing744/starcoin-move/blob/support-document-symbol-psl/language/move-analyzer/editors/code/tests/lsp.test.ts

PR #259, opened 2022-07-11
[Move analyzer] Added support for computing first symbol info synchronously

awelc:
## Motivation

The goal of this PR is to enable unit tests that require symbolication information to be available to the IDE client right away, that is right after a connection with the language server is successfully established. Otherwise, we either need to use hard-coded delays (which is arguably not the prettiest nor most reliable approach) or create a more elaborate synchronization mechanism (which is otherwise unnecessary and may affect server responsiveness). 

Yes


awelc:
This has been confirmed to work for the intended use case: https://github.com/move-language/move/pull/218

emmazzz (approved):
LGTM

PR #260, opened 2022-07-12
[Move analyzer] Changed the max address length limit

awelc:
## Motivation

Max length of the address in core Move is 16 bytes but this is not true for other Move applications where this length can be as long as 32 bytes. This PR modifies Move analyzer (and Move analyzer only) to use longer max length to avoid compilation errors for Move uses outside of Move core. 

The max length is by default set to 32 bytes, but build-time options (via Cargo features) have been provided to either set it to core Move default (by resetting the default features of the Move analyzer) or by choosing 20 (or 32) bytes via selecting a specific Cargo feature.

EDIT:

It seems like defaulting to 32. bytes is not going to work without a more major restructuring of core Move code (as the feature the propagates to tests etc.), and this does not seem worthwhile at this point. A simpler option is to update the installation instructions after this PR lands.

Yes

### DIFF
--- a/language/move-analyzer/Cargo.toml
+++ b/language/move-analyzer/Cargo.toml
@@ -26,3 +26,8 @@ move-compiler = { path = "../move-compiler" }
 move-ir-types = { path = "../move-ir/types" }
 move-package = { path = "../tools/move-package" }
 move-symbol-pool = { path = "../move-symbol-pool" }
+
+[features]
+default = ["address32"]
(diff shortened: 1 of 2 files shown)


tnowacki (approved):

tnowacki:
I think this will be a bigger and bigger tooling headache as time goes on, but happy to keep kicking this can down the road for now.

PR #265, opened 2022-07-13
[vm] Allow native functions to reflect over the stack frame context.

wrwg:
This gives access to `ExecutionState` (stack frame of calls) to native functions. This is needed e.g. to implement
programmatic access to the address of the calling function.


vgao1996 (approved):
LGTM

Issue #236, opened 2022-07-04
[Feature Request] Add better `--only-deps` to refresh dependencies inside the `.move` directory


mkurnikov:
# 🚀 Feature Request

I would like to have a flag to the build pipeline, not to touch the current project. Instead, it should refresh the `.move` directory with new dependencies, to allow to re-index them in IDE later. 

## Motivation

IDE needs to have an internal project model for resolving modules and named addresses. Currently, intellij-move builds the address graph (no need for `BuildInfo.yaml`) and constructs the correct `.move` paths for all the dependencies. It does not, however, have git support, so I cannot refresh the dependencies, and would like `move` binary to do it for me. 

Related https://github.com/move-language/move/pull/104

PR #267, opened 2022-07-14
Rename `--only-deps` to `--fetch-deps-only`, make it only populate MOVE_HOME

mkurnikov:
## Motivation

Fixes https://github.com/move-language/move/issues/236

Overwrites https://github.com/move-language/move/pull/104

AFAIK, the only IDE that uses dependencies source code is the Intellij-move plugin. Move-analyzer yet to support that, I assume no breakages there. 

@vladimirovmm @awelc @tnowacki 

Yes

## Test Plan

Testing would require mocking `MOVE_HOME`, calling the command with the flag, and checking that correct git repos are cloned. I want to hear the feedback first, then I'll try to write one.


language/tools/move-package/src/resolution/resolution_graph.rs
```diff
@@ -515,6 +515,33 @@ impl ResolvingGraph {
...
+        // include dev dependencies if in dev mode
+        let additional_deps = if build_options.dev_mode {
+            package.dev_dependencies.clone()
+        } else {
+            BTreeMap::new()
+        };
+
+        for (dep_name, dep) in package
+            .dependencies
+            .into_iter()
```
tnowacki, line 532: should this just be an `iter()`? Then everything else here can be by reference?

tnowacki (commented):
No issue with the change to the flag. Just some comments on the implementation and then should be good to go :)

language/tools/move-package/src/resolution/resolution_graph.rs
mkurnikov: I've updated the code, hopefully it's what you had in mind.

mkurnikov:
On the test side of things, it's a network-heavy feature that requires mocking many things. Considering its niche use case, I'd leave it untested.

language/tools/move-package/src/resolution/resolution_graph.rs
mkurnikov: Yes, but in 
```
        if config.fetch_deps_only {
            config.download_deps_for_package(&rerooted_path)?;
            return Ok(());
        }
```
https://github.com/move-language/move/pull/267/files#diff-5d804e276a879688a14ffe2fc7342accb0050d6e84418a8cd855a9faa93d1ce4R18
`config` is destroyed after the `download_deps_for_package` anyway, so it doesn't need to be a reference. 

Same for the `SourceManifest`, it's not used outside the `download_dependency_repos` call, so there's no need for a reference. Unless I'm missing something. 

I've fixed the `root_path`, it does remove one `clone()`.

tnowacki: > config is destroyed after the download_deps_for_package anyway

Sure, but you aren't actually using it by-value in any meaningful or required way. So if anyone needs to modify surrounding code afterwards, they are now stuck.

It is best to use things by-reference unless you have some reason not to (prevents clones, ensures proper mutations, etc)

mkurnikov: Could you review it again? I made everything into a by-reference, and extracted the only mutation required for the `mut self` to the higher scope.

language/tools/move-package/src/resolution/resolution_graph.rs
```diff
@@ -515,6 +515,30 @@ impl ResolvingGraph {
...
     }
 
+    pub fn download_dependency_repos(
+        manifest: &SourceManifest,
+        build_options: &BuildConfig,
+        root_path: &Path,
+    ) -> Result<()> {
+        // include dev dependencies if in dev mode
+        let additional_deps = if build_options.dev_mode {
+            manifest.dev_dependencies.clone()
```
tnowacki, line 525: This clone isn't needed if you are just using `.iter()` later.

I imagine you did this `clone` due to Rust borrow checker issues, and it is indeed an annoying one. 

You might need to do this:
```
let empty_deps;
let additional_deps = if build_options.dev_mode {
        manifest.dev_dependencies.clone()
} else {
    empty_deps = BTreeMap::new();
    &empty_deps
};
```

language/tools/move-package/src/resolution/resolution_graph.rs
```diff
@@ -515,6 +515,30 @@ impl ResolvingGraph {
...
+        } else {
+            BTreeMap::new()
+        };
```
tnowacki, line 528:
```suggestion
        let empty_deps;
        let additional_deps = if build_options.dev_mode {
            manifest.dev_dependencies.clone()
        } else {
            empty_deps = BTreeMap::new();
            &empty_deps
        };
```
If you just want to apply it here :)

tnowacki (commented):
Everything looks good otherwise! Sorry for the clone pedantry, I would normally just land it and make any changes as needed later, but I don't have a ton of time for that sort of thing these days unfortunately.

tnowacki (approved):
Sorry for being such a curmudgeon, thanks for the PR!

PR #272, opened 2022-07-17
[table] add information about the table types when adding new tables

davidiw:
Getting this information requires deep understanding of changesets to otherwise get the typeinfo for the key and value. And doing so expensive. This is relatively cheap and can be used by the node to support lookup and decoding of types in storage.

### DIFF
--- a/language/extensions/move-table-extension/src/lib.rs
+++ b/language/extensions/move-table-extension/src/lib.rs
@@ -12,6 +12,7 @@ use move_binary_format::errors::{PartialVMError, PartialVMResult};
 use move_core_types::{
     account_address::AccountAddress,
     gas_schedule::{GasAlgebra, GasCarrier, InternalGasUnits},
+    language_storage::TypeTag,
     value::MoveTypeLayout,
     vm_status::StatusCode,
 };
@@ -49,10 +50,31 @@ impl Display for TableHandle {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct TableInfo {
+    pub key_type: TypeTag,
+    pub value_type: TypeTag,
+}
+
+impl TableInfo {
+    pub fn new(key_type: TypeTag, value_type: TypeTag) -> Self {
+        Self {
+            key_type,
+            value_type,
+        }
+    }
+}
+
+impl Display for TableInfo {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "Table<{}, {}>", self.key_type, self.value_type)
+    }
+}
+
 /// A table change set.
 #[derive(Default)]
 pub struct TableChangeSet {
-    pub new_tables: BTreeSet<TableHandle>,
+    pub new_tables: BTreeMap<TableHandle, TableInfo>,
     pub removed_tables: BTreeSet<TableHandle>,
     pub changes: BTreeMap<TableHandle, TableChange>,
 }
@@ -116,7 +138,7 @@ const _NOT_EMPTY: u64 = (102 << 8) + _ECATEGORY_INVALID_STATE as u64;
 /// of the overall context so we can mutate while still accessing the overall context.
 #[derive(Default)]
 struct TableData {
-    new_tables: BTreeSet<TableHandle>,
+    new_tables: BTreeMap<TableHandle, TableInfo>,
     removed_tables: BTreeSet<TableHandle>,
     tables: BTreeMap<TableHandle, Table>,
 }
@@ -336,9 +358,10 @@ pub fn table_natives(table_addr: AccountAddress) -> NativeFunctionTable {
 
 fn native_new_table_handle(
     context: &mut NativeContext,
-    mut _ty_args: Vec<Type>,
+    ty_args: Vec<Type>,
     args: VecDeque<Value>,
 ) -> PartialVMResult<NativeResult> {
+    assert_eq!(ty_args.len(), 2);
     assert!(args.is_empty());
 
     let table_context = context.extensions().get::<NativeTableContext>();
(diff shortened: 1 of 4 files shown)


wrwg:
Note that events moving to nursery actually means its dropped as general Move support. It is just temporarily in the nursery during transition. You could make this change directly in aptos_core, because we eventually nuke the nursery in aptos, and put types like events or others we need from it into our library.

sblackshear:
Agreed. I don't mind making the change here, but would also be ok with just deleting events from the nursery. I think we understand now that event structure + ID's make more sense as a platform-specific concept than a generic one in the stdlib.

davidiw:
Well I just repurposed this PR, so uh, gonna delete the comments above :P

wrwg (commented):
Well done! Do you have a way to test this?

language/extensions/move-table-extension/src/lib.rs
```diff
@@ -49,10 +49,31 @@ impl Display for TableHandle {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct TableInfo {
+    pub key_type: MoveTypeLayout,
+    pub value_type: MoveTypeLayout,
```
msmouse, line 55: I think a type tag is what we need..

davidiw:
Hahaha, I was hoping you weren't going to ask that... I did println the stuff from the existing test and it looked pretty good :).

any thoughts on a more thoughtful testing scheme -- short of we know that this will have the type info and we know how to use the type info via a convert to move forward?

language/extensions/move-table-extension/src/lib.rs
```diff
@@ -352,7 +374,9 @@ fn native_new_table_handle(
     Digest::update(&mut digest, table_data.new_tables.len().to_be_bytes());
     let bytes: [u8; 16] = digest.finalize()[0..16].try_into().unwrap();
     let id = u128::from_be_bytes(bytes);
-    assert!(table_data.new_tables.insert(TableHandle(id)));
+    let key_type = get_type_layout(context, &ty_args[0])?;
+    let value_type = get_type_layout(context, &ty_args[1])?;
```
msmouse, line 378: I'm under the impression `ty_args[2]` is the value type, but I don't remember why.

language/extensions/move-table-extension/src/lib.rs
```diff
@@ -49,10 +49,31 @@ impl Display for TableHandle {
...
+        Self {
+            key_type,
+            value_type,
+        }
+    }
+}
+
+impl Display for TableInfo {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "T<{}, {}>", self.key_type, self.value_type)
```
msmouse, line 69: `Table<{}, {}>`

language/extensions/move-table-extension/src/lib.rs
msmouse: I'm stupid. ignore me.

language/extensions/move-table-extension/src/lib.rs
davidiw: That may be a better choice. Let me look at what that spits out:

https://github.com/aptos-labs/aptos-core/blob/main/api/types/src/convert.rs#L468-L482

davidiw: {TableHandle(294284197927973660676668354886091714884): TableInfo { key_type: U64, value_type: Struct(StructTag { address: 00000000000000000000000000000002, module: Identifier("TableTests"), name: Identifier("Balance"), type_params: [] }) }}
[ PASS    ] 0x2::TableTests::test_remove_fail
{TableHandle(294284197927973660676668354886091714884): TableInfo { key_type: U64, value_type: U64 }}
[ PASS    ] 0x2::TableTests::test_remove_removed
{TableHandle(294284197927973660676668354886091714884): TableInfo { key_type: Address, value_type: Struct(StructTag { address: 00000000000000000000000000000002, module: Identifier("TableTests"), name: Identifier("Balance"), type_params: [] }) }}
[ PASS    ] 0x2::TableTests::test_struct
{TableHandle(294284197927973660676668354886091714884): TableInfo { key_type: Address, value_type: Struct(StructTag { address: 00000000000000000000000000000002, module: Identifier("Table"), name: Identifier("Table"), type_params: [Address, U128] }) }}
{TableHandle(187267989122364641638725157156538820950): TableInfo { key_type: Address, value_type: U128 }, TableHandle(294284197927973660676668354886091714884): TableInfo { key_type: Address, value_type: Struct(StructTag { address: 00000000000000000000000000000002, module: Identifier("Table"), name: Identifier("Table"), type_params: [Address, U128] }) }}
{TableHandle(160937847609835965615946043614408101562): TableInfo { key_type: Address, value_type: U128 }, TableHandle(187267989122364641638725157156538820950): TableInfo { key_type: Address, value_type: U128 }, TableHandle(294284197927973660676668354886091714884): TableInfo { key_type: Address, value_type: Struct(StructTag { address: 00000000000000000000000000000002, module: Identifier("Table"), name: Identifier("Table"), type_params: [Address, U128] }) }}
[ PASS    ] 0x2::TableTests::test_table_of_tables

yeah looks better and this will give us type / field names

language/tools/move-unit-test/src/extensions.rs
```diff
@@ -72,7 +72,7 @@ fn print_table_extension<W: Write>(w: &mut W, extensions: &mut NativeContextExte
             writeln!(
                 w,
                 "new tables {}",
-                cs.new_tables.iter().map(|h| h.to_string()).join(", ")
+                cs.new_tables.keys().map(|h| h.to_string()).join(", ")
```
zekun000, line 75: can we print the type info as well?

zekun000 (approved):

wrwg (approved):

PR #274, opened 2022-07-18
Change move syntax highlighting

uvd:
Change  syntax highlighting to move

## Motivation

Change move syntax highlighting 

Yes

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,5 +11,5 @@
 *.txt text
 *.yml text
 
-# Use Rust syntax highlighter for Move and Move IR code
+# Use Move syntax highlighter for Move IR code
 *.mvir linguist-language=Move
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -11,5 +11,5 @@
 *.txt text
 *.yml text
 
-# Use Rust syntax highlighter for Move and Move IR code
+# Use Move syntax highlighter for Move IR code
```
tnowacki, line 14: Do we need to do something for `*.move` files? or is that covered by default?

tnowacki (approved):

.gitattributes
uvd: > Do we need to do something for `*.move` files? or is that covered by default?

*.move files will use github  default syntax highlighting for Move

PR #278, opened 2022-07-18
[WIP][move-vm] allow gas metering to be customized

vgao1996:
**Note: this is still a WIP and some things don't work yet.**

## Overview
This is a major change to the gas metering system that
1. Adds a trait `GasMeter` that allows gas metering to be fine-tuned by the clients
2. Makes native functions closures, allowing them to carry their own contexts, which is now necessary for custom gas parameters
3. Moves the original `GasStatus` and related definitions from `move_vm_types/move_core_types` into `move_vm_test_utils`
4. Changes how native functions are constructed -- you now have maker functions that takes gas parameters and return closures

## Design Philosophy
1. Gas costs for both instructions and native functions should be customizable by the clients (adapters)
2. Gas customization for native functions should be provided by the packages, not the VM. This is because gas metering is part of the operational semantics of the native functions, and may be highly specific to the exactly algorithms used.
3. Native functions, including their corresponding gas parameters, should be composable
	a. For example, we shouldn't need to know the indices assigned to the native functions in other crates
4. It shouldn't really matter to the Move VM how exactly gas is measured and tracked. The VM just needs to let the client know when certain operations have been executed, and the client should let the VM know if it should stop the execution.

## Future Plans
1. Pass a mutable reference to the gas meter to the native functions, so they can charge gas during computation
2. Make the native function makers parametric over the gas meter types so we can get rid of `charge_in_native_unit`. Once this is done, the Move VM will be 100% agnostic of the gas meter implementation, allowing the clients to pick whatever suitable for them.
3. Make the gas metering for native functions more customizable. (Right now they use fixed formulae and you are only allowed to tune some of the parameters.)

## Benchmarks
I did some crude benchmarks using `time cargo test -p move-stdlib`:
```
Before
________________________________________________________
Executed in  890.49 millis    fish           external 
   usr time  966.87 millis    0.00 micros  966.87 millis 
   sys time  131.50 millis  661.00 micros  130.84 millis 

________________________________________________________
Executed in  896.15 millis    fish           external 
   usr time  978.78 millis  389.00 micros  978.39 millis 
   sys time  123.88 millis  257.00 micros  123.62 millis 

________________________________________________________
Executed in  907.73 millis    fish           external 
   usr time  961.42 millis  430.00 micros  960.99 millis 
   sys time  143.52 millis  286.00 micros  143.23 millis 

________________________________________________________
Executed in  883.58 millis    fish           external 
   usr time  962.73 millis  425.00 micros  962.30 millis 
   sys time  127.95 millis  282.00 micros  127.67 millis 

After
________________________________________________________
Executed in  907.26 millis    fish           external 
   usr time  928.28 millis  674.00 micros  927.61 millis 
   sys time  169.44 millis    0.00 micros  169.44 millis

________________________________________________________
Executed in  915.07 millis    fish           external 
   usr time  977.16 millis  675.00 micros  976.49 millis 
   sys time  132.85 millis    0.00 micros  132.85 millis 

________________________________________________________
Executed in  894.62 millis    fish           external 
   usr time  999.22 millis  646.00 micros  998.57 millis 
   sys time  108.85 millis    0.00 micros  108.85 millis 

________________________________________________________
Executed in  896.96 millis    fish           external 
   usr time  923.23 millis    0.00 micros  923.23 millis 
   sys time  172.13 millis  653.00 micros  171.48 millis 
```
This method is definitely not ideal, but at least I'm not observing any drastic performance differences.
I may do some more refined benchmarks if I have time.

**Note: this is still a WIP and these things don't work yet:**
- Some tests/tools now have all gas parameters set to zero
- Native functions are now boxed using Arc, only because a test assumes we need to have the ability to access them from multiple threads, which is something we may want to reconsider
- Documentation


language/move-stdlib/src/natives/string.rs
```diff
@@ -51,33 +81,73 @@ pub fn native_is_char_boundary(
...
     let s_ref = s_arg.as_bytes_ref();
     let s_str = unsafe {
         // This is safe because we guarantee the bytes to be utf8.
         std::str::from_utf8_unchecked(s_ref.as_slice())
     };
     let v = Value::vector_u8((&s_str[i..j]).as_bytes().iter().cloned());
-    // TODO: extensible native cost tables
-    let cost = native_gas(context.cost_table(), NativeCostIndex::PUSH_BACK, 1);
+
+    let cost = gas_params.base_cost + gas_params.unit_cost * (j - i) as u64;
```
vgao1996, line 127: The old formula is semantically equivalent to `unit_cost * max(j - i, 1)`. I wonder if it's okay for us to make a change like this, that makes it impossible to maintain fully backward compatibility, no matter the values you pick for your new gas parameters?

PR #281, opened 2022-07-19
[move-vm] allow gas metering to be customized

vgao1996:
## Overview
This is a major change to the gas metering system that
1. Adds a trait `GasMeter` that allows gas metering to be fine-tuned by the clients
2. Makes native functions closures, allowing them to carry their own contexts, which is now necessary for custom gas parameters
3. Moves the original `GasStatus` and related definitions from `move_vm_types/move_core_types` into `move_vm_test_utils`
4. Changes how native functions are constructed -- you now have maker functions that takes gas parameters and return closures

## Design Philosophy
1. Gas costs for both instructions and native functions should be customizable by the clients (adapters)
2. Gas customization for native functions should be provided by the packages, not the VM. This is because gas metering is part of the operational semantics of the native functions, and may be highly specific to the exactly algorithms used.
3. Native functions, including their corresponding gas parameters, should be composable
	a. For example, we shouldn't need to know the indices assigned to the native functions in other crates
4. It shouldn't really matter to the Move VM how exactly gas is measured and tracked. The VM just needs to let the client know when certain operations have been executed, and the client should let the VM know if it should stop the execution.

## Future Plans
1. Pass a mutable reference to the gas meter to the native functions, so they can charge gas during computation
2. Make the native function makers parametric over the gas meter types so we can get rid of `charge_in_native_unit`. Once this is done, the Move VM will be 100% agnostic of the gas meter implementation, allowing the clients to pick whatever suitable for them.
3. Make the gas metering for native functions more customizable. (Right now they use fixed formulae and you are only allowed to tune some of the parameters.)

## Benchmarks
I did some crude benchmarks using `time cargo test -p move-stdlib`:
```
Before
________________________________________________________
Executed in  890.49 millis    fish           external 
   usr time  966.87 millis    0.00 micros  966.87 millis 
   sys time  131.50 millis  661.00 micros  130.84 millis 

________________________________________________________
Executed in  896.15 millis    fish           external 
   usr time  978.78 millis  389.00 micros  978.39 millis 
   sys time  123.88 millis  257.00 micros  123.62 millis 

________________________________________________________
Executed in  907.73 millis    fish           external 
   usr time  961.42 millis  430.00 micros  960.99 millis 
   sys time  143.52 millis  286.00 micros  143.23 millis 

________________________________________________________
Executed in  883.58 millis    fish           external 
   usr time  962.73 millis  425.00 micros  962.30 millis 
   sys time  127.95 millis  282.00 micros  127.67 millis 

After
________________________________________________________
Executed in  907.26 millis    fish           external 
   usr time  928.28 millis  674.00 micros  927.61 millis 
   sys time  169.44 millis    0.00 micros  169.44 millis

________________________________________________________
Executed in  915.07 millis    fish           external 
   usr time  977.16 millis  675.00 micros  976.49 millis 
   sys time  132.85 millis    0.00 micros  132.85 millis 

________________________________________________________
Executed in  894.62 millis    fish           external 
   usr time  999.22 millis  646.00 micros  998.57 millis 
   sys time  108.85 millis    0.00 micros  108.85 millis 

________________________________________________________
Executed in  896.96 millis    fish           external 
   usr time  923.23 millis    0.00 micros  923.23 millis 
   sys time  172.13 millis  653.00 micros  171.48 millis 
```
This method is definitely not ideal, but at least I'm not observing any drastic performance differences.
I may do some more refined benchmarks if I have time.

**The PR is mostly complete but there are still some minor things I may want to revisit**
- Native functions are now boxed using Arc, only because a test assumes we need to have the ability to access them from multiple threads, which is something we may want to reconsider


vgao1996:
@wrwg @tnowacki I did something stupid and had to close the previous PR.. Please use this one instead.

language/move-vm/types/src/gas.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// Charge a given amount in the unit of measurement native to the GasMeter implementation.
+    /// Should fail if not enough gas units are left.
+    ///
+    /// This is used for metering native functions currently.
+    /// However, in the future, we may want to remove this and directly pass a reference to the GasMeter
+    /// instance to the native functions to allow gas to be deducted during computation.
+    fn charge_in_native_unit(&mut self, amount: u64) -> PartialVMResult<()>;
+}
+
+impl GasMeter for () {
```
tnowacki, line 27: Why?

language/move-vm/runtime/src/native_functions.rs
```diff
@@ -8,23 +8,27 @@ use crate::{
...
-pub type NativeFunction =
-    fn(&mut NativeContext, Vec<Type>, VecDeque<Value>) -> PartialVMResult<NativeResult>;
+pub type UnboxedNativeFunction = dyn Fn(&mut NativeContext, Vec<Type>, VecDeque<Value>) -> PartialVMResult<NativeResult>
+    + Send
+    + Sync
+    + 'static;
+
+pub type NativeFunction = Arc<UnboxedNativeFunction>;
 
 pub type NativeFunctionTable = Vec<(AccountAddress, Identifier, Identifier, NativeFunction)>;
```
tnowacki, line 33: I do not understand this change. I was expecting to see something much simpler. Just 
```
pub type NativeFunctionTable = Vec<(AccountAddress, Identifier, Identifier, NativeFunction, GasCost)>;
```

And then we could add the `GasCost` to the `NativeContext`

tnowacki (commented):
Beginning to peel the onion here. Just some starter questions in the files you pointed at! (thanks for that btw)

language/move-vm/runtime/src/native_functions.rs
vgao1996: Good question. My original plan was to add a single `GasCost`, but I ended up implementing a more generalized version based on the following reasoning: 
1. Only a package knows how its gas formulae can be customized and what gas parameters are needed, so ideally, the package themselves should have the freedom (and responsibility) to provide such customization to its own clients, in ways that make sense to the specific algorithms its native functions use.
2. From the Move VM's perspective, we really don't care how a native function wants to meter its own gas usage. The current protocol just requires it to return a number. Mandating the usage of a single `GasCost` is both restrictive and adds unnecessary coupling. I think in the general case, it's quite reasonable for a native function to have more than 1 gas parameter and for different native functions to have different number of gas parameters, potentially with different types.

language/move-vm/types/src/gas.rs
tnowacki: My concern is bumping into this without meaning too. I'd just prefer something a bit more intentional, like `EmptyGasMeter` or `UncheckedGasMeter`

language/move-vm/runtime/src/native_functions.rs
tnowacki: I'd prefer to keep this a bit simpler unless you have a hard use case. Reading this I feel like it really complicates the code. I get that it is "more general", but I feel like our goal should be to serve the common use cases we have. Really gnarly use cases can always dive into `NativeExtensions` and stuff gas data into that.

I don't understand this unnecessary coupling comment quite fully. But if you want something a little bit more flexible...
Given `GasCost` is `pub struct GasCost { pub instruction_gas: InternalGasUnits<GasCarrier>, pub memory_gas: InternalGasUnits<GasCarrier>, }`

We could do something like `pub struct NativeGasCost { pub instruction_gas: Option<InternalGasUnits<GasCarrier>>, pub memory_gas: Option<InternalGasUnits<GasCarrier>>, }`

wrwg: Pulling in @davidiw and @zekun000 / @msmouse; also interested in @sblackshear opinion.

I agree with @tnowacki that this looks complex at the first sight. However, I think it is easier to use then apparent from this PR. That is because the design pattern Victor choose is clear and easy to follow.

Basically, if you define a native, you also have to define a data structure for representing the parameters for the costing. Like here:

```
pub struct MyNativeGasParameters {
    pub base_cost: u64,
    pub unit_cost: u64,
}

fn my_native(
    gas_params: &MyNativeGasParameters,
    _context: &mut NativeContext,
    _ty_args: Vec<Type>,
    mut args: VecDeque<Value>,
) -> PartialVMResult<NativeResult> {
    ...
    let cost = gas_params.base_cost + gas_params.unit_cost * <size of data>;
    NativeResult::map_partial_vm_result_one(cost, ...)
}
```

Then when you integrate this as a new native into a Move package, you do something like

```
struct PackageGasParameters {
   // existing parameters
   ...
   my_native_params: MyNativeGasParameters
```

I feel confident that this does not make the introduction of new functions significantly more difficult than it already is. 

The problem this solves, really, is that the computation of the cost of some operations cannot be done upfront just on basis of the input parameters. At least, not efficiently -- computing the cost can be as expensive as doing the operation. Consider for example a table read. We may want to cost a table read differently based on whether it is the first or subsequent access in a given txn. So instead of hacking around this for tables and doing it somehow over the native context extension, Victor's design provides a general clean model for costs of natives which can be applied to any current and potential future use cases.

vgao1996: I hear your concerns. Let me explain my thoughts from the perspective of each individual component:
1. For the Move VM
  a. The point of the decoupling is to give the control of gas to the individual crates, and it's up to them to choose how they want to customize their gas formula. The VM supports all options: complex, simple, or even non-customizable at all. This way the VM's own logic remains simple. I understand your argument that passing in a single GasCost would cover a lot of use cases, but it still feels a bit like a leaky abstraction to me. (My current move-stdlib implementation opted for the complex path and probably left you the bad impression that this whole thing is inherently complex. I'll address this later)
  b. The change from `fn(...)` to `Arc<dyn Fn(...)>` actually results in very minimal changes to the VM code itself -- only a handful lines were are to this. In terms of performance, this does add a virtual function call, but I highly doubt if this is noticeable at all. The `Arc` is only used to make the existing loader implementation happy and satisfy its questionable assumption that the function needs to be accessed by multiple threads.
  c. I think you made a good point about `NativeExtensions`. The table extension for example, gets its gas parameters this way. My concern about this however, is that this prevents code inlining and other compiler optimizations compared to the closure approach.
2. For move-stdlib (or other packages with native functions)
  a. As I hinted above, it is my current implementation for this package that opted for the more complex path, defining gas parameters for every function and every upper layer (module/crate). This does grant us maximum composability but I also feel like it's a bit too cumbersome. I would be open to switch to a simpler model. (e.g. `fn make_natives(address: AccountAddress, gas_costs: &[GasCost])` + getting rid the individual gas parameter structs and makers)
  b. See `examples/diem-framework/natives` for a simpler (actually now non-configurable) example. There is very little added code to make it work with the new system. Again, I would argue that the complexity is really an issue on the package side -- in general, the more composable/customizable you want your package to be, the more complex your boilerplate code will be.

language/move-stdlib/src/natives/hash.rs
```diff
@@ -26,11 +34,7 @@ pub fn native_sha2_256(
 
     let hash_arg = pop_arg!(arguments, Vec<u8>);
 
-    let cost = native_gas(
-        context.cost_table(),
-        NativeCostIndex::SHA2_256,
-        hash_arg.len(),
-    );
+    let cost = gas_params.base_cost + gas_params.unit_cost * usize::max(hash_arg.len(), 1) as u64;
```
vgao1996, line 37: A general question: am I allowed to change a gas formula so that there doesn't exist any configuration that would allow it to maintain full backward compatibility? In other words, can I make breaking changes?

We have `unit_cost * max(len, 1)` in a few places and I kinda want to replace them with `base_cost + unit_cost * len`

language/move-vm/runtime/src/native_functions.rs
wrwg: > I would be open to switch to a simpler model. (e.g. fn make_natives(address: AccountAddress, gas_costs: &[GasCost]) + getting rid the individual gas parameter structs and makers)

I feel not good with this. This is optimizing friction at the wrong level of abstraction. We aren't designing APIs for end users.

We should make this canonical and maximally expressive. There is some boilerplate, but its easy to create as long as you have some existing code to start from.

language/move-stdlib/src/natives/hash.rs
tnowacki: I think it is fine for this case? As if it is an issue for anyone they can write their own native.

But we might want to just stick two versions of the function in this file to help people?

language/move-vm/runtime/src/native_functions.rs
vgao1996: FYI, I did try to write a macro that would save us from having to write the boilerplate code and it worked:
```
define_native!{
    name = foo;

    gas_params = {
        base_cost: u64,
    }

    (gas_params, context, ty_args, args) => {
        ...
    }
}
```
The problem however, is that we lose move-analyzer's auto-formatting and type annotation, which made me give this up.

tnowacki: @wrwg 
> I feel confident that this does not make the introduction of new functions significantly more difficult than it already is.

I tend to agree. My only observation was that most of these structs are more or less just a `GasCost`. And in the cases they aren't, we could maybe get away with a similar version of `GasCost` but where each field is an `Option`.

I guess what I'm trying to say is: it would be nice if we don't have to define these structs (and from looking at the examples it feels a bit like overkill), but it is not the end of the world if we have to.

language/move-stdlib/src/natives/hash.rs
vgao1996: I think in the short term, I can just add another gas parameter `min_len` to allow the new formula to be compatible with the old one.

This doesn't answer my original question though. We still need to decide come up with a general guideline/policy on how we handle breaking changes to the gas formulae.

vgao1996: @jolestar Any comments on this? You guys are ahead of us in terms of running a production network and I wonder if you have any insights to share.

jolestar: If the gas formula cost != cost_table cost, will break the compatibility. But I think the formula approach is better than the cost table, so we look forward to seeing this change.

We can try to adjust the gas params and let the two costs be the same, or just fork the `native_sha2_256` and write some code like if chain height > xxx for compatibility.

language/move-vm/runtime/src/native_functions.rs
jolestar: The `PackageGasParameters` is easier to read than cost_table, but how we should keep the `PackageGasParameters`? On Starcoin, we keep the `cost_table` to Move module.

https://github.com/starcoinorg/starcoin-framework/blob/d3bb99ff026e408341377ecd73ce37108c278066/sources/VMConfig.move#L224-L233

This approach makes it easier to add a new native method or update the gas cost of a native method.

If the GasParameters is a struct, it is harder to extend than vector.

language/move-stdlib/src/natives/hash.rs
vgao1996: Sounds good. Also had a discussion with @davidiw and we came to the conclusion that we should treat gas formula changes just like any other breaking changes, try to maintain backward compatibility for a transient period at least. This is to allow node operators to update their rust binaries in an asynchronous manner.

vgao1996:
I updated the PR and here's a summary of the new changes:
1. Some `legacy_*` gas parameters are added to provide backward compatibility for move-stdlib natives.
2. Nursery native functions are split out into a separate list (`fn nursery_natives`)
3. I documented the gas formulae for each move-stdlib native in the comment section
4. I reviewed the tests/tools that use native functions/the gas meter and identified some issues. While I'm not planning to do anything in this PR, I've wrote down some of my analysis as TODOs.

Let me know if this looks good to you now!

language/move-vm/runtime/src/native_functions.rs
vgao1996: @jolestar this is actually a very good question. 
1. The reason why I opted for a PackageGasParameters struct is because I wanted to have different gas parameters for different native functions.
2. I should make it clear that this struct is only intended to be used as an interface to configure the move-stdlib natives. You don't necessarily have to use it in other parts of your program or store it on-chain. You can and are encouraged to use your own format if you find that more convenient. It might be helpful to write a conversion function from your format to PackageGasParameters. We are also planning to do something similar in the Aptos framework: we'll define our own on-chain gas schedule format, which will likely be vector-based so it can be extended in the future.

language/move-vm/types/src/natives/function.rs
```diff
@@ -36,15 +33,14 @@ pub use move_core_types::vm_status::StatusCode;
...
-    pub cost: InternalGasUnits<GasCarrier>,
     /// Result of execution. This is either the return values or the error to report.
+    pub cost: u64,
     pub result: Result<SmallVec<[Value; 1]>, u64>,
 }
 
 impl NativeResult {
     /// Return values of a successful execution.
-    pub fn ok(cost: InternalGasUnits<GasCarrier>, values: SmallVec<[Value; 1]>) -> Self {
+    pub fn ok(cost: u64, values: SmallVec<[Value; 1]>) -> Self {
```
tnowacki, line 43: Why the removal of `InternalGasUnits`?

language/move-stdlib/src/natives/mod.rs
```diff
@@ -7,90 +7,164 @@ pub mod event;
...
+    signer: signer::GasParameters,
+    string: string::GasParameters,
+    vector: vector::GasParameters,
+
+    #[cfg(feature = "testing")]
+    unit_test: unit_test::GasParameters,
+}
+
+impl GasParameters {
+    pub fn zeros() -> Self {
```
tnowacki, line 35: What is this used for in most cases? Can we copy over the initial schedule from whatever was there before? Or is there some reason not to?

language/move-stdlib/src/natives/string.rs
```diff
@@ -24,8 +23,21 @@ use std::collections::VecDeque;
 // create a `&str` view on the bytes without a copy. Once we have this
 // view, we can call ut8 functions like length, substring, etc.
 
-pub fn native_check_utf8(
-    context: &mut NativeContext,
+/***************************************************************************************************
```
tnowacki, line 26: I appreciate the comments :) But they could maybe be a tad smaller ;) Not needed to change though if you like it. 

I just feel the size maybe outweighs the importance. A `///` might be fine enough

language/move-stdlib/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "move-stdlib"
 version = "0.1.1"
-edition = "2018"
+edition = "2021"
 authors = ["Diem Association <opensource@diem.com>"]
 description = "Diem stdlib"
 repository = "https://github.com/diem/diem"
```
tnowacki, line 7: Unrelated but we should probably change this lol

language/move-core/types/src/gas_schedule.rs
```diff
@@ -151,127 +151,5 @@ pub const REFERENCE_SIZE: AbstractMemorySize<GasCarrier> = AbstractMemorySize(8)
...
-/// Any transaction over this size will be charged `INTRINSIC_GAS_PER_BYTE` per byte
-pub const LARGE_TRANSACTION_CUTOFF: AbstractMemorySize<GasCarrier> = AbstractMemorySize(600);
-
 /// For exists checks on data that doesn't exists this is the multiplier that is used.
 pub const MIN_EXISTS_DATA_SIZE: AbstractMemorySize<GasCarrier> = AbstractMemorySize(100);
-
-pub const MAX_TRANSACTION_SIZE_IN_BYTES: GasCarrier = 4096;
-
-#[derive(Clone, Debug, Serialize, PartialEq, Deserialize)]
-pub struct GasConstants {
```
tnowacki, line 166: I saw the change elsewhere, but didn't quite grok, why the move?

language/move-vm/types/src/gas.rs
```diff
@@ -0,0 +1,47 @@
...
+    ///
+    /// This is used for metering native functions currently.
+    /// However, in the future, we may want to remove this and directly pass a reference to the GasMeter
+    /// instance to the native functions to allow gas to be deducted during computation.
+    fn charge_in_native_unit(&mut self, amount: u64) -> PartialVMResult<()>;
+}
+
+/// A dummy gas meter that does not meter anything.
+/// Charge operations will always succeed.
+pub struct EmptyGasMeter;
```
tnowacki, line 29: Ah! Random thought based on seeing this usage in tests: thoughts on naming this `UnmeteredGasMeter` :p  Or maybe `UnmeteredGas`? 

This name is fine too though

tnowacki (commented):
Things are shaping up pretty well! Some operational questions mostly, and then this should be good to go!

language/move-vm/types/src/gas.rs
vgao1996: I also felt that `EmptyGasMeter` was a bit weird. `UnmeteredGasMeter` seems better. It's a bit long but at least explicit.

language/move-vm/types/src/natives/function.rs
vgao1996: Previously we've been doing a lot of gas calculations inside the VM, so there was a need to not accidentally mix different gas values up. However with the new system, it is up to the GasMeter implementations to perform the actual bookkeeping, and in this particular case, the VM is simply passing whatever value native functions returns to the GasMeter. I do find the GasAlgebra monad unnecessarily complicated for our current needs so I just decided to use a plain integer instead.

language/move-stdlib/src/natives/mod.rs
vgao1996: Good question. 

> Can we copy over the initial schedule from whatever was there before?
I thought about doing this previously and actually listed this as a work item. However after inspecting the values, it felt to me that some of the values were so absurd and inconsistent that I was no longer sure if I should even bother keeping this old schedule. Here are some examples:
1. `vector::swap` costs 1437 gas units -- for simply doing a shallow swap? `vector::borrow` costs 1335 -- for creating a simple reference? In comparison, `vector::empty` costs 85; `vector::length` costs 99; `vector::destroy_empty` costs 573.
  a. It's weird that operations that allocate memories cost even less than some of the ones that only touches a few words.
2. `vector::push_back` costs 54 * size_of(val being pushed) while `vector::pop_back` costs a constant 228 units
3. The string natives all costs some constant numbers even though most operations are linear

I took a look at the callsites of `GasParameters::zeros` and it seemed to me they can be classified into 3 types:
1. Tests that don't care about gas metering -- using the all-zero gas schedule is the right thing to do
2. Tests that may care about gas metering but don't yet -- we can come up with something that actually makes sense to them when we start to care
3. Tools (move-cli/unit tests) -- these really shouldn't have a hardcoded gas schedule. I have left some TODOs to remind us that we'll need to redesign these parts in the future.

language/move-core/types/src/gas_schedule.rs
vgao1996: With the addition of the `trait GasMeter`, adapters are now likely to implement their own gas meters. The `GasStatus` carries a lot of legacies and is quite complicated TBH. I want to avoid giving clients the impression that we still have this official `GasMeter` implementation, at least before we have the time to clean it up a bit.

language/move-stdlib/Cargo.toml
vgao1996: Looks like all of our crates still have these... Let's do a codemod and get them replaced

language/documentation/examples/diem-framework/crates/natives/src/account.rs
```diff
@@ -6,38 +6,32 @@ use move_binary_format::errors::PartialVMResult;
...
     ty_args: Vec<Type>,
     mut arguments: VecDeque<Value>,
 ) -> PartialVMResult<NativeResult> {
     debug_assert!(ty_args.is_empty());
     debug_assert!(arguments.len() == 1);
 
     let address = pop_arg!(arguments, AccountAddress);
-    let cost = native_gas(context.cost_table(), NativeCostIndex::CREATE_SIGNER, 0);
-    Ok(NativeResult::ok(cost, smallvec![Value::signer(address)]))
+    Ok(NativeResult::ok(25, smallvec![Value::signer(address)]))
```
movekevin, line 23: Why hardcode to 25 here?

movekevin: Same question for other changes below

vgao1996: This is old diem-framework code we imported for testing purposes. We don't need fancy customization for it.

language/move-vm/types/src/natives/function.rs
tnowacki: Ah, I see. 

I still have some worry with accidentally using unchecked arithmetic or something. And it might be beneficial to use the InternalGasUnits for that. But not sure it is worth the pain overall

tnowacki: Thinking a bit more on this:
Not using the GasUnits system means you are going to have to manually re-check all the arithmetic for over/underflow (and any div-by-zero, though this is less likely). 
This is definitely an attack vector to consider, especially around things like hash/string/bcs/vector etc where you might be dealing with rather large sizes.

I think we either need to:
- Switch back to InternalGasUnits
- Or have some other mechanism for doing this simply inside the native functions
While it might not entirely be possible, I think it is a bit of a safety regression when compared to what was there before

tnowacki: Turns out InternalGasUnits and the GasAlgebras in general are not checked. Wild

tnowacki (approved):
Let's ship it. There are some things to be worked out down the road for the whole overflow/underflow issue. But that issue is already there

PR #287, opened 2022-07-20
[move-vm] allow casting from Value to Vec<Value>

vgao1996:
This adds a missing cast from `Value` to `Vec<Value>` so that native functions can access the elements of non-specialized vectors.


wrwg (approved):

tnowacki (approved):
Kind of amazing we didn't have this yet, lol

PR #288, opened 2022-07-20
[stackless bytecode] fix a bug in VecPack and VecUnpack

emmazzz:
## Motivation

This PR fixes a bug in stackless bytecode generation of vector bytecode instructions. 

Previously, `VecPack` is translated to:
```rust
let temp = vector::new();
for element in operands {
    vector::push_back(temp, element);
}
``` 
This is buggy because `vector::new()` returns a new vector object but `vector::push_back` takes in a mutable reference to a vector. This PR adds that missing `borrow_mut`:
```rust
let temp = vector::new();
let mut_ref = &mut temp;
for element in operands {
    vector::push_back(mut_ref, element);
}
``` 
Same reasoning applies to `VecUnpack`.

### DIFF
--- a/language/move-prover/bytecode/src/stackless_bytecode_generator.rs
+++ b/language/move-prover/bytecode/src/stackless_bytecode_generator.rs
@@ -1200,14 +1200,26 @@ impl<'a> StacklessBytecodeGenerator<'a> {
                     vec![],
                     None,
                 ));
-                for operand in operands {
-                    self.code.push(Bytecode::Call(
-                        attr_id,
-                        vec![],
-                        mk_vec_function_operation("push_back", vec![ty.clone()]),
-                        vec![temp_index, operand],
-                        None,
+                if !operands.is_empty() {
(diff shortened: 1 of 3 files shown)


language/move-prover/bytecode/src/stackless_bytecode_generator.rs
```diff
@@ -1200,14 +1200,26 @@ impl<'a> StacklessBytecodeGenerator<'a> {
...
-                        mk_vec_function_operation("push_back", vec![ty.clone()]),
-                        vec![temp_index, operand],
-                        None,
+                if !operands.is_empty() {
+                    let mut_ref_index = self.temp_count;
+                    self.local_types.push(Type::Reference(
+                        true,
+                        Box::new(Type::Vector(Box::new(ty.clone()))),
                     ));
+                    self.temp_count += 1;
```
wrwg, line 1209: At some point we should introduce the vector opcodes in the stackless bytecode.

wrwg (approved):

PR #284, opened 2022-07-19
[prover] repro test for borrow analysis bug

emmazzz:
## Motivation

This PR adds to prover test crate a test crafted by @awelc that reproduces the bug described by issue #270 .

Issue #270, opened 2022-07-16
[Bug] Move prover bytecode translator error


awelc:
# 🐛 Bug

I have discovered a problem with the bytecode translator on certain snippets of Move code (please see below for an example on how to reproduce this). I have done some investigation, and I am happy to continue digging, but if someone more familiar with the codebase could offer some help or advice, it would be great.

The problem manifests itself in the Boogie backend in [bytecode_translator.rs:1632](https://github.com/move-language/move/blob/c51b11589554c6acfef15c74fca211def4ea0731/language/move-prover/boogie-backend/src/bytecode_translator.rs#L1629) - the `inst` vector has length 0:
```rust
                    let (read_aggregate, update_aggregate, elem_ty) = match dst_ty {
                        Type::Vector(et) => ("ReadVec", "UpdateVec", et.as_ref().clone()),
                        // If its not a vector, we assume it is the Table type.
-->                     Type::Struct(_, _, inst) => ("GetTable", "UpdateTable", inst[1].clone()),
                        _ => unreachable!(),
                    };
```

The reason for this is that `BorrowEdge::Index` edge type is used to indicate borrowing both from a vector and from a table, and in this case it is incorrectly identified as the edge indicating borrowing from a table though no tables are used in the code.

I am pretty sure that the problem is in the pass responsible for borrow analysis, and I suspect it is related to how hyper edges summarize and propagate borrow information of entire functions to callers of these functions.

When you look at the output produced by borrow analysis pass for the `inner` function in the example below, you will see borrow information of this kind (notice index edge indicated by `[]`):
```
# borrowed_by: Reference($t0) -> {(.v (vector<m1::HasAnotherVector>), Reference($t1))}, Reference($t1) -> {([], Reference($t3))}
```
It makes perfect sense that as the index edge represents borrowing from a vector due to `$t1` and `$t3` types:
```
var $t1: &mut vector<m1::HasAnotherVector>
var $t3: &mut m1::HasAnotherVector
```

In the output for the `outer` function, we see something a bit different:
```
# borrowed_by: Reference($t0) -> {([]/.v (vector<m1::HasAnotherVector>), Reference($t2))}
```

I haven't yet figured out how hyper edges are used in the translation process, but the above may imply that `$t0` and `$t2` will be connected via an index edge eventually, which would be a problem due to their types, none of which is a vector:
```
fun m1::outer($t0|has_vector: &mut m1::HasVector) { ... }
var $t2: &mut m1::HasAnotherVector
```

[...]

PR #289, opened 2022-07-21
[move-prover] Fixed hyper edge ordering problem

awelc:
## Motivation

This is an attempt to fix the Move prover problem reported and analyzed in https://github.com/move-language/move/issues/270

Please note that the bytecode translator changes are failing - due to changes that have been made expected result no longer matches the expected output - if the solution is correct we should update expected output.

Admittedly this solution has been crated "by feel" rather than by a super-deep knowledge of the borrow analyzer, but perhaps even if it's not 100% correct, it will inspire a proper solution.

Some observations:
1. It seems to me that hyper edges in `borrowed_by` and `borrowed_from` sets should be reverted, which is currently not the case (hence edge revert operation in `consolidate`).
2. The failure in the bytecode translator for the example in the related issue (https://github.com/move-language/move/issues/270) is due to a hyper edge in the `borrowed_from` set being in the wrong order, which means that considering the bullet above, it's the `borrowed_from` set should keep the "correct" (from the bytecode translator's perspective) order
3. The final bit is that in constructing function annotation (in `summarize` and then `construct_hyper_edges`) we use `borrowed_from` edge ("correct" order) to construct a `borrowed_by` edge (incorrect order), so we have to reverse the hyper edge (but not any edge as before).

Yes

## Test Plan

I don't think we have Boogie backend tests, but with this PR in the test exposing the bug (added in https://github.com/move-language/move/pull/284) no longer crashes the prover.

### DIFF
--- a/language/move-prover/bytecode/src/borrow_analysis.rs
+++ b/language/move-prover/bytecode/src/borrow_analysis.rs
@@ -214,7 +214,7 @@ impl BorrowInfo {
                 self.borrows_from
                     .entry(dst.clone())
                     .or_default()
-                    .insert((src.clone(), edge.clone()));
+                    .insert((src.clone(), edge.clone().reverse()));
             }
         }
     }
@@ -260,7 +260,7 @@ impl BorrowInfo {
     ) {
         for (dest, edge) in outgoing.iter() {
             let mut path = prefix.to_owned();
-            path.extend(edge.flatten().into_iter().cloned());
+            path.extend(edge.clone().reverse().flatten().into_iter().cloned());
             if let Some(succs) = ret_info.borrows_from.get(dest) {
                 self.construct_hyper_edges(leaf, ret_info, path, succs);
             } else {
@@ -268,7 +268,6 @@ impl BorrowInfo {
                 let edge = if path.len() == 1 {
                     path.pop().unwrap()
                 } else {
-                    path.reverse();
                     BorrowEdge::Hyper(path)
                 };
                 self.borrowed_by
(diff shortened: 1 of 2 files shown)


emmazzz:
Thanks! Awesome investigation @awelc ! Your solution looks sensible to me but will let Wolfgang check and approve it.

language/move-prover/bytecode/src/borrow_analysis.rs
```diff
@@ -260,15 +260,14 @@ impl BorrowInfo {
...
-            path.extend(edge.flatten().into_iter().cloned());
+            path.extend(edge.clone().reverse().flatten().into_iter().cloned());
             if let Some(succs) = ret_info.borrows_from.get(dest) {
                 self.construct_hyper_edges(leaf, ret_info, path, succs);
             } else {
                 // Reached a leaf.
                 let edge = if path.len() == 1 {
                     path.pop().unwrap()
                 } else {
-                    path.reverse();
```
meng-xu-cs, line 271: Sorry for an unexpected chiming-in here:

I think the `reverse()` here is needed. The issue/bug is that `flatten()` is called too early.

Think about the case where we have
`path = [ Hyper(Index, Field) ]`

Doing a flatten-then-reverse yield
```
[ Hyper(Index, Field) ] --flatten--> [ Index, Field ] --reverse--> [ Field, Index ] --repack--> [ Hyper(Field, Index) ]
```
This will distort the original hyper-edge.

While doing a reverse-then-flatten yield
```
[ Hyper(Index, Field) ] --reverse--> [ Hyper(Index, Field) ] --flatten--> [ Index, Field ] --repack--> Hyper(Index, Field)
```
This preserves the original hyper-edge.

meng-xu-cs: The PR is here: #298

awelc:
Superseded by https://github.com/move-language/move/pull/298

PR #291, opened 2022-07-21
[move-cli] Added "docgen" command

villesundell:
Move Prover's `--docgen` switch is not currently exposed to the user, so added a top level command `docgen` to do the trick.

Move Prover is also mentioned in the description, so the user knows they must have a working Move Prover setup in order to use this feature.

Also added a smoketest for the `docgen` command so it would not break because of a regression. A non-existent subcommand `smoketest` is intentionally used so the output has as little variable output as possible.

## Motivation

Move Prover's `--docgen` switch is not currently exposed to the user, so added a top level command `docgen` to do the trick.

Yes.

## Test Plan

In addition to the smoke test (`cargo test -p move-cli`) you can test the actual documentation generation by running:
```
cargo build -p move-cli
./target/debug/move --path ./language/documentation/tutorial/step_8/BasicCoin docgen
```

You should get something like:

```
[INFO] generating documentation
[INFO] Module `0xcafe::BasicCoin` in file `doc/BasicCoin.md` will be generated
[INFO] 0.148s checking, 0.075s generating
```

And then you should have the final generated markdown documentation at `./language/documentation/tutorial/step_8/BasicCoin/doc/BasicCoin.md`.

(In order to use the docgen you must have a working Move Prover, verify this by running `./target/debug/move --path ./language/documentation/tutorial/step_8/BasicCoin prove`)

### DIFF
--- a/language/tools/move-cli/src/base/docgen.rs
+++ b/language/tools/move-cli/src/base/docgen.rs
@@ -1,0 +1,48 @@
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use crate::base::prove::{Prove, ProverOptions};
+use clap::*;
+use move_package::BuildConfig;
+use std::path::PathBuf;
+
+/// Generating javadoc style documentation for packages using the Move Prover
+#[derive(Parser)]
+#[clap(name = "docgen")]
+pub struct Docgen {
+    /// Do not print out other than errors
+    #[clap(long = "quiet", short = 'q')]
+    pub quiet: bool,
+    /// A template for documentation generation.
+    #[clap(long = "template", short = 't', value_name = "FILE")]
+    pub template: Option<String>,
+}
+
+impl Docgen {
+    /// Simply calling the Move Prover with "--docgen"
+    /// Forwarding `--docgen-template` and setting `--verbose` to `error` if applicable
+    pub fn execute(self, path: Option<PathBuf>, config: BuildConfig) -> anyhow::Result<()> {
+        let mut options_list = vec!["--docgen".to_string()];
+
+        if self.quiet {
+            options_list.push("--verbose".to_string());
+            options_list.push("error".to_string());
+        }
+        if self.template.is_some() {
+            options_list.push("--docgen-template".to_string());
+            options_list.push(self.template.unwrap());
+        }
+
+        let options = Some(ProverOptions::Options(options_list));
+
+        Prove::execute(
+            Prove {
+                target_filter: None,
+                for_test: false,
+                options,
+            },
+            path,
+            config,
+        )
+    }
+}
--- a/language/tools/move-cli/tests/build_tests/simple_build_with_docs/args.txt
+++ b/language/tools/move-cli/tests/build_tests/simple_build_with_docs/args.txt
@@ -1,0 +1,5 @@
+new --path . Foo
+build
+docgen --quiet --template template.md
+> grep documentation doc/Foo.md
+> cat doc/template.md
--- a/language/tools/move-cli/src/base/mod.rs
+++ b/language/tools/move-cli/src/base/mod.rs
@@ -4,6 +4,7 @@
 pub mod build;
 pub mod coverage;
 pub mod disassemble;
+pub mod docgen;
 pub mod errmap;
 pub mod info;
 pub mod new;
--- a/language/tools/move-cli/src/lib.rs
+++ b/language/tools/move-cli/src/lib.rs
@@ -3,8 +3,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use base::{
-    build::Build, coverage::Coverage, disassemble::Disassemble, errmap::Errmap, info::Info,
-    new::New, prove::Prove, test::Test,
+    build::Build, coverage::Coverage, disassemble::Disassemble, docgen::Docgen, errmap::Errmap,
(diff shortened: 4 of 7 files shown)


villesundell:
Because I can't add reviewers, asking review from my mentors @awelc and @tnowacki this way :memo:

tnowacki:
I believe the same functionality can be achieved with `build --generate-docs`?

tnowacki:
It is not unreasonable to make this it's own command though.

villesundell:
@tnowacki Unfortunately `build --generate-docs` doesn't exist anymore, and `--docs` is a no-op. `--docs` is burried so deeply in `move-cli` that it would need another PR to remove, and that would be logically separate from this change (there seems to be a lot of logic behind the switch, despite being effectively a no-op). I can work on such a PR if this gets merged (rendering `--docs` obsolete)

tnowacki:
> @tnowacki Unfortunately build --generate-docs

villesundell:
@tnowacki I get:

```
move build --generate-docs
error: Found argument '--generate-docs' which wasn't expected, or isn't valid in this context

	If you tried to supply `--generate-docs` as a value rather than a flag, use `-- --generate-docs`

USAGE:
    move build [OPTIONS]

For more information try --help

```

tnowacki:
Looks like it is `move build --doc` no s, go figure.

I agree though it's rather buried, and a top level command would be nice. Especially since it took me a while to get the thing working at all lol

I'll take a deeper look at the PR later today

language/tools/move-cli/src/base/docgen.rs
```diff
@@ -0,0 +1,28 @@
...
+
+/// Generating javadoc style documentation for packages using the Move Prover
+#[derive(Parser)]
+#[clap(name = "docgen")]
+pub struct Docgen;
+
+impl Docgen {
+    /// Simply calling the Move Prover with "--docgen"
+    pub fn execute(self, path: Option<PathBuf>, config: BuildConfig) -> anyhow::Result<()> {
+        let options = Some(ProverOptions::Options(vec!["--docgen".to_string()]));
```
tnowacki, line 17: I think there are a bunch of options for docgen; I feel like we should expose them in the CLI here. Thoughts?

tnowacki (commented):
Looks good but two big points:
- What do we do about tests?
- Should we expose all of the configs/arguments for the docgen?

And as a followup?
- What about --doc in the build config? Should we remove it or fix it?

language/tools/move-cli/src/base/docgen.rs
villesundell: @tnowacki What do you think, how stable these options will be in the future? Could we introduce `--` in order to forward extra options and avoid reduplication?

villesundell:
@tnowacki Good points! My takes:
1. Do you mean adding more tests?
2. (commented on the review comment)
3. I would vote for removing, but in another PR: to me `--docs` seems to be burried so deeply in move-cli that it would need another PR to remove, and that would be logically separate from this change (there seems to be a lot of logic behind the switch, despite being effectively a no-op). But I can gladly work on such a PR if this gets merged, and renders --docs obsolete.

language/tools/move-cli/src/base/docgen.rs
tnowacki: No idea on the stability of it. It seems the args are rather hand rolled. It would ben nice to get them in the same struct opt definition if possible. But not necessary for this PR.

I would be fine with just duplicating them all here in struct opt, and fixing it later. Feels a bit more user friendly for this command.

tnowacki:
1) Yes, it seems there are no tests that actually run it?

3) works for me :)

villesundell:
@tnowacki Added an actual test and `--template` (exposing `--docgen-template` to the user, which is currently the only switch exposed by the Prover)  :slightly_smiling_face: While I think `move docgen` should eventually use `Docgen::` instead of `Prove::` (and hence expose all the other options to the user), currently I see this current approach being the fastest way to fix Docgen for the regular user. I can start working next on a proper `Docgen::` implementation of `docgen`.

language/tools/move-cli/tests/build_tests/simple_build_with_docs/args.txt
```diff
@@ -0,0 +1,5 @@
+new --path . Foo
+build
+docgen --quiet --template template.md
+> grep documentation doc/Foo.md
+> grep template doc/template.md
```
tnowacki, line 5: Maybe just `cat` here?

tnowacki:
> @tnowacki Added an actual test and `--template` (exposing `--docgen-template` to the user, which is currently the only switch exposed by the Prover) 🙂 While I think `move docgen` should eventually use `Docgen::` instead of `Prove::` (and hence expose all the other options to the user), currently I see this current approach being the fastest way to fix Docgen for the regular user. I can start working next on a proper `Docgen::` implementation of `docgen`.

This seems like a fair approach, thanks for the explanation! 
Just a comment on the test, then I think this should be good to go!

villesundell:
@tnowacki Great! :partying_face: The test is now fixed, and is using `cat` instead of `grep` which makes sense :+1:

junkil-park (approved):
LGTM. Thanks! I approve it because it seems the latest commit seems to have addressed Todd's comment on tests.

tnowacki:
Thanks again for this!

PR #292, opened 2022-07-21
[Move analyzer] Added support for building and running on Windows

awelc:
## Motivation

This PR adds additional support for the Move analyzer to correctly build and run on Windows. In particular the "standard" version of path canonicalization produces non-standard output on Windows (so called UNC format).

Yes

### DIFF
--- a/language/move-analyzer/src/symbols.rs
+++ b/language/move-analyzer/src/symbols.rs
@@ -334,9 +334,7 @@ impl SymbolicatorRunner {
         let runner = SymbolicatorRunner { mtx_cvar };
 
         thread::Builder::new()
-            // building Move code requires a larger stack size on Windows (16MB has been chosen
-            // rather arbitrarily)
-            .stack_size(16 * 1024 * 1024)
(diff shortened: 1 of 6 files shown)


language/move-analyzer/src/symbols.rs
```diff
@@ -333,7 +333,9 @@ impl SymbolicatorRunner {
         let thread_mtx_cvar = mtx_cvar.clone();
         let runner = SymbolicatorRunner { mtx_cvar };
 
-        thread::spawn(move || {
+        thread::Builder::new()
+            .stack_size(16 * 1024 * 1024) // building Move code requires a larger stack size on Windows
```
tnowacki, line 337: bit magical, an reasoning behind the size?

tnowacki (approved):
Looks good, just maybe some doc about how you came to that stack size

language/move-analyzer/src/symbols.rs
awelc: It has been chosen somewhat arbitrarily, based on experience with using the compiler in a Move adapter.

Issue #166, opened 2022-05-26
[Feature Request] move fmt


Manahip:
# 🚀 Feature Request

Move formatter. `move fmt`, just like how `cargo fmt` formats rust code.

## Motivation

Useful in enforcing consistent coding style.

yubing744:
@awelc Does move-analyzer plan to support the textDocument/formatting PSL protocol, and [starcoin-ide](https://github.com/starcoinorg/starcoin-ide) also wants to integrate this function?

Manahip:
I've been meddling with Move ASTs in my Move-to-TypeScript transpiler. I can probably do a formatter myself. Do you guys know where I could get a grant to speed things up? We're not raising yet so any bit of cash would help.

yubing744:
@Manahip You can participate in https://github.com/starcoinorg/startrek organized by starcoin, there will be bonuses after graduation.

yubing744:
@Manahip I found a project that has implemented move-fmt, repository: https://github.com/victoryang00/move-fmt

jolestar:
@LemonHX Do you know the author?

wrwg:
The move-fmt project is for Move IR, an older (and simpler) instance of Move.

yubing744:
> The move-fmt project is for Move IR, an older (and simpler) instance of Move.

The move-fmt project uses pest parsing to get the AST, and then does the formatting. If the move grammar he configured is older, we can update the grammar file https://github.com/victoryang00/move-fmt/blob/master/src/grammar.pest

zsluedem:
The move fmt could use the internal move parser to get the ast and do the formatting
https://github.com/move-language/move/blob/main/language/move-compiler/src/parser/mod.rs#L29-L132 .

In that case, move fmt could be consistent with the compiler even if there are some updates.

yubing744:
> The move fmt could use the internal move parser to get the ast and do the formatting https://github.com/move-language/move/blob/main/language/move-compiler/src/parser/mod.rs#L29-L132 .
> 
> In that case, move fmt could be consistent with the compiler even if there are some updates.

this is a good idea.

zsluedem:
> this is a good idea.

I want to give this idea a try and get myself into it now. 
After we get the ast, the move fmt could do something similar like rustfmt 
https://github.com/rust-lang/rustfmt/blob/2403f827bf1e427a04ce45c88a64cbebe91041d7/src/rewrite.rs#L16-L19 .
[...]

PR #294, opened 2022-07-22
[Move-compiler] introduce additional stepped compiler Translate

zsluedem:
## Motivation
The goal of the pr is to introduce an additional stepped compiler between parsing and expansion.

Compiler is doing from `code string -> ast -> expansion ast -> .....` to `code string -> cst -> ast -> expansion ast`.

A new cst structure is introduced which is lossless, meaning that they represent Move programs in sufficient detail to go from text to trees and vice-versa.

Incremental step on the way to https://github.com/move-language/move/issues/166

Yes.

## Test Plan

Tests are added.


language/move-compiler/src/command_line/compiler.rs
```diff
@@ -215,12 +218,14 @@ impl<'a> Compiler<'a> {
...
         let (source_text, pprog_and_comments_res) =
-            parse_program(&mut compilation_env, maps, targets, deps)?;
-        let res: Result<_, Diagnostics> = pprog_and_comments_res.and_then(|(pprog, comments)| {
-            SteppedCompiler::new_at_parser(compilation_env, pre_compiled_lib, pprog)
-                .run::<TARGET>()
-                .map(|compiler| (comments, compiler))
-        });
+            parse_program(&mut compilation_env, targets, deps)?;
+        let res: Result<_, Diagnostics> = pprog_and_comments_res
+            .and_then(|pprog| translate_program(&mut compilation_env, maps, pprog))
```
tnowacki, line 223: This would need to become a new step in the stepped compiler.

language/move-compiler/src/parser/parsed_tree/translate/translate_address.rs
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) The Diem Core Contributors
```
tnowacki, line 1: I'm not a fan of these separate modules/files for each section of the ast. I know the files get very large, but having everything in one place makes understanding the entire flow easier in my opinion. (plus it is then more consistent with the rest of the compiler)

tnowacki (changes requested):
Thanks for starting this! It will be a lot of work, but it will be worth it! And this looks like a good first step.  

I will try to leave some more specific comments in individual files. But some comments:

- It's parse tree or concrete syntax tree, not parse*d* tree :) 
- The ast/translate naming combo is off here. We do `<pass>::ast` and `<pass>::translate(<previous_pass>::ast::Program) -> <pass>::ast::Program`. In other words, the `translate` for a pass is the one that makes that AST. But I think things are a bit funky for the parser as it is doing a lot. I could see a few possible setups:
  - `parser/cst.rs` and `parser/ast.rs` with `parser/translate.rs` that converts the `cst` to an `ast`
  - `parser/parse_tree.rs` and `parser/ast.rs` with `parser/convert_tree.rs` that converts the `parse_tree` to an `ast`
  - `parser/cst.rs` and `<cool_new_verb>/ast.rs` with `<cool_new_verb>/translate.rs` that converts the `cst` to the `ast`
- I want to highlight again that I think all these individual sections should just be in one file. I find it very hard to follow otherwise.

language/move-compiler/src/parser/parsed_tree/ast.rs
```diff
@@ -0,0 +1,625 @@
...
+pub type NameAccessChain = SpannedWithComment<Vec<Name>>;
+pub type LeadingNameAccess = Name;
+pub type SpecApplyFragment = ParsedToken;
+
+pub type QuantKind = SpannedWithComment<QuantKind_>;
+pub type BinOp = SpannedWithComment<BinOp_>;
+pub type UnaryOp = SpannedWithComment<UnaryOp_>;
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum ParsedTree {
```
tnowacki, line 127: This is going to need to be generalized a lot more (as briefly discussed in discord). The goal here is to fully capture the syntax that is in the file, with all whitespace and comment information. It can also be helpful to be able to represent invalid trees (or at least partially invalid trees). As such, parse trees typically do not use distinct types for different syntactic groups

language/move-compiler/src/parser/parsed_tree/translate/translate_address.rs
zsluedem: OK. Copy that.

zsluedem: I was thinking I would migrate these files into one when the work is done(tests passed) because I am more comfortable with seperate files while developing. I planned to ask after the work is done. But nvm, I would definitely keep this consistent with other packages.

language/move-compiler/src/parser/parsed_tree/ast.rs
zsluedem: But I was having some parsing conflicts with 

```move
spec <Identifier>{
...
}
```
 which could be parsed into `Module` representation or `SpecBlock` representation.
 
And I want to distinguish `Module` and `SpecBlock` because the items inside Module and SpecBlock are different and I want to distinguish them. So I fall back into different syntactic groups.

Hm.... Maybe I should just take `Module` and `SpecBlock` as the same enum and distinguish them later in the translation steps. 

Do you have any preferences on the point above?

zsluedem:
@tnowacki The pr is far from review ready right now. Thank you very much for your time. It could really help keep me on the right track.

language/move-compiler/src/parser/parsed_tree/ast.rs
tnowacki: You will probably have to parse it more generally, and then separate it into distinct AST nodes in this separate translation step I described above. 

In short, expect this parse tree to be _very_ general. It will more or less correspond to the structure of the current parser, but not any of the Rust code inside that structure. All the more complex/contextual analysis will need to be done in this pass that converts the parse tree to the abstract syntax tree

language/documentation/examples/diem-framework/move-packages/DPN/sources/AccountAdministrationScripts.move
```diff
@@ -356,7 +356,7 @@ module DiemFramework::AccountAdministrationScripts {
 
         /// **Access Control:**
         /// Only the Diem Root account can process the admin scripts [[H9]][PERMISSION].
-        requires Roles::has_diem_root_role(dr_account); /// This is ensured by DiemAccount::writeset_prologue.
+        requires Roles::has_diem_root_role(dr_account); // This is ensured by DiemAccount::writeset_prologue.
```
zsluedem, line 359: This was a `DocComment` which doesn't match any `DocComment` described language item. It didn't throw warning of `InvalidDocComment`. I made some of the same doc comments changes below.

language/documentation/examples/diem-framework/move-packages/DPN/sources/DiemAccount.move
```diff
@@ -2335,7 +2335,7 @@ module DiemFramework::DiemAccount {
     spec module {} // switch documentation context back to module level
 
     /// # Access Control
-
+    ///
```
zsluedem, line 2338: The doc comment didn't require the consistency of the comments. I made some changes which the doc comments need to be consistent to be recognized. I got this idea because sometimes people would do comments like 
```
/// ------------------------------ line split mark -----------------------------------

/// doc comment content
fun a(){}
```

In this situation, the `line split mark` would take as doc comment with `doc comment content` which doesn't make too much sense to me.

I made the changes which the doc comment is only matched right before the language item.

Please let me know if you think this is not ok.

zsluedem: A lot of comment changes in `diem-framework` are made because of it.

language/evm/stdlib/sources/Evm.move
```diff
@@ -1,5 +1,5 @@
 /// Module which provides access to EVM functionality, including information about the executing transaction.
-////
+///
```
zsluedem, line 2: I don't know why but removing this could help this failing test case pass https://github.com/move-language/move/runs/8219288895?check_suite_focus=true . I don't understand how comments affect the results. It would be great someone could give me some lights.

language/move-compiler/src/command_line/compiler.rs
```diff
@@ -214,9 +217,11 @@ impl<'a> Compiler<'a> {
...
         )?;
         let mut compilation_env = CompilationEnv::new(flags);
-        let (source_text, pprog_and_comments_res) =
-            parse_program(&mut compilation_env, maps, targets, deps)?;
-        let res: Result<_, Diagnostics> = pprog_and_comments_res.and_then(|(pprog, comments)| {
+        let (source_text, pprog_res) = parse_program(&mut compilation_env, targets, deps, maps)?;
+        let mut comments = CommentMap::new();
+
+        let res: Result<_, Diagnostics> = pprog_res.and_then(|pprog| {
+            find_comments(&pprog, &mut compilation_env, &mut comments)?;
```
zsluedem, line 224: Maybe this is a `TODO`. It would be greate to move this only to `doc-gen` command.

language/move-compiler/tests/move_check/flavors/async/derive_errors.exp
```diff
@@ -69,12 +69,6 @@ error[E13001]: attribute derivation failed
 22 │     #[message] fun f7(_x: &A, _a1: u64, _a2: u64, _a3: u64, _a4: u64, _a5: u64, _a6: u64,
    │       ^^^^^^^ too many arguments for message handler (current max is 8)
 
-error[E03003]: unbound module member
-   ┌─ tests/move_check/flavors/async/derive_errors.move:22:7
-   │
-22 │     #[message] fun f7(_x: &A, _a1: u64, _a2: u64, _a3: u64, _a4: u64, _a5: u64, _a6: u64,
-   │       ^^^^^^^ Invalid module access. Unbound function 'send__9' in module '(Async=0x20)::Runtime'
-
```
zsluedem, line 77: This error message is gone because I moved the checking into any layer which is `Expansion`. The derived check failed first but the unbound check haven't started.

language/move-compiler/tests/move_check/parser/spec_parsing_inside_fun.exp
```diff
@@ -1,54 +1,9 @@
...
-   │
-34 │         &mut spec {};
-   │         ^^^^^^^^^^^^
-   │         │    │
-   │         │    Expected a single non-reference type, but found: '()'
-   │         Invalid borrow
+   │                 ^
+   │                 │
+   │                 Unexpected '+'
+   │                 Expected an expression term
```
zsluedem, line 8: I am actually a little bit confused that why a `Spec` could be a member of `Expression` https://github.com/move-language/move/blob/164bf33c8b396b9a29305d9c1b9a0d3b0c5b8cef/language/move-compiler/src/parser/ast.rs#L643-L644. Maybe I missed something that is important. Please let me know.

I chose to let `cst` behaved similarly with `ast` which `Exp` includes `Spec` to make the test https://github.com/move-language/move/blob/164bf33c8b396b9a29305d9c1b9a0d3b0c5b8cef/language/move-compiler/tests/move_check/typing/spec_block_fail.move#L1-L7 pass.

I'd like to let this fail earlier in the parsing which `spec` doesn't work like `Exp` . 

Please let me know your ideas on it.

language/move-compiler/src/parser/parsed_tree/ast.rs
zsluedem: @tnowacki Please take a look whether the pr is on the right track or not~!

zsluedem:
Got other plans

yuyang-ok:
@zsluedem What's you new plans??? :-)

zsluedem:
@yuyang-ok Well. I got a crazy idea to rewrite the parser in another repo and probably rewrite the vs code plugin.

The parser would parse the code into a red-green tree ( lossless syntax trees).
The parser should also do the incremental reparsing  easily which the vscode plugin would benefit a lot.

PR #296, opened 2022-07-22
[vm] Allow module publishing skipping compatability checks.

wrwg:
By default module publishing checks for downwards compatibility of a module when it is re-published. This PR adds a new entry point which allows to relax this check if desired.

### DIFF
--- a/language/move-vm/runtime/src/runtime.rs
+++ b/language/move-vm/runtime/src/runtime.rs
@@ -75,6 +75,7 @@ impl VMRuntime {
         sender: AccountAddress,
         data_store: &mut impl DataStore,
         _gas_status: &mut GasStatus,
+        compat_check: bool,
     ) -> VMResult<()> {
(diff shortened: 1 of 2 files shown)


tnowacki (approved):
LGTM

tnowacki:
We should probably consider some additional flexibility here too, around all sorts of the verification steps.

Currently, the loader re-verifies the modules every time they are loaded. I think it would be fine to have some config for a less paranoid loader :) There are possibly some other configs we might want to include, maybe this one would be in there, who knows

wrwg:
> We should probably consider some additional flexibility here too, around all sorts of the verification steps.
> 
> Currently, the loader re-verifies the modules every time they are loaded. I think it would be fine to have some config for a less paranoid loader :) There are possibly some other configs we might want to include, maybe this one would be in there, who knows

Agreed, we could come up with an enum parameter instead to specify level of checking.

PR #297, opened 2022-07-22
[move-prover] Add missing location line in error trace

jyao15:
## Motivation

The Move prover does not print the correct line number in some error traces. Consider the example below.
```
public fun test_loop(x: u64) {
    if (x != 0) {
        x = 1;
    }
    else {
        x = 2;
    };
    spec {
        assert x == 1;
    };
}
```
The following error trace is given.
```
error: unknown assertion failed
  ┌─ ./sources/loop.move:9:9
  │
9 │         assert x == 1;
  │         ^^^^^^^^^^^^^^
  │
  =     at ./sources/loop.move:1: test_loop
  =         x = 0
  =     at ./sources/loop.move:2: test_loop
  =         x = 2
  =     at ./sources/loop.move:8: test_loop
  =     at ./sources/loop.move:9: test_loop
```
However, the assignment `x = 2` occurs at Line 6 rather than Line 2. This pull request will insert `at ./sources/loop.move:6: test_loop` above `x = 2`.

Yes

## Test Plan

cargo test --release

### DIFF
--- a/language/move-prover/boogie-backend/src/bytecode_translator.rs
+++ b/language/move-prover/boogie-backend/src/bytecode_translator.rs
@@ -748,6 +748,11 @@ impl<'env> FunctionTranslator<'env> {
             *last_tracked_loc = None;
         }
         self.track_loc(last_tracked_loc, &loc);
+        if matches!(bytecode, Label(_, _)) {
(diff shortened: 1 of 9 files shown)


wrwg (approved):
Great!

PR #298, opened 2022-07-23
[move-prover] fix hyper-edge constructor in borrow analysis

meng-xu-cs:
This commit fixes issue #270 where the edge ordering are reversed in a
hyper-edge when that hyper-edge is inlined to construct a borrow graph.

In Move Prover, hyper-edges are used to summarize the borrow-relation
between function arguments and function return values:

In the example in #270, given

```move
fun inner(has_vector: &mut HasVector): &mut HasAnotherVector {
    vector::borrow_mut(&mut has_vector.v, 7)
}
```

The `inner` function will yield a hyper-edge:
```
return_ref --borrows_from--> { arg: has_vector, edge: Hyper(.v/[]) }
```

Now, this hyper-edge will be further used to derive the borrow-relation
for the `mid` function:

```
fun mid(has_vector: &mut HasVector): &mut HasAnotherVector {
    inner(has_vector)
}
```

The expected summary for `mid` is
```
return_ref --borrows_from--> { arg: has_vector, edge: Hyper(.v/[]) }
```

However, due to the bug, the summary becomes:
```
return_ref --borrows_from--> { arg: has_vector, edge: Hyper([]/.v) }
                                                            ^^^^^
                                                            // reversed
```
which causes the crash found in #270.

This commit fixes this bug. In summary, the fix is: when constructing
hyper-edges, reverse the edges first before flattening all the edges.

## Motivation

As above

Yes

## Test Plan

Unit tests

### DIFF
--- a/language/move-prover/bytecode/src/borrow_analysis.rs
+++ b/language/move-prover/bytecode/src/borrow_analysis.rs
@@ -260,7 +260,7 @@ impl BorrowInfo {
     ) {
         for (dest, edge) in outgoing.iter() {
             let mut path = prefix.to_owned();
-            path.extend(edge.flatten().into_iter().cloned());
+            path.push(edge.clone());
             if let Some(succs) = ret_info.borrows_from.get(dest) {
                 self.construct_hyper_edges(leaf, ret_info, path, succs);
             } else {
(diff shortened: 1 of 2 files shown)


language/move-prover/bytecode/src/borrow_analysis.rs
```diff
@@ -260,7 +260,7 @@ impl BorrowInfo {
     ) {
         for (dest, edge) in outgoing.iter() {
             let mut path = prefix.to_owned();
-            path.extend(edge.flatten().into_iter().cloned());
```
wrwg, line 263: This indeed looks wrong

wrwg (approved):

meng-xu-cs:
@awelc very good question. I think the edges in `borrowed_by` and `borrows_from` are kept the same (including their orderings) on purpose for simplicity of coding. While doing the abstract interpretation for borrow analysis, only `borrowed_by` gets updated, `borrows_from` is only updated when the abstract interpretation finishes (in the `BorrowInfo::consolidate()` function to be specific).

So if we would like to reverse the ordering of the edges, we probably want to do it in the `consolidate()` function as well. But I am not entirely sure whether we want to do that.

PR #300, opened 2022-07-24
Use deserialized constant pool for LdConst

oxade:
## Motivation
`LdConst` currently deserializes the BCS encoded constants on every invocation. This is unnecessary and also makes it considerably slower than `Ld*` of same data size.
This PR uses a pool of pre-deserialized constants to avoid this.

Yes

## Test Plan

Standard tests
Test failure due to `ValueImpl` not impl `Send`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3665,7 +3665,6 @@ dependencies = [
  "move-vm-runtime",
  "move-vm-types",
  "once_cell",
- "serde 1.0.143",
  "sha3 0.9.1",
  "smallvec",
  "tempfile",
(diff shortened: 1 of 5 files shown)


language/move-vm/runtime/src/loader.rs
```diff
@@ -1636,9 +1654,6 @@ impl Module {
 // (rather then "compiled") to make available data needed for execution
 // #[derive(Debug)]
 struct Script {
-    // primitive pools
-    script: CompiledScript,
```
tnowacki, line 1640: Was this just unused?

tnowacki (commented):
Overall, things look great! thanks for doing this!

language/move-vm/runtime/src/loader.rs
oxade: It was only used to fetching constants. Now that I have a replacement, it's no longer needed.

vgao1996:
@oxade are you sure this was committed correctly? It seems like there are no new commits associated with this PR.

oxade:
> @oxade are you sure this was committed correctly? It seems like there are no new commits associated with this PR.

@vgao1996 No there was no commit. I think this is a Github bug.

Issue #299, opened 2022-07-23
[Bug] [move-prover] unconditional write-back for conditional borrows


meng-xu-cs:
# 🐛 Bug

The following Move function should be verifiable, but the prover will complain about post-condition does not hold

```move
module 0x42::Test {
    struct T has key {
        x: u64,
    }

    fun diff_address(cond: bool, a1: address, a2: address) acquires T {
        let x = if (cond) {
            let t1 = borrow_global_mut<T>(a1);
            &mut t1.x
        } else {
            let t2 = borrow_global_mut<T>(a2);
            &mut t2.x
        };
        *x = 0;
    }

    spec diff_address {
        aborts_if cond && !exists<T>(a1);
        aborts_if !cond && !exists<T>(a2);
        ensures if (cond) global<T>(a1).x == 0 else global<T>(a2).x == 0;
    }
}
```

## To reproduce

Suppose the above Move code is saved in a file called `test.move`

**Code snippet to reproduce**
```bash
cd language/move-prover
cargo run -- test.move
```

**Error message**
```
error: post-condition does not hold
   ┌─ test.move:19:9
   │
19 │         ensures if (cond) global<T>(a1).x == 0 else global<T>(a2).x == 0;
   │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   │
   =     at test.move:6: diff_address
   =         cond = true
   =         a1 = 0x29
   =         a2 = 0x4823
   =     at test.move:7: diff_address
   =         t1 = &Test.T{x = 6334}
   =     at test.move:9: diff_address
   =     at test.move:7: diff_address
   =         x = &6334
   =     at test.move:14: diff_address
   =     at test.move:15: diff_address
   =     at test.move:17: diff_address (spec)
   =     at test.move:18: diff_address (spec)
   =     at test.move:19: diff_address (spec)

exiting with verification errors
```

## Expected Behavior

Verification should pass, as the code is in confirmation with the spec.

## Suspected cause

When borrow is conditional, the write-back should be conditional as well.

Following is a detailed elaboration:

```bash
cargo run -- test.move --dump-bytecode --keep
cat test.move_7_memory_instr.bytecode
```

```
============ after processor `memory_instr` ================

[variant baseline]
fun Test::diff_address($t0|cond: bool, $t1|a1: address, $t2|a2: address) {
     var $t3|tmp#$3: &mut u64
[...]

PR #302, opened 2022-07-25
[move-prover] write-back following the whole ancestor tree

meng-xu-cs:
The current memory model in Move Prover has a flaw in writing back
dying references, as identified in issue https://github.com/move-language/move/issues/299.

The gist of the flaw is this:
we can only write-back a reference on code paths where it is guaranteed
to be defined. Otherwise in the prover's point of view, the write-back
may cause arbitrary value to be written back to the parent `Mutation`
value and cause unexpected verification failures.

As a side note, this is also visible in the concrete executor of the
spec + move (i.e., the stackless VM).

The solution, as implemented in this commit, is to ravamp the way
write-backs are done.

In summary, the solution is to write-back the dying references according
to DFS paths instead of BFS layers.

Taking the example in https://github.com/move-language/move/issues/299:

The borrow graph looks like this:

```
     --borrows_from--> t7 --borrows_from--> Global(t1)
t3 --|
     --borrows_from--> t8 --borrows_from--> Global(t2)
```

And the old approach is to write-back according to BFS layers,
i.e., t3, then t7 and t8, which yields the following psudo-code.

```
if IsParent[t7](t3) {
    WriteBack[t7](t3);
}

if IsParent[t8](t3) {
    WriteBack[t8](t3);
}

// both t7 and t8 are not always defined when code reaches here
WriteBack[Global](t7);
WriteBack[Global](t8);
```

The correct behavior should be DFS paths, i.e.:

```
if IsParent[t7](t3) {
    WriteBack[t7](t3);
    // t7 is guaranteed to be defined when code reaches here
    WriteBack[Global](t7);
}

if IsParent[t8](t3) {
    WriteBack[t8](t3);
    // similarly, t8 is guaranteed to be defined in this block
    WriteBack[Global](t8);
}
```

## Motivation

To fix issue #299 

Yes

## Test Plan

A new move-prover functional test is added

Inconsistency tests manually executed on DPN code:
```bash
cd language/documentation/examples/diem-framework/move-packages/core
cargo run -p df-cli -- prove -- --check-inconsistency

cd language/documentation/examples/diem-framework/move-packages/DPN
cargo run -p df-cli -- prove -- --check-inconsistency

cd language/documentation/examples/diem-framework/move-packages/experimental
cargo run -p df-cli -- prove -- --check-inconsistency
```

### DIFF
--- a/language/move-prover/bytecode/src/borrow_analysis.rs
+++ b/language/move-prover/bytecode/src/borrow_analysis.rs
@@ -21,11 +21,7 @@ use move_model::{
     ty::Type,
     well_known::{TABLE_BORROW_MUT, VECTOR_BORROW_MUT},
 };
-use std::{
-    borrow::BorrowMut,
-    collections::{BTreeMap, BTreeSet},
-    fmt,
-};
+use std::{borrow::BorrowMut, collections::BTreeMap, fmt};
 
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Default)]
 pub struct BorrowInfo {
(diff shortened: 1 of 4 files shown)


wrwg (commented):
Very cool. Does this fix the same issue you identified last summer? 

Also can we confirm that DPN verification tests are run by CI.

meng-xu-cs:
> Does this fix the same issue you identified last summer?
Yes, this fixes the exact same issue as the [stale PR](https://github.com/diem/diem/pull/8430) does, but in a much cleaner way.

meng-xu-cs:
> Also can we confirm that DPN verification tests are run by CI.

From this setup, it seems so: https://github.com/move-language/move/blob/main/.github/workflows/ci-pre-land.yml#L154

I also manually ran `prove_all.sh` and it does not complain anything. So, seems to be working fine. I'll request for review after I resolve the merge conflicts.

wrwg (approved):
Very happy to see this fixed. I also verified that indeed DPN verification tests are run. Could you also manually run the inconsistency tests before merging? See also #304, they seem to be broken currently.

meng-xu-cs:
Inconsistency check manually executed and all passed. Updated the test plan section in the PR desc.

Note that the way of running the prover inconsistency check is not correct in #304, also left a comment there.

PR #309, opened 2022-07-27
[move-prover] Bump up the version number of the Prover dependencies

junkil-park:
- Updated the Z3 version to the latest one.
- Updated the Boogie version to latest one that is compatible with Prover.
  - Incompatibility issue filed in the Boogie repo: https://github.com/boogie-org/boogie/issues/609
- Fixed a few bugs in `boogie_wrapper`.

## Motivation

To bump up the version number of Prover's deps.

yes

## Test Plan
cargo test -p move-prover

### DIFF
--- a/language/move-prover/boogie-backend/src/boogie_wrapper.rs
+++ b/language/move-prover/boogie-backend/src/boogie_wrapper.rs
@@ -1682,7 +1682,7 @@ impl<'s> ModelParser<'s> {
 
     fn parse_key(&mut self) -> Result<ModelValue, ModelParseError> {
         let mut comps = vec![];
-        while !self.looking_at("->") {
+        while !self.looking_at("->") && self.at < self.input.len() {
             let value = self.parse_value()?;
             comps.push(value);
         }
--- a/scripts/dev_setup.sh
+++ b/scripts/dev_setup.sh
@@ -15,10 +15,10 @@
 # fast fail.
 set -eo pipefail
 
-Z3_VERSION=4.8.13
+Z3_VERSION=4.10.2
 CVC5_VERSION=0.0.3
-DOTNET_VERSION=5.0
-BOOGIE_VERSION=2.9.6
+DOTNET_VERSION=6.0
+BOOGIE_VERSION=2.13.4
 SOLC_VERSION="v0.8.11+commit.d7f03943"
 
 SCRIPT_PATH="$(cd "$(dirname "$0")" >/dev/null 2>&1 && pwd)"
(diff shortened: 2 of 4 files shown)


language/move-prover/boogie-backend/src/boogie_wrapper.rs
```diff
@@ -1682,7 +1682,7 @@ impl<'s> ModelParser<'s> {
 
     fn parse_key(&mut self) -> Result<ModelValue, ModelParseError> {
         let mut comps = vec![];
-        while !self.looking_at("->") {
+        while !self.looking_at("->") && self.at < self.input.len() {
```
meng-xu-cs, line 1685: I am running Ubuntu 20.04.4 LTS and I realize that my prover tests pass without this line of change...

scripts/dev_setup.sh
```diff
@@ -15,10 +15,10 @@
 # fast fail.
 set -eo pipefail
 
-Z3_VERSION=4.8.13
+Z3_VERSION=4.10.1
```
meng-xu-cs, line 18: Also, Z3 [4.10.2](https://github.com/Z3Prover/z3/releases/tag/z3-4.10.2) is released as well.

language/move-prover/boogie-backend/src/boogie_wrapper.rs
meng-xu-cs: So I still don't quite understand what this line is doing...

scripts/dev_setup.sh
junkil-park: Done.

language/move-prover/boogie-backend/src/boogie_wrapper.rs
junkil-park: Sud0u53r at our Discord channel points out that this is a bug which causes Prover fall into an infinite loop when `boogie_wrapper` tries to parse an incomplete map expression with no value part (like `var ->`). This is a clear bug that needs to be fixed.

junkil-park: > I am running Ubuntu 20.04.4 LTS and I realize that my prover tests pass without this line of change...
If you upgrade Boogie version to 2.14.0, Boogie will start producing incomplete expressions like (`$abort_code ->`). If you run Prover again without this fix, Prover will not terminate due to the infinite loop bug. If you run Prover with this fix, Prover will terminate, but complain about such incomplete expressions.

I am trying to be confirmed whether such incomplete expressions are the expected behavior of Boogie or a bug. In the meantime, we can safely upgrade and use the Boogie version of 2.13.4.

meng-xu-cs: I see. Fair enough

meng-xu-cs (approved):

PR #311, opened 2022-07-27
[move-prover] Add evaluation logic for choice operators

ckaiwxm:
## Motivation

- Complete the evaluator for the `choose` and `choose min` operators. All the values in the range that satisfy the predicate are extracted and the value in the last position of the vector is returned deterministically for the `choose` operator.
- Update the evaluator for assume expression to return `EvalResult` to distinguish `None` results from choice operators and other assertions. Move `skip_specs` from the function context to the local state to enable the `let` binding. If the choice operators return `None`, skip the following specs to log the error message.
- When no value satisfies the predicate for the choice operators, log the error message and abort the execution.

Yes.

## Test Plan

### DIFF
--- a/language/move-prover/interpreter-testsuite/tests/concrete_check/choice.exp
+++ b/language/move-prover/interpreter-testsuite/tests/concrete_check/choice.exp
@@ -1,0 +1,67 @@
+Running Move unit tests
+[ FAIL    ] 0x2::A::simple_number_min_range_failure
+[ FAIL    ] 0x2::A::simple_number_range_failure
+[ FAIL    ] 0x2::A::vector_choose_min_unsatisfied_predicate
+[ PASS    ] 0x2::A::vector_choose_success
+[ FAIL    ] 0x2::A::vector_choose_unsatisfied_predicate
+
+Test failures:
+
+Failures in 0x2::A:
+
+┌── simple_number_min_range_failure ──────
+│ error: failed to evaluate expression: enumeration of a non-address type domain is not supported
+│    ┌─ tests/concrete_check/choice.move:56:42
+│    │
+│ 56 │         ensures result <= (choose min x: u64 where x >= 4);
+│    │                                          ^^^
+│
+│ error: failed to evaluate expression: unexpected error code
+│    ┌─ tests/concrete_check/choice.move:56:17
+│    │
+│ 56 │         ensures result <= (choose min x: u64 where x >= 4);
+│    │                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+│
+│
+└──────────────────
+
+
+┌── simple_number_range_failure ──────
+│ error: failed to evaluate expression: enumeration of a non-address type domain is not supported
+│    ┌─ tests/concrete_check/choice.move:49:38
+│    │
+│ 49 │         ensures result <= (choose x: u64 where x >= 4);
+│    │                                      ^^^
+│
+│ error: failed to evaluate expression: unexpected error code
+│    ┌─ tests/concrete_check/choice.move:49:17
+│    │
+│ 49 │         ensures result <= (choose x: u64 where x >= 4);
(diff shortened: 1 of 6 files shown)


language/move-prover/interpreter-testsuite/tests/concrete_check/choice.exp
```diff
@@ -0,0 +1,67 @@
...
+Failures in 0x2::A:
+
+┌── simple_number_min_range_failure ──────
+│ error: failed to evaluate expression: enumeration of a non-address type domain is not supported
+│    ┌─ tests/concrete_check/choice.move:62:42
+│    │
+│ 62 │         ensures result <= (choose min x: u64 where x >= 4);
+│    │                                          ^^^
+│
+│ error: failed to evaluate expression: unexpected error code
```
meng-xu-cs, line 19: Is there a way to remove this additional error message?

language/move-prover/interpreter-testsuite/tests/concrete_check/choice.move
```diff
@@ -0,0 +1,64 @@
...
+        v
+    }
+
+    spec vector_choose_success {
+        let post choice = choose i in 0..len(result) where result[i] == 1;
+        ensures choice == 0 || choice == 2;
+        ensures (choose min i in 0..len(result) where result[i] == 1) == 0;
+    }
+
+    #[test, expected_failure]
```
meng-xu-cs, line 25: `[expected_failure]` is intended for the Move code execution without the spec checking. In this case, the code itself executes successfully without aborts, so we should remove this flag.

language/move-prover/interpreter-testsuite/tests/concrete_check/choice.move
```diff
@@ -0,0 +1,64 @@
...
+        vector::push_back(&mut v, 2);
+        vector::push_back(&mut v, 1);
+        v
+    }
+
+    spec vector_choose_unsatisfied_predicate {
+        let post choice = choose i in 0..len(result) where result[i] == 3;
+    }
+
+    #[test, expected_failure]
```
meng-xu-cs, line 38: Same applies for this `[expected_failure]` and the rest in this file.

meng-xu-cs (commented):
Thanks for the PR! It looks quite good to me! See if you can get a fix on the two minor comments here and we are good to go.

language/move-prover/interpreter-testsuite/tests/concrete_check/choice.move
ckaiwxm: I see, I have removed the label for this file and for the `quantifier.move` file that I created earlier. Thank you for the explanation!

language/move-prover/interpreter-testsuite/tests/concrete_check/choice.exp
ckaiwxm: I'm thinking of defining a function similar to `eval_failure_code` to distinguish between domain type failure and other evaluation failures. Then, when we reach the function `check_assert`, we can check the error check to see if it is a domain failure, and stop logging `failed to evaluate expression: unexpected error code` if that is the case?

meng-xu-cs: Yes, that is the intended solution. But I just realized that doing this change is outside the scope of this PR (because some of the arithmetic errors are also reported as `unexpected error code`), so let's leave the code as it is now.

meng-xu-cs:
Thanks for the revision, the code looks good to me (you need to update the `exp` file though). I'll help with the final stamping after you are done with the update.

language/move-prover/interpreter-testsuite/tests/concrete_check/choice.exp
ckaiwxm: Yep, I'll make this change in another PR, in that case maybe we can have more informative error message for arithmetic errors too

ckaiwxm:
> Thanks for the revision, the code looks good to me (you need to update the `exp` file though). I'll help with the final stamping after you are done with the update.

meng-xu-cs (approved):

PR #317, opened 2022-07-29
Use standard traits for FromStr

gregnazario:
## Motivation

FromStr helps a lot of other tools properly parse types without having to go find the functions related.  Since these already exist, it should make it easier.

Also, allow AccountAddress to load with 0x in front.  I'm not going to try to deal with the output path in this PR.

Yes

## Test Plan

The current unit tests


wrwg (approved):

wrwg:
Please consider you can't update aptos to this change yet because there are incompatibilities in the core apis which meed to be resolved first. @vgao1996

PR #227, opened 2022-06-27
[Movey] Add command to upload metadata to Movey

ea-open-source:
## Motivation

Add `move movey-upload` command to upload move package to [Movey](https://www.movey.net/), modify `move build` command to call Movey API to increase download count.

### How to use it
Before you use the new command to upload metadata to Movey, please make sure:
* your package must be on Github
* the remote must be named “origin”
* file $HOME/.move/movey_credential.toml must exist, and the content should be like(you can get the token on [Movey](https://www.movey.net/)):
```
[registry]
token = "example-token"
```
Note: you could use the `move login` command from this [PR](https://github.com/move-language/move/pull/226) to save your Movey's api token using the CLI.

Step 1: Navigate to the folder containing your package.
Step 2: Run the following command:
`move movey-upload`

### Have you read the [Contributing Guidelines on pull requests](https://github.com/diem/diem/blob/main/CONTRIBUTING.md#pull-requests)?

Yes

## Test Plan

cargo test --package move-cli --test cli_tests upload_package
cargo test --package move-cli --lib get_api_token

PR #318, opened 2022-07-29
[Movey] Call Movey API to increase download count

ea-open-source:
## Motivation
According to [Sam](https://github.com/move-language/move/pull/227#discussion_r922850757), we moved the call to  Movey into this pull request.
### How to use it
As usual any build action that involves cloning from github will also calls to Movey to increase download count metre.

Note: you could use the flag`--skip-movey-call` to skip this call, although it has negligible impact on build performance.
Example: `move movey-upload --skip-movey-call`
Yes


awelc (commented):
I have to say I am still confused what this PR does... I though this addition is to avoid increasing GitHub's download count when uploading the repo content to Movey (so that the counter value is not artificially inflated by multiple uploads). Looking at the code, though, it seems like we are actually incrementing the counter (which would otherwise not be incremented as part of the upload?), so I am not even sure if I got the notion of the counter right... Could you please provide some additional explanation?

awelc:
Adding @tnowacki as a reviewer here as this PR involves a change to package resolver and I am not sure if spawning a thread and calling to Movey on (seemingly) each dependency check (for already downloaded dependencies) during resolution graph building is the right strategy here.

ea-open-source:
> Adding @tnowacki as a reviewer here as this PR involves a change to package resolver and I am not sure if spawning a thread and calling to Movey on (seemingly) each dependency check (for already downloaded dependencies) during resolution graph building is the right strategy here.

I think you misunderstood the logic, calling to the Movey only happens when `download_and_update_if_remote` is called and this only happens after the dependency tree is resolved. So in essence, Movey only increases download counts when users download a package from remote.

ea-open-source:
@awelc @tnowacki would you guys help us review this PR?

awelc (changes requested):
Thank you for the additional explanation! It looks like the call to Movey indeed only happens right where `download_and_update_if_remote` is called, which makes me  wonder why you did not include it as part of `download_and_update_if_remote` function's definition. You even have the right checks there already:
```
        if let Some(git_info) = &dep.git_info {
            if !git_info.download_to.exists() {
```
This would make the logic here more explicit right off the bat and I suggest you do it unless there is a good reason not to.

I am still confused by what you are trying to accomplish here. You call into Movey when resolving dependencies during a build process and a developer may be building a package that has nothing to do with Movey. I don't understand why you want to call to Movey in such cases, and even more importantly, even if this was the case, this should be presented as opt-in to the user (they choose to increase the Movey download count) rather than opt-out (they have to choose not to increase the count). 

Additionally, the `--skip-movey` flag is presented in the description of this PR as an addition to the `move movey-upload`, but it's part of build config and (I think) `move movey-upload` does not even do any building. Please clarify.

language/tools/move-package/src/resolution/resolution_graph.rs
```diff
@@ -526,8 +540,17 @@ impl ResolvingGraph {
...
+                if !git_info.download_to.exists() {
+                    Self::increase_movey_download_count(
+                        movey_constants::MOVEY_URL.to_string(),
+                        git_info,
+                        build_options.skip_movey,
+                    );
+                }
+            }
 
+            Self::download_and_update_if_remote(*dep_name, dep)?;
```
tnowacki, line 553: I feel like it would make more sense to include this logic in `download_and_update_if_remote`. Is there some reason not to?

language/tools/move-package/src/resolution/resolution_graph.rs
```diff
@@ -572,6 +595,24 @@ impl ResolvingGraph {
         }
         Ok(())
     }
+
+    fn increase_movey_download_count(movey_url: String, git_info: &GitInfo, skip_movey: bool) {
+        if skip_movey {
+            return;
+        }
```
tnowacki, line 602: I feel like it is cleaner to move this to outside of this function. Feels awkward to have a bool flag that is just immediately used to return or not

language/tools/move-package/src/resolution/resolution_graph.rs
```diff
@@ -526,8 +540,17 @@ impl ResolvingGraph {
         };
 
         for (dep_name, dep) in manifest.dependencies.iter().chain(additional_deps.iter()) {
-            Self::download_and_update_if_remote(*dep_name, dep)?;
+            if let Some(git_info) = &dep.git_info {
+                if !git_info.download_to.exists() {
```
tnowacki, line 544: Consider also 
```
let already_downloaded = dep.git_info.map(|gi| gi.download_to.exists()).unwrap_or(true);
if !already_downloaded {
   Self::increase_movey_download_count( ... )
}
```

language/tools/move-package/src/resolution/resolution_graph.rs
```diff
@@ -813,3 +854,85 @@ impl ResolvedPackage {
...
+                .x_www_form_urlencoded_tuple("subdir", test_subdir);
+            then.status(status_code);
+        });
+
+        (server_mock, git_info)
+    }
+
+    fn test_thread_wait(wait_time: Option<u64>) {
+        // make sure the spawn thread has enough time to run
+        let thread_sleep_time = time::Duration::from_millis(wait_time.unwrap_or(20));
```
tnowacki, line 895: nit maybe move `20` to a constant

tnowacki:
In the PR description, you have

> Example: move movey-upload --skip-movey

I think this should be `move build --skip-movey`? I'm not sure how this control flow path will ever be hit in `movey-upload`?

language/tools/move-package/src/resolution/resolution_graph.rs
ea-open-source: This can be resolved depending on the solution of https://github.com/move-language/move/pull/318#discussion_r968187683

language/tools/move-package/src/resolution/resolution_graph.rs
ea-open-source: This can be resolved depending on the solution of https://github.com/move-language/move/pull/318#discussion_r968187683

ea-open-source:
@awelc @tnowacki There are multiple reasons why we made this explicitly opt-in:

1. Most of our existing packages were crawled from Github and they include a lot of unused or for practice packages. We collect these information to generate **Movey** info page about truly important packages, to set them apart from automatically crawled packages.
2. In order to do this, we aim to integrate ourselves into the Move ecosystem as the official Move package repository. Therefore we set opt-in as the default behavior.

ea-open-source:
> Additionally, the `--skip-movey` flag is presented in the description of this PR as an addition to the `move movey-upload`, but it's part of build config and (I think) `move movey-upload` does not even do any building. Please clarify.




> In the PR description, you have
> 
> Example: move movey-upload --skip-movey
> 
> I think this should be `move build --skip-movey`? I'm not sure how this control flow path will ever be hit in `movey-upload`?

Yes, It should be `move build  --skip-movey` not `move movey-upload --skip-movey`. Our bad.

tnowacki (changes requested):
After discussing it, we have decided it would be best if movey was it's own custom source provider. These were recently added, see #318 

This way, movey packages will be resolved with their own syntax, something similar to `CustomPackage = { node: "movey", address: "0x1", package: "custom-package" }`

lwsinclair:
Movey is a centralized registry for Move packages but not an independent repository ("its own custom source provider"). Among the reasons for not being a repository are (1) developers likely prefer Github and this may hamper adoption, and (2) there would be a lot more engineering involved and this would take time and be redundant ("reinventing the wheel"), (3) this would take focus away from features that add value not found elsewhere.

Perhaps what is being suggested here is that Movey should provide a pointer to existing on-chain locations for packages. And indeed, we do have that on our roadmap. At first these links might be informational (no guarantees) later they can be validated and certified by us. 

@tnowacki perhaps we are missing something. What is the compelling advantage?

nhphong:
With the introduction of the new tool `Movey CLI`, this PR is no longer needed. So I'll close it.

ea-open-source:
After further discussions about package dependencies resolving, we have released a new tool [MoveyCLI](https://crates.io/crates/movey_cli). This PR is no longer needed. Closing.

PR #322, opened 2022-07-29
make instruction execution bound an option and increment default value

emmazzz:
## Motivation

This PR makes instruction execution bound an option in the CLI options and increments the default value since 5000 is too low.

## Test Plan

cargo test


wrwg (approved):
Thanks for doing! This bothered me also already.

PR #327, opened 2022-07-31
[move-cli] A proper "docgen" implementation

villesundell:
A rudimentary version of the "docgen" command was implemented in 55a40d583 ("[move-cli] Added "docgen" command (#291)", 2022-07-28). However, the implementation was relying on the Move Prover, only exposing one Docgen switch to the user.

This implementation is using the actual Dogen, giving us a possibility to expose all the Docgen switches to the user. The enable/disable logic is modeled against current default settings for Docgen. This is also an indirect way to communicate the default settings to the user.

Addressing the actual Docgen as move_docgen::Docgen, because both, this subcommand and the Move Docgen are entitled to that name: this subcommand because of the Move CLI coding convention, and the actual Move prover because it's the most descriptive short name.

Removed --quiet, since output on successful execution is now deterministic.

The test is currently testing all the switches, creating (almost) totally opposite settings from the default settings.

## Motivation

The first implementation of "docgen" used Move Prover. Now using the actual Docgen instead to access all the functionality.

Yes.

## Test Plan

In addition to the unit test (`cargo test -p move-cli`) you can test the actual documentation generation by running:

```
cargo build -p move-cli
./target/debug/move --path ./language/documentation/tutorial/step_8/BasicCoin docgen
```

You should get something like:

```
Generated "doc/BasicCoin.md"

Documentation generation successful!
```

And then you should have the final generated markdown documentation at `./language/documentation/tutorial/step_8/BasicCoin/doc/BasicCoin.md`.

### DIFF
--- a/language/tools/move-cli/src/base/docgen.rs
+++ b/language/tools/move-cli/src/base/docgen.rs
@@ -1,48 +1,125 @@
 // Copyright (c) The Move Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::base::prove::{Prove, ProverOptions};
+use super::reroot_path;
 use clap::*;
-use move_package::BuildConfig;
-use std::path::PathBuf;
+use move_docgen::DocgenOptions;
+use move_package::{BuildConfig, ModelConfig};
+use std::{fs, path::PathBuf};
 
-/// Generating javadoc style documentation for packages using the Move Prover
+/// Generate javadoc style documentation for Move packages
 #[derive(Parser)]
 #[clap(name = "docgen")]
 pub struct Docgen {
-    /// Do not print out other than errors
-    #[clap(long = "quiet", short = 'q')]
-    pub quiet: bool,
-    /// A template for documentation generation.
+    /// The level where we start sectioning. Often markdown sections are rendered with
+    /// unnecessary large section fonts, setting this value high reduces the size
+    #[clap(long = "section-level-start", value_name = "HEADER_LEVEL")]
+    pub section_level_start: Option<usize>,
+    /// Whether to exclude private functions in the generated docs
+    #[clap(long = "exclude-private-fun")]
+    pub exclude_private_fun: bool,
+    /// Whether to exclude specifications in the generated docs
+    #[clap(long = "exclude-specs")]
+    pub exclude_specs: bool,
+    /// Whether to put specifications in the same section as a declaration or put them all
+    /// into an independent section
+    #[clap(long = "independent-specs")]
+    pub independent_specs: bool,
+    /// Whether to exclude Move implementations
+    #[clap(long = "exclude-impl")]
+    pub exclude_impl: bool,
+    /// Max depth to which sections are displayed in table-of-contents
+    #[clap(long = "toc-depth", value_name = "DEPTH")]
+    pub toc_depth: Option<usize>,
+    /// Do not use collapsed sections (<details>) for impl and specs
+    #[clap(long = "no-collapsed-sections")]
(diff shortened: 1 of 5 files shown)


villesundell:
Asking review from my mentors @awelc and @tnowacki :memo:

language/tools/move-cli/src/base/docgen.rs
```diff
@@ -1,48 +1,125 @@
...
-/// Generating javadoc style documentation for packages using the Move Prover
+/// Generate javadoc style documentation for Move packages
 #[derive(Parser)]
 #[clap(name = "docgen")]
 pub struct Docgen {
-    /// Do not print out other than errors
-    #[clap(long = "quiet", short = 'q')]
-    pub quiet: bool,
-    /// A template for documentation generation.
+    /// The level where we start sectioning. Often markdown sections are rendered with
```
wrwg, line 14: I wonder whether it would be better to "clapify" the existing `DocgenOptions` and flatten them into the Docgen command?

villesundell: @wrwg That's a good point, that way we could avoid reduplication :+1: Will do that!

villesundell: @wrwg To me it looks like that by flattening we would lose the current "divert only when not default" model. ("_The enable/disable logic is modeled against current default settings for Docgen. This is also an indirect way to communicate the default settings to the user._"). User would need to figure out themselves what are the current defaults, and if to diver from those. This could make it a tad harder for the user: first to figure out the defaults, and then divert from those when/if needed.

Is this OK?

wrwg: Sorry for the delay. I'm not sure I understand. Aren't the current options also equipped with defaults via clap, and aren't those not displayed to the user on help? In any case if you think we need to duplicate this for usability reasons that is also fine with me.

villesundell: @wrwg To me the problem seems to be, that currently (because `move_docgen::Docgen` haven't been clap'd before) the options are mostly [booleans without default values](https://github.com/move-language/move/blob/e2bb7e9822a6c15154149599f3c44f34258b72c0/language/move-prover/move-docgen/src/docgen.rs#L43). Instead defaults are defined separately in [`move_docgen::DocgenOptions::default()`](https://github.com/move-language/move/blob/e2bb7e9822a6c15154149599f3c44f34258b72c0/language/move-prover/move-docgen/src/docgen.rs#L100).

But maybe the true long term solution would be deleting `move_docgen::DocgenOptions::default()` and use clap'd defaults instead? This would require some refactoring, though. Maybe in next iteration? What do you think?

sblackshear: I think deleting `default` and using clap defaults (which is fine to do in a follow-up) makes sense. The doc functionality is quite hidden/underused right now, so I'm not too worried about folks being surprised by changes in defaults.

But the docgen functionality is super valuable, so I am glad to see this made more prominent!

sblackshear (approved):

sblackshear:
@villesundell let me know if there are more changes you want to make here or if this is ready to merge

language/tools/move-cli/src/base/docgen.rs
wrwg: I don't understand why we introduce a total new options type here instead of refactoring the old, but fine for me to merge.

villesundell:
@sblackshear No incoming changes for this iteration of "docgen", ready to merge :+1:

PR #328, opened 2022-07-31
[move-analyzer] Add Docstring rendering on-hover

Deagler:
## Motivation

The existing `move-analyzer` does not render docstrings when you hover over functions, structs, or fields. 
image

Due to the lack of docstring rendering, new Move developers need to keep a separate window open where they can reference the standard library and the Sui/Aptos codebase.

These changes speed up DX by adding docstring rendering when hovering over functions, structs, constants, etc...
image

Yes

## Test Plan

Build the move-analyzer locally and install the move analyzer extension. Now setup `.vscode/settings.json` like this:
```json
{
  "move-analyzer.server.path": "<some path>/move/target/debug/move-analyzer"
}
```

You will see that the docstrings are now attached on-hover (test via the files provided in`move-analyzer/tests/symbols/sources`)


awelc:
I was wondering if you'd consider adding an automated test to the newly extended testing framework to test this feature.

awelc:
Also a bit of heads up, modifying source code of the Move test files will break the move-analyzer tests (`cargo test`) as they hard-code line/column numbers for testing if symbolication returns correct results. 

The only Move source file you can modify freely without worrying about it is `M2.move`

This in addition to the current problem which seems unrelated (`unwrap` is failing at some point).

awelc:
> The only Move source file you can modify freely without worrying about it is `M2.move`

Crap... I guess I was wrong about that... My apologies... You have to adjust line numbers in the test portion of `symbols.rs` or create a new module for the new tests.

Deagler:
> I was wondering if you'd consider adding an automated test to the newly extended testing framework to test this feature.

By newly extended testing framework do you mean the tests in `editors/code/tests` ? Or would it be sufficient to add tests into `symbols.rs`

awelc:
> By newly extended testing framework do you mean the tests in `editors/code/tests` ? Or would it be sufficient to add tests into `symbols.rs`

I see the unit tests (currently in `symbols.rs` but we may decide to move the to a separate file if the number gets totally unwieldy) as more comprehensive ones (e.g., they cover all typed AST nodes to make sure that symbolication works correctly for them) whereas the integration tests are more to guarantee that the info is correctly received by the editor and are in my opinion also worthwhile as they can, for example, indicate changes in the language server protocol upon the editor version upgrade.

Admittedly there are not "integration" tests (`editors/code/tests`) for existing "newer" features, but I will try adding the when I have a spare moment.

So, in short, ideally we'd have both - more functionally testing (if necessary) in unit tests and perhaps just one integration test just to make sure that things work properly between the move analyzer and the editor.

Deagler:
> So, in short, ideally we'd have both - more functionally testing (if necessary) in unit tests and perhaps just one integration test just to make sure that things work properly between the move analyzer and the editor.

I've added a range of unit tests + two integration tests (one that tests docstring recognition within a module, and one for external modules).

jolestar:
@yubing744 bors r+

Help review

yubing744:
@Deagler You don't seem to be dealing with another form of documentation comments /**, which is mentioned in the lexical analysis related code: https://github.com/move-language/move/blob/f618b1b1c0d59960b49ff15147ecdb5de8e72408/language/move-compiler/src/ parser/lexer.rs#L212

Deagler:
> You don't seem to be dealing with another form of documentation comments /**, which is mentioned in the lexical analysis related code

Thanks for the heads up. For some reason I thought Move only supported /// as docstrings. I've added the relevant changes + unit tests + integration test to support that format.

**multi-line asterix docstrings**
image
image

**single-line asterix docstrings**
image
image

jolestar (approved):
LGTM

language/move-analyzer/editors/code/tests/lsp-demo/sources/M6.move
```diff
@@ -0,0 +1,42 @@
+module Symbols::M6 {
```
awelc, line 1: It seems like these test files are duplicates of those in `language/move-analyzer/tests/symbols/sources`. Could we keep just one copy?

I suggest we we adopt the following convention. If a move source file is used only for integration test, it would be in `language/move-analyzer/editors/code/tests/lsp-demo/sources`, but if it's used for both integration and unit tests, let's keep it in `language/move-analyzer/tests/symbols/sources`.

language/move-analyzer/editors/code/src/commands/lsp_command.ts
```diff
@@ -22,3 +22,21 @@ export async function textDocumentDocumentSymbol(
     // Send the request to the language client.
     return client.sendRequest(DocumentSymbolRequest.type, params);
 }
+
+
+/**
+ * An LSP command textDocument/hover
+ */
+export async function textDocumentHover(
```
awelc, line 30: Just for my own information - this is only used in tests, right (and would not be needed otherwise)?

language/move-analyzer/tests/symbols/sources/M1.move
```diff
@@ -133,4 +133,4 @@ module Symbols::M1 {
         }
     }
 
-}
+}
```
awelc, line 136: Could we fix this please so that this file is unchanged?

language/move-analyzer/src/symbols.rs
```diff
@@ -762,14 +772,14 @@ impl Symbolicator {
                 }
             };
 
-            // process the struct itself
```
awelc, line 765: Why is this comment removed?

language/move-analyzer/src/symbols.rs
```diff
@@ -1051,6 +1070,68 @@ impl Symbolicator {
         get_loc(&pos.file_hash(), pos.start(), files, file_id_mapping)
     }
 
+    /// Extracts the docstring (/// or /** ... */) for a given definition by traversing up from the line definition
+    fn extract_doc_string(&self, name_start: &Position, file_hash: &FileHash) -> String {
+        let mut doc_string = String::new();
```
awelc, line 1075: Nit. It would help readability a bit if you decreased the nesting level here. Please consider using the following pattern for the first two if-s:
```
let val = match ... {
  Some(v) => v,
  None => return doc_string,
}
```

awelc (changes requested):
Overall looks great! I requested a few small(ish) changes but we are almost good to go!

language/move-analyzer/src/symbols.rs
```diff
@@ -147,6 +147,8 @@ pub struct UseDef {
     def_loc: DefLoc,
     /// Location of the type definition
     type_def_loc: Option<DefLoc>,
+    /// Doc string for the relevant identifier/function
```
awelc, line 150: I also realized that it might be more intuitive to make the `doc_string` part of `DefLoc` (possibly renaming it to `DefInfo`), particularly keeping in mind that we may want to optimize this code at some point and use a reference to `DefLoc` in `UseDef` rather than "inlining" it.

language/move-analyzer/src/symbols.rs
Deagler: Refactored

language/move-analyzer/src/symbols.rs
Deagler: Accident - my bad. Readded.

language/move-analyzer/tests/symbols/sources/M1.move
Deagler: Done

language/move-analyzer/editors/code/src/commands/lsp_command.ts
Deagler: Yes - this is only used in the integration tests. I've followed the same format as the existing LSP command in that file (`textDocumentDocumentSymbol`).

language/move-analyzer/editors/code/tests/lsp-demo/sources/M6.move
Deagler: I've been unable to reliably access files from `language/move-analyzer/tests/symbols/sources` via the test-runner in `editors/code`. 

Instead, I've stripped the irrelevant portions of code from the tests so they can be used purely as integration tests. Similar to what @yubing744 did for porting `M1.move` over as an integration test.

language/move-analyzer/src/symbols.rs
Deagler: Do you want this change in this PR? 

It will cause a lot of diffs across `symbols.rs` and it will require a few changes to the code itself. I don't think I can include a String inside `DefLoc` as it has the `Copy` trait and `DefLoc` is copied in multiple places across this file.

awelc: > It will cause a lot of diffs across `symbols.rs` and it will require a few changes to the code itself. I don't think I can include a String inside `DefLoc` as it has the `Copy` trait and `DefLoc` is copied in multiple places across this file.
> 
> Unless there's a simple way to do it, I won't have time to make these changes in this PR.

Not necessarily. It's just that after looking at this again it seemed like comment is associated with a def and we have a structure that describes the def already, so it would be a more natural place for the doc string. I think it may be worth doing in the future, though.

language/move-analyzer/editors/code/tests/lsp-demo/sources/M6.move
awelc: It just rubs me the wrong way to have Move source files used for testing (and similar ones at that) in two different locations.

> I've been unable to reliably access files from `language/move-analyzer/tests/symbols/sources` via the test-runner in `editors/code`.

If the opposite is true (you can access `editors/code` tests from symbolicator tests, and if changing your tests to use the same files is not a huge problem, please use tests in `editors/code` for everything and I will move my tests there (will also try to re-do @yubing744's  integration test to use the original, possibly augmented, M1 module code).

Deagler: I can't access `editors/code` tests from symbolicator tests with the current setup because `cargo test` runs the tests at a different level compared to `editors/code`

This causes the reference to `MoveStdLib` to break (see: `Move.toml` in `editors/code`). If I update the path `MoveStdLib` path then the symbolicator tests start working but the `editors/code` tests break.

If you know how to fix that reference to the StdLib then I think what you're proposing will work!

awelc: > If you know how to fix that reference to the StdLib in both cases then I think what you're proposing will work!

Technically we could pull the dependency remotely but perhaps it's indeed not worth it at the moment. Thank you for your patience!

awelc:
@Deagler - linting is failing and it's a required check....

Deagler:
> @Deagler - linting is failing and it's a required check....

Fixed

awelc (approved):

PR #330, opened 2022-08-01
[error] move the debug assertion properly so it doesn't panic under O…

zekun000:
…UT_OF_GAS

## Motivation

(Write your answer here.)

## Test Plan

### DIFF
--- a/language/move-binary-format/src/errors.rs
+++ b/language/move-binary-format/src/errors.rs
@@ -82,10 +82,17 @@ impl VMError {
                 VMStatus::Error(StatusCode::ABORTED)
             }
 
-            // TODO Errors for OUT_OF_GAS do not always have index set
             (major_status, sub_status, location)
                 if major_status.status_type() == StatusType::Execution =>
             {
+                let abort_location = match &location {
+                    Location::Script => vm_status::AbortLocation::Script,
+                    Location::Module(id) => vm_status::AbortLocation::Module(id.clone()),
+                    Location::Undefined => {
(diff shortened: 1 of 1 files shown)


vgao1996:
@tnowacki I remember you trying a similar fix a long time ago... Any idea what prevented that from happening?

tnowacki:
This seems like a reasonable change. 
And @vgao1996, Bob was going to take a look but never got around to it IIRC

language/move-binary-format/src/errors.rs
```diff
@@ -82,10 +82,17 @@ impl VMError {
...
                 if major_status.status_type() == StatusType::Execution =>
             {
+                let abort_location = match &location {
+                    Location::Script => vm_status::AbortLocation::Script,
+                    Location::Module(id) => vm_status::AbortLocation::Module(id.clone()),
+                    Location::Undefined => {
+                        return VMStatus::Error(major_status);
+                    }
+                };
+                // Errors for OUT_OF_GAS do not always have index set, it should already return above.
```
tnowacki, line 95: This should read a bit tighter than that
```suggestion
                // Errors for OUT_OF_GAS do not always have index set: if it does not, it should already return above.
```

tnowacki (approved):

vgao1996:
@zekun000 we don't have bors in this repo any more... You simply need to make sure all your tests pass and then there will be a rebase button you can push. (In this particular case, it's likely a flaky test. I've issued a re-run for you)

PR #334, opened 2022-08-03
[vm] Invalidate and flush code loader cache

wrwg:
This is a hack to workaround well-known issues with the VMs code loader cache. This cache is not designed to support module upgrade or deletion. This leads to situations where the cache does not reflect the state of storage:

1. On module upgrade, the upgraded module is in storage, but the old one still in the cache.
2. On an abandoned code publishing transaction, the cache may contain a module which was never committed to storage by the adapter.

The solution is to add a flag to `Loader` marking it as 'invalidated'. For scenario (1), the VM sets the flag itself. For scenario (2), a public API allows the adapter to set the flag.

If the cache invalidated, it can (and must) still be used until the end of the session. This is because during session execution, data might have been extracted which indirectly refers to the current cache (e.g. `CachedStructIndex`). The adapter is responsible to flush an invalidated cache.

### DIFF
--- a/language/move-vm/runtime/src/loader.rs
+++ b/language/move-vm/runtime/src/loader.rs
@@ -442,7 +442,7 @@ pub(crate) struct Loader {
     type_cache: RwLock<TypeCache>,
     natives: NativeFunctions,
 
-    // A boolean indicating that this cache has become invalid and needs to be flushed
+    // A boolean indicating that this cache has become outdated and needs to be flushed
     // for the next session. This is to work around multiple bugs in the loader architecture:
     //
     // - After module upgrade, the new module version isn't yet in the cache, which causes
@@ -477,7 +477,7 @@ impl Loader {
     }
 
     /// Mark this cache as invalidated.
-    pub(crate) fn invalidate(&self) {
+    pub(crate) fn mark_as_invalid(&self) {
         *self.invalidated.write() = true;
     }
 
--- a/language/move-vm/runtime/src/loader.rs
+++ b/language/move-vm/runtime/src/loader.rs
@@ -442,15 +442,31 @@ pub(crate) struct Loader {
     type_cache: RwLock<TypeCache>,
     natives: NativeFunctions,
 
-    // A boolean indicating that this cache has become outdated and needs to be flushed
-    // for the next session. This is to work around multiple bugs in the loader architecture:
+    // The below field supports a hack to workaround well-known issues with the
+    // loader cache. This cache is not designed to support module upgrade or deletion.
+    // This leads to situations where the cache does not reflect the state of storage:
     //
-    // - After module upgrade, the new module version isn't yet in the cache, which causes
-    //   inconsistencies if this module is accessed subsequently. The logic for upgrading
-    //   inside of this module will set this flag.
-    // - An external call on the VM may set this flag as well. For example, if module
-    //   publishing succeeds but the transaction is not committed to global storage, the
-    ///  module is still in the cache if it shouldn't.
+    // 1. On module upgrade, the upgraded module is in storage, but the old one still in the cache.
+    // 2. On an abandoned code publishing transaction, the cache may contain a module which was
+    //    never committed to storage by the adapter.
+    //
+    // The solution is to add a flag to Loader marking it as 'invalidated'. For scenario (1),
+    // the VM sets the flag itself. For scenario (2), a public API allows the adapter to set
+    // the flag.
+    //
+    // If the cache is invalidated, it can (and must) still be used until there are no more
+    // sessions alive which are derived from a VM with this loader. This is because there are
+    // internal data structures derived from the loader which can become inconsistent. Therefore
+    // the adapter must explicitly call a function to flush the invalidated loader.
+    //
+    // This code (the loader) needs a complete refactoring. The new loader should
+    //
+    // - support upgrade and deletion of modules, while still preserving max cache lookup
+    //   performance. This is essential for a cache like this in a multi-tenant execution
+    //   environment.
+    // - should delegate lifetime ownership to the adapter. Code loading (including verification)
+    //   is a major execution bottleneck. We should be able to reuse a cache for the lifetime of
+    //   the adapter/node, not just a VM or even session (as effectively today).
     invalidated: RwLock<bool>,
 }
 
--- a/language/move-vm/runtime/src/loader.rs
+++ b/language/move-vm/runtime/src/loader.rs
@@ -441,6 +441,17 @@ pub(crate) struct Loader {
     module_cache: RwLock<ModuleCache>,
     type_cache: RwLock<TypeCache>,
     natives: NativeFunctions,
+
+    // A boolean indicating that this cache has become invalid and needs to be flushed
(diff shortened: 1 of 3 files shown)


language/move-vm/runtime/src/runtime.rs
```diff
@@ -193,6 +201,11 @@ impl VMRuntime {
 
         // All modules verified, publish them to data cache
         for (module, blob) in compiled_modules.into_iter().zip(modules.into_iter()) {
+            if data_store.exists_module(&module.self_id())? {
+                // This is an upgrade, so invalidate the loader cache, which still contains the
+                // old module.
+                self.loader.invalidate();
```
vgao1996, line 207: Still curious, can we actually just flush the cache here?

vgao1996 (commented):
> If the cache invalidated, it can (and must) still be used until the end of the session. This is because during session execution, data might have been extracted which indirectly refers to the current cache (e.g. CachedStructIndex)

This doesn't feel totally right to me. When a republishing happens, it means the loader has entered a bad state and executing any additional transactions could cause inconsistency issues. I was actually expecting us to check if the flag is set when calling `execute_function` or other entry points. Am I missing something?

wrwg:
> > If the cache invalidated, it can (and must) still be used until the end of the session. This is because during session execution, data might have been extracted which indirectly refers to the current cache (e.g. CachedStructIndex)
> 
> This doesn't feel totally right to me. When a republishing happens, it means the loader has entered a bad state and executing any additional transactions could cause inconsistency issues. I was actually expecting us to check if the flag is set when calling `execute_function` or other entry points. Am I missing something?

There is the possibility that further Move code is executed within the given transaction. As those code publishing transactions are now normal transactions, at least the epilogue will be executed, but in general other code possibly as well. 

There is no "bad state" in which the VM is. It's only it doesn't yet know for this session about the new code. This is actually even what you want, because the upgraded code may actually be involved in the current transaction (self-upgrade).

language/move-vm/runtime/src/runtime.rs
wrwg: See top-level comment. Meng did something similar for the epilogue problem if I read this right in https://github.com/diem/diem/pull/7367.

vgao1996:
Okay I see your point. You are basically saying that an adapter may still want to execute some code (e.g. epilogue) after republishing a module, and it's safe to do so if the adapter knows that code doesn't touch the republished module.

vgao1996 (approved):
Looks good to me

language/move-vm/runtime/src/loader.rs
```diff
@@ -450,9 +461,26 @@ impl Loader {
...
+        if *invalidated {
+            *self.scripts.write() = ScriptCache::new();
+            *self.module_cache.write() = ModuleCache::new();
+            *self.type_cache.write() = TypeCache::new();
+            *invalidated = false;
+        }
+    }
+
+    /// Mark this cache as invalidated.
+    pub(crate) fn invalidate(&self) {
```
zekun000, line 480: probably should name mark_for_invalidation since it doesn't do the invalidate

language/move-vm/runtime/src/loader.rs
```diff
@@ -450,9 +461,26 @@ impl Loader {
             module_cache: RwLock::new(ModuleCache::new()),
             type_cache: RwLock::new(TypeCache::new()),
             natives,
+            invalidated: RwLock::new(false),
         }
     }
 
+    /// Flush this cache if it is marked as invalidated.
+    pub(crate) fn flush_if_invalidated(&self) {
```
zekun000, line 469: better name invalidate_if_marked?

language/move-vm/runtime/src/runtime.rs
```diff
@@ -49,7 +49,13 @@ impl VMRuntime {
         })
     }
 
+    pub(crate) fn invalidate_loader_cache(&self) {
+        self.loader.invalidate()
+    }
+
     pub fn new_session<'r, S: MoveResolver>(&self, remote: &'r S) -> Session<'r, '_, S> {
+        // If the loader cache is invalidated, flush it for the new session
+        self.loader.flush_if_invalidated();
```
zekun000, line 58: should we just call new_session_with_extensions here with NativeContextExtensions::default()?

meng-xu-cs:
This PR looks good to me as well.

I am also worrying about a case on parallel execution:

Say, in a block where there are many transactions `[T1, T2, ..., Tn]`. Among which `[..., Te, ..., Tm ...,]` are two transactions that interest us:
- `Te` executes a function in module `M` and
- `Tm` updates module `M`.

The old module `M` can be
```move
module 0x2::M {
  struct R1 has key {};
  public entry foo(s: signer) {
    move_to(&s, R1 {});
  }
}
```
And the updated module `M` can be
```diff
module 0x2::M {
  struct R1 has key {};
+ struct R2 has key {};
  public entry foo(s: signer) {
    move_to(&s, R1 {});
+   move_to(&s, R2 {}); 
  }
}
```

This is a fully compatible upgrade and the VM will happily accept that.

However, this does not mean that `Te` and `Tm` can be parallelized (in different VMs or in different sessions), even after this PR. The ordering of `Te` and `Tm` must be ensured by some dependency analysis which is out-of-scope for the VM and is probably the job of the adapter or even the layers above.

zekun000:
this pr is not meant to solve the parallel execution, we'll need another longer discussion for that.

language/move-vm/runtime/src/loader.rs
wrwg: I think the current name is right. "If the cash has been marked as invalidated, flush it now".

language/move-vm/runtime/src/runtime.rs
```diff
@@ -49,19 +49,21 @@ impl VMRuntime {
...
+        self.new_session_with_extensions(remote, NativeContextExtensions::default())
     }
 
     pub fn new_session_with_extensions<'r, S: MoveResolver>(
         &self,
         remote: &'r S,
         native_extensions: NativeContextExtensions<'r>,
     ) -> Session<'r, '_, S> {
+        // If the loader cache is invalidated, flush it for the new session
+        self.loader.flush_if_invalidated();
```
tnowacki, line 66: Should we do this automagically like this? 

I thought we would just poison it, block all VM session actions, and require the adapter to manually flush the cache

wrwg: I prefer to do this automatically. It's a bug in the VM and we should fix it on VM side. The real fix (hopefully coming soon) is rewrite the loader to be actually updatable.

tnowacki: My concern is it could cause other silent failures. Lot's of endpoints (`load_script`, `load_function`, `load_type`, `get_type_layout`, etc) have pointers back into the cache. 
Invalidating the cache without warning the adapter I think is going to cause a lot of accidental errors here, unless I'm missing something.

wrwg: Ok, the assumption here was that no one has more than one session open at a time, but you are right that we can't enforce this. Removed the automatic flushing now and exposed another public function so the adapter can do the flushing.

PR #341, opened 2022-08-04
[vm] Add ability for adapter to observe module loader cache hits

wrwg:
If a VM is used to execute multiple sessions, the adapter is not cabable to see which modules have been loaded by a given transaction, because modules are cached internaly by the VM, and no call back to the adapters data store happens.

This PR adds a small change which allows the adapter to read out this information, roughly as follows:

```
let vm = MoveVM::new();
let session = vm.new_session();
<<work with this session>>
let cache_hits = vm.get_and_clear_module_cache_hits();
let used_modules = cache_hits + modules_read_from_storage()
// next session starts with 0 cache hits
let session = vm.new_session();
...
```

### DIFF
--- a/language/move-vm/runtime/src/loader.rs
+++ b/language/move-vm/runtime/src/loader.rs
@@ -468,6 +468,11 @@ pub(crate) struct Loader {
     //   is a major execution bottleneck. We should be able to reuse a cache for the lifetime of
     //   the adapter/node, not just a VM or even session (as effectively today).
     invalidated: RwLock<bool>,
+
+    // Collects the cache hits on module loads. This information can be read and reset by
+    // an adapter to reason about read/write conflicts of code publishing transactions and
+    // other transactions.
+    module_cache_hits: RwLock<BTreeSet<ModuleId>>,
 }
 
 impl Loader {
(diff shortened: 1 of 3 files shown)


vgao1996 (approved):
Looks reasonable. Let's get this in first.

language/move-vm/runtime/src/loader.rs
```diff
@@ -888,6 +922,7 @@ impl Loader {
     ) -> VMResult<Arc<Module>> {
         // if the module is already in the code cache, load the cached version
         if let Some(cached) = self.module_cache.read().module_at(id) {
+            self.module_cache_hits.write().insert(id.clone());
```
vgao1996, line 925: Let's say we have modules A and B, where A uses B. Assume B is already loaded. Then when we load A, we will register B as a cache hit correctly, right? The algorithm seems correct to me.

junkil-park (approved):
LGTM

zekun000 (approved):

language/move-vm/runtime/src/loader.rs
wrwg: Yes!

PR #343, opened 2022-08-04
[move-prover][interpreter] Add unit test for choice operator

ckaiwxm:
## Motivation

- Add unit test for address type domain for choice operators.

Yes.

## Test Plan

### DIFF
--- a/language/move-prover/interpreter-testsuite/tests/concrete_check/choice.move
+++ b/language/move-prover/interpreter-testsuite/tests/concrete_check/choice.move
@@ -1,6 +1,28 @@
 module 0x2::A {
     use std::vector;
 
+    struct R has key { i: u64 }
+
+    #[test(a=@0x2)]
+    public fun unroll_address_success(a: &signer) {
+        let r = R{i: 1};
+        move_to(a, r);
+    }
+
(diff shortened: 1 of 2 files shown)


language/move-prover/interpreter-testsuite/tests/concrete_check/choice.move
```diff
@@ -1,6 +1,20 @@
 module 0x2::A {
+    use std::signer;
     use std::vector;
 
+    struct R has key { i: u64 }
+
+    #[test(a=@0x2)]
+    public fun move_to_test(a: &signer) {
```
meng-xu-cs, line 8: For consistency, maybe call this test 'unroll_address_satisfiable'? And correspondingly, there should be an unsatisfiable case?

meng-xu-cs (commented):
Thanks for the test case!

language/move-prover/interpreter-testsuite/tests/concrete_check/choice.move
ckaiwxm: Make a lot of sense! Renamed and added test case

meng-xu-cs (approved):

PR #344, opened 2022-08-04
fix example coin-swap error lib usage

zoey-t:
## Motivation

The example coin-swap has compilation error
Issue is the usage of error in move std

Yes, I have read it and have run `cargo x lint && cargo xfmt && cargo xclippy --all-targets`

## Test Plan

I only updated the error lib usage according to move std lib document


wrwg (approved):
Thanks!

PR #351, opened 2022-08-06
display bitwise or operation incorrectly.

zouguangxian:


sblackshear (approved):

tnowacki (commented):
I think this change was necessary to get it rendered on github properly. So this might need to be reverted

tnowacki:
Actually it looks fine. All I remember is some weird edge cases... strange

Issue #346, opened 2022-08-05
[Feature Request] Add username to package_lock file


fardream:
# 🚀 Feature Request

Add user name to the package lock file created by `language/tools/move-package/src/package_lock.rs`

## Motivation

We develop on a shared linux host. If one user creates the lock file (`language/tools/move-package/src/package_lock.rs`), no other user can lock the file and all builds fails.

## Pitch

If we can append user name to the package lock file (such as `/tmp/move_package_lock_{username}.lock`), different users will not impact each other's build.

(I can open a pr for this).

sblackshear:
Good find! Adding the username makes sense + would gladly accept a PR for this.

PR #354, opened 2022-08-07
[move-cli] add username to package lock file.

fardream:
This is to address #346.

## Motivation

The same package lock file is used for different users (e.g.
`/tmp/move_package_lock.lock`), this is problematic on hosts shared by
many users. This add the username to the end of the package lock file
(e.g. `/tmp/move_package_Lock_username.lock`).

Yes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3447,6 +3447,7 @@ dependencies = [
  "termcolor",
  "toml",
  "walkdir",
+ "whoami",
 ]
 
 [[package]]
@@ -6734,6 +6735,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d743fdedc5c64377b5fc2bc036b01c7fd642205a0d96356034ae3404d49eb7fb"
 dependencies = [
  "cc",
+]
+
+[[package]]
+name = "whoami"
+version = "1.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "524b58fa5a20a2fb3014dd6358b70e6579692a56ef6fce928834e488f42f65e8"
+dependencies = [
+ "wasm-bindgen",
+ "web-sys",
 ]
 
 [[package]]
--- a/language/tools/move-package/Cargo.toml
+++ b/language/tools/move-package/Cargo.toml
@@ -43,6 +43,8 @@ termcolor = { version = "1.1.2", optional = true }
 hex = { version = "0.4.3", optional = true }
 reqwest = { version = "0.10", features = ["blocking", "json"] }
 
+whoami = { version = "1.2.1" }
+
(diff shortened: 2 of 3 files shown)


tnowacki:
I think this opens up some weird edge cases with multiple users trying to compile the same code in the same location. But I imagine this isn't going to be very common.

tnowacki:
I also worry that this exposes to strange git edge cases, as described in the file. 

> Without this lock we can easily get into race conditions around caching and overwriting of packages (e.g., thread 1 and thread 2 compete to build package P in the same location), as well as downloading of git dependencies (thread 1 starts downloading git dependency, meanwhile thread 2 sees the git directory before it has been fully populated but assumes it has been fully downloaded and starts building the package before the git dependency has been fully downloaded by thread 1. This will then lead to file not found errors). These same issues could occur across processes, this is why we grab both a thread lock and process lock.

Thoughts about this @fadream?

vgao1996:
Does rust or other languages do the same thing?

fardream:
> I also worry that this exposes to strange git edge cases, as described in the file.
> 
> > Without this lock we can easily get into race conditions around caching and overwriting of packages (e.g., thread 1 and thread 2 compete to build package P in the same location), as well as downloading of git dependencies (thread 1 starts downloading git dependency, meanwhile thread 2 sees the git directory before it has been fully populated but assumes it has been fully downloaded and starts building the package before the git dependency has been fully downloaded by thread 1. This will then lead to file not found errors). These same issues could occur across processes, this is why we grab both a thread lock and process lock.
> 
> Thoughts about this @fadream?

This will prevent the same user building move packages on the same host concurrently, but will fail if two different users are running build in the same directory. To be honest, I rarely run into a scenario where two users are building in the same location.

After further reading the code, it looks like the dependencies are downloaded and all codes are compiled in the `build` directory where `Move.toml` is located - if this is true, shall we place the lock file in the `build` directory?

tnowacki:
> if this is true, shall we place the lock file in the build directory?

This seems like a much more sensible option to me. But I know it would be a larger PR.

I agree that two users building in the same directory is not very likely, so I'm happy to take this change as is. But would prefer a build-dir lock if possible

fardream:
> > if this is true, shall we place the lock file in the build directory?
> 
> This seems like a much more sensible option to me. But I know it would be a larger PR.
> 
> I agree that two users building in the same directory is not very likely, so I'm happy to take this change as is. But would prefer a build-dir lock if possible

It looks like changing to build would be a quite a big change (the code right now cannot control the location of the lock file and doesn't take parameters), so I think maybe we can merge this for now and add a build location specific lock later.

tnowacki (approved):
@fardream that seems reasonable to me, thanks!

PR #356, opened 2022-08-07
Make abi generation work copyable struct txn arguments

areshand:
## Motivation
Abi generation current is not working struct txn arguments. 
Without abi files, the language sdk builder doesn't work.

This PR is to
1. generate abi files for script function using struct as transaction arguments.
2.  the struct has to have copy ability and no key ability

Note: this PR only make abi generation work. Whether the struct is allowed is controlled by VM adapter and not relaxed by this PR.

Yes
(Write your answer here.)

## Test Plan

add unit test


wrwg (approved):
Awesome!

vgao1996 (approved):
LGTM. Please fix the lint.

PR #357, opened 2022-08-07
[packages] New dependency type 'node' + custom package info properties

wrwg:
This adds a new dependency type to the package manifest which allows to specify a node from which to receive a package:

```
[dependencies]
MoveStdlib = { node: "localhost:8080", address: "0x1", package: "move-stdlib" }
```

Resolution of such kind of a dependency is handled via a new trait `PackageHooks` which can be globally registered on a per-process basis. If no hook is registered, an error will be produced that `node` deps are not supported.

> Note: After investigating some time whether the resolver could be passed in as a parameter, it became clear that a larger refactoring woul be needed as the package system is simply not designed for compositionality like this.

In addition to the new dependency type, this also allows additional fields in the `[package]` section, as in

```
[package]
name = "Foobar"
upgrade_policy = ..
```

The valid extra names are again determined by the `PackageHooks`.

## Motivation

Customizable package system

Yes

## Test Plan

New smoke test


language/tools/move-package/src/package_hooks.rs
```diff
@@ -0,0 +1,45 @@
...
+pub trait PackageHooks {
+    /// Returns custom fields allowed in `PackageInfo`.
+    fn custom_package_info_fields(&self) -> Vec<String>;
+
+    /// A resolver for `node` dependencies in the manifest. This is called to download the
+    /// dependency from the node into the `info.local_path` location, similar as with git
+    /// dependencies.
+    fn resolve_node_dependency(&self, dep_name: Symbol, info: &NodeInfo) -> anyhow::Result<()>;
+}
+static HOOKS: Lazy<Mutex<Option<Box<dyn PackageHooks + Send + Sync>>>> =
```
vgao1996, line 21: Could I ask why we're implementing this as a global instead of passing in the resolver as an argument? To avoid breaking the existing API?

wrwg: Please see the PR description. Its a large refactoring touching move-cli and move-packages. 

- Pass it as `<X>` you now have the type parameter everywhere (going deep down into static functions)
- Pass it as `dyn X` you can attach this do any of the existing context data structures because they need to be serializable/copyable etc. which doesn't work with traits.

This thing is not well designed if it comes to compositionally and extensibility. Really, any reasonable tool should have some global `Context`, `Env` whatever which is uniformly used everywhere, and does have bounds so you can add arbitrary fields. The move model has it, the vm, and it was added to the compiler sometime last year. 

Getting away without this global context is pure functional programming ideology, and it doesn't work in practice as we are seeing here once again.

vgao1996: > Pass it as dyn X you can't attach this to any of the existing context data structures because they need to be serializable/copyable etc. which doesn't work with traits.

I think you can make `Clone, Serialize etc.` super traits, no? Anyway I feel like this is an orthogonal issue you'll have to deal with no matter you use a global context or manually pass it around. 

> Really, any reasonable tool should have some global Context, Env whatever which is uniformly used everywhere, and does have no bounds so you can add arbitrary fields. The move model has it, the vm, and it was added to the compiler sometime last year.

> Getting away without this global context is pure functional programming ideology, and it doesn't work in practice as we are seeing here once again.

Yeah I have fixed feelings about this. On one hand I love the fact that passing the global context around manually makes it easy to reason about side effects, but on the other hand I also I agree that it can get quite annoying, especially when you have a complex call graph. I must also admit that Rust has never been "global-free" in the first place, with its usage of the global allocator and stdio. Some popular logging crates also provide a rather smooth user experience using global variables. To this point this actually feels like a philosophical question to me and I guess I really need to study this more ; )

Anyway, I think I'm fine with the current design, and I actually think it is the right thing to do to allow us to get the functionalities we need first, even in the case we eventually decide to switch to the other way around.

tnowacki:
Could you explain the meaning behind the name `node`? Maybe it will make more sense to me with some context

language/tools/move-package/src/package_hooks.rs
tnowacki: With the lack of some global env/context for the package system, would it be enough to stick this thing in the `ResolutionGraph` for now? where it would be passed in during `ResolutionGraph::new`

My only real worry with sticking this in `static` (the `static HOOKS` if I'm following this closely) is discoverability. Discoverability in the sense that it is a bit of a gotcha to know it exists, a global env or anything similar is always nice and in your face when you are constructing this tool, but obviously out of scope for this PR.

wrwg: @tnowacki that is a good point so I tried to stick this into the `ResolutionGraph`. It worked to some extend (need to remove a number of `Clone` derives) but it failed with the manifest parser which is just a bunch of static functions.

I'm wondering whether the static hooks are actually the cleaner solution anyway instead of bloating up the code. The thing with this kind of hooks is that there are once set at program startup and then stay fixed. It's a bit like loading a code plugin, different than using a trait which can represent different things during process execution.

wrwg:
> Could you explain the meaning behind the name `node`? Maybe it will make more sense to me with some context

(full-)node -- any kind of network endpoint which can be used to talk to some blockchain. Didn't found anything better. `chain` or `network` did not appear better.

wrwg:
Any comment on how to move forward with this? I really like to refactor the package system significantly to avoid the need for such static `hooks`, but its not the right time for this.

language/tools/move-package/src/package_hooks.rs
tnowacki: I think the hooks are fine for now, a small comment saying something like "this should really be in a global env/context but the package system doesn't have one " might be helpful, but not really necessary

tnowacki:
> Any comment on how to move forward with this? I really like to refactor the package system significantly to avoid the need for such static hooks, but its not the right time for this.

Agreed. 

> (full-)node -- any kind of network endpoint which can be used to talk to some blockchain. Didn't found anything better. chain or network did not appear better.

It is indeed a tricky one... Can we think of other non-blockchain locations we might use here? I could imagine this also being used for some third party package repository (thought I think we are looking into alternatives for that one specifically)

wrwg:
> > Any comment on how to move forward with this? I really like to refactor the package system significantly to avoid the need for such static hooks, but its not the right time for this.
> 
> Agreed.
> 
> > (full-)node -- any kind of network endpoint which can be used to talk to some blockchain. Didn't found anything better. chain or network did not appear better.
> 
> It is indeed a tricky one... Can we think of other non-blockchain locations we might use here? I could imagine this also being used for some third party package repository (thought I think we are looking into alternatives for that one specifically)

What about `url`? Couldn't get more generic...

language/tools/move-package/src/source_package/parsed_manifest.rs
```diff
@@ -43,6 +44,7 @@ pub struct Dependency {
     pub version: Option<Version>,
     pub digest: Option<PackageDigest>,
     pub git_info: Option<GitInfo>,
+    pub node_info: Option<NodeInfo>,
```
tnowacki, line 47: Are these two mutually exclusive? Should it be something like `external_info`?

tnowacki (commented):
Maybe something like `external_node` or `external_location` or something might be helpful?

I do think though that some use cases might help name this. 
`node` isn't too bad for some use cases
`node = "aptos"` or `node = "sui"` feels reasonable?
If it was used for something like Movey, does that work too? `node = "Movey"`

language/tools/move-package/src/source_package/parsed_manifest.rs
wrwg: Very different

tnowacki (approved):
I'll defer to your judgement for now on naming as I don't know how exactly it will be used. Maybe we can discuss it on Thursday? 

But I'll approve this for now with the assumption that there might be some breaking changes in the future

wrwg:
> Maybe something like `external_node` or `external_location` or something might be helpful?
> 
> I do think though that some use cases might help name this. `node` isn't too bad for some use cases `node = "aptos"` or `node = "sui"` feels reasonable? If it was used for something like Movey, does that work too? `node = "Movey"`

Well now I renamed it to `url` and it is used as in `Name = { url = "localhost:8080" }` or any other endpoint which supports the protocol the provider gives to the custom extension. So I think `url` is actually good. If we see the need we can at a later point do `{ url = "https://movey.org", protocol = "movey", ... }` (or `protocol = "sui" or whatever). Logic here would be protocol has a default in a particular application context.

wrwg:
@tnowacki comment let me gave it a 2nd thought and now one can, via the hooks, actually customize the name of custom dependency. So the hooks can choose whether its `node`, `url`, `movey` or whatever.

Issue #353, opened 2022-08-06
Provide Chinese Move Book


geometryolife:
Hello, developers who contribute to Move language. I am an open source enthusiast, I have learned a lot of knowledge from the open source community, so that I have grown up well. At the same time, I am also a novice to explore Web3.
In the past three months, I have learned what is a blockchain and what is Web3. I have been looking forward to the world of Web3.
Move is the first smart contract programming language I contacted, and I already like this language. 
I want to introduce this language to more Chinese friends who are yearning for Web3, so I set up a [*Move Chinese community (MoveCC)*](https://github.com/move-cc) to preach to Chinese developers.
I am planning to [translate the Move Book](https://github.com/move-cc/move-book-zh), I have done a small step now.

jolestar:
I also want to propose this, and I suggest maintaining a multi-language version at https://github.com/move-language/move/tree/main/language/documentation/book.

@sblackshear @wrwg

sblackshear:
I am very supportive of this! 

Perhaps what we can do is add a `book/translations` directory and add the Chinese version in a subdirectory there? Will happily approve a PR for this, and I expect we'll see other translations soon enough (e.g., there are a lot of Turkish Move developers these days...)

But of course, also happy to keep translation in a separate repo and maintain a markdown list that points to translations in external repos. Let me know what you prefer @geometryolife--you are the one that's done the hard work here!

lshoo:
This is another Chinese version of the Move book: [Move Book Chinese version](https://github.com/move-dao/move-book-zh/) by MoveDAO, we started from  July 25,.
@sblackshear @jolestar

geometryolife:
You have a high degree of completion. If you can migrate your results to `book/translations/move-book-zh`, we can improve `move-book-zh` faster. During the migration process, the English content needs to be retained first. In the machine translation process, some charts are not translated, only the text content is translated. @lshoo

lshoo:
good idea @geometryolife

ChengYueJia:
hi everyone, we we can discuss the mdbook-i18n-solution.  #512 
Waiting for your opinion ~~

PR #359, opened 2022-08-08
migrate Move Book of Chinese version from MoveCC

geometryolife:
## Motivation

After the discussion in *[Provide Chinese Move Book #353](https://github.com/move-language/move/issues/353)*, I migrated the Move Book source code that was machine translated to `book/translations/move-book-zh`.

My initial plan is to delete the English source code after correcting the Chinese content. At present, these source files are in a mixed state of Chinese and English, which is conducive to translation work.

yes

## Test Plan


sblackshear (approved):
CC'ing some Move core contributors + Mandarin speakers who might be able to help with future reviews!

I am a lame English-only speaker and don't want to burden these folks with reviewing the entire translation, so let's merge as-is. Thank you!!

PR #369, opened 2022-08-10
[verifier] add option to limit nested loop depth

vgao1996:
## Summary
This allows the bytecode verifier to limit the depth of nested loops. To enable this check, you'll need to pass in a `VerifierConfig` via the new entry points (`xxx_with_config`). The old entry points are kept to maintain backward compatibility, but they delegate to the new entry points under the hood.

## Test Plan
New verifier unit tests & move vm integration tests have been added to ensure this check can be activated properly

### DIFF
--- a/language/move-bytecode-verifier/src/code_unit_verifier.rs
+++ b/language/move-bytecode-verifier/src/code_unit_verifier.rs
@@ -7,7 +7,7 @@
 //! abstract_interpreter.rs. CodeUnitVerifier simply orchestrates calls into these two files.
 use crate::{
     acquires_list_verifier::AcquiresVerifier, control_flow, locals_safety, reference_safety,
-    stack_usage_verifier::StackUsageVerifier, type_safety,
+    stack_usage_verifier::StackUsageVerifier, type_safety, verifier::VerifierConfig,
 };
 use move_binary_format::{
     access::ModuleAccess,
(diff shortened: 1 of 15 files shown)


wrwg (approved):
LGTM but I'm not acquainted with the verifier code so need another set of eyes

language/move-bytecode-verifier/src/verifier.rs
```diff
@@ -16,6 +16,19 @@ use move_binary_format::{
     file_format::{CompiledModule, CompiledScript},
 };
 
+#[derive(Debug, Clone)]
+pub struct VerifierConfig {
+    pub max_loop_depth: usize,
```
tnowacki, line 21: Might be a bit cleaner to just make this an `Option`?

And thoughts on sticking in the config for the script verifier? Making the default `no_additional_script_signature_checks`?

language/move-bytecode-verifier/src/code_unit_verifier.rs
```diff
@@ -28,26 +28,33 @@ pub struct CodeUnitVerifier<'a> {
 }
 
 impl<'a> CodeUnitVerifier<'a> {
-    pub fn verify_module(module: &'a CompiledModule) -> VMResult<()> {
-        Self::verify_module_impl(module).map_err(|e| e.finish(Location::Module(module.self_id())))
+    pub fn verify_module(module: &'a CompiledModule, max_loop_depth: usize) -> VMResult<()> {
```
tnowacki, line 31: Would it make more sense to just pass in the config?

language/move-bytecode-verifier/src/verifier.rs
```diff
@@ -27,6 +40,10 @@ use move_binary_format::{
 /// minimize the code locations that need to be updated should a new checker
 /// is introduced.
 pub fn verify_module(module: &CompiledModule) -> VMResult<()> {
+    verify_module_with_config(module, &VerifierConfig::default())
+}
+
+pub fn verify_module_with_config(module: &CompiledModule, config: &VerifierConfig) -> VMResult<()> {
```
tnowacki, line 46: Random nit, typically we stick these sort of contexts/configs before the module, as the arg is shared between the various calls. But this is fine and I don't really care

language/move-bytecode-verifier/src/code_unit_verifier.rs
vgao1996: Sounds like a good idea. This way we won't have to touch the argument passing part again if we want to add another one in the future.

language/move-bytecode-verifier/src/verifier.rs
vgao1996: You mean having something like `fn verify_script(config: &VerifierConfig, ..)` and `fn verify_script_with_default_config(..)`?

vgao1996: NVM I think I misunderstood. 

> And thoughts on sticking in the config for the script verifier? Making the default 
I get you mean. Sounds like a great idea. We can do that in a separate PR

PR #371, opened 2022-08-11
[move-cli]: run test ignore compiler warnings

lerencao:
## Motivation

When runing `move test`, if compiling package produces warnings, then test will not run.
But most time, we just want to ignore the warnings, and fix it later after test ok.
I'd like to add a config entry to control the behavior.

yes

## Test Plan

### DIFF
--- a/language/tools/move-cli/src/base/test.rs
+++ b/language/tools/move-cli/src/base/test.rs
@@ -60,6 +60,11 @@ pub struct Test {
     /// Show the storage state at the end of execution of a failing test
     #[clap(name = "global_state_on_error", short = 'g', long = "state_on_error")]
     pub report_storage_on_error: bool,
+
+    /// Ignore compiler's warning, and continue run tests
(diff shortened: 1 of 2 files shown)


wrwg (approved):
Very nice, thank you!

language/tools/move-cli/src/base/test.rs
```diff
@@ -60,6 +60,11 @@ pub struct Test {
     /// Show the storage state at the end of execution of a failing test
     #[clap(name = "global_state_on_error", short = 'g', long = "state_on_error")]
     pub report_storage_on_error: bool,
+
+    /// Ignore compiler's warning, and continue run tests
+    #[clap(name = "ignore_compile_warnings", long = "ignore_compile_warnings")]
+    pub ignore_compile_warnings: bool,
```
tnowacki, line 66: This defaults to false right? 

I think it would be fine to flip this flag, I think they should be ignored by default

PR #373, opened 2022-08-11
[move-core] better ChangeSet impl

vgao1996:
This makes a few improvements to the `ChangeSet`:
1. `Option<Vec<u8>>` gets replaced by `enum Op { New(..), Modify(..), Delete }`. This means the `ChangeSet` now differentiates writing a new entry from modifying an existing one.
2. Some old APIs with ambiguous semantics (e.g. publish_xxx) gets replaced by new ones (add_xxx). The `ChangeSet` now functions more like a delta rather than some sort of ad-hoc storage.

This change benefits tools like move-cli by providing the information they need (new vs modify). It also makes it possible to charge gas based on the number of new entries created.

### DIFF
--- a/language/extensions/async/move-async-vm/tests/sources/AccountStateMachine.exp
+++ b/language/extensions/async/move-async-vm/tests/sources/AccountStateMachine.exp
@@ -5,10 +5,10 @@ publishing vector
 publishing AccountStateMachine
 actor 0x4 created from 0x3::AccountStateMachine
   SUCCESS
-  commit 0x3::AccountStateMachine::AccountStateMachine[0x4] := [00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00]
+  commit 0x3::AccountStateMachine::AccountStateMachine[0x4] := New("[00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00]")
 actor 0x5 created from 0x3::AccountStateMachine
   SUCCESS
-  commit 0x3::AccountStateMachine::AccountStateMachine[0x5] := [00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00]
+  commit 0x3::AccountStateMachine::AccountStateMachine[0x5] := New("[00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00]")
 actor 0x4 handling 0x3::AccountStateMachine::start (hash=0x5438F379BC9E3BCB)
   SUCCESS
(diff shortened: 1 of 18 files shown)


language/extensions/async/move-async-vm/tests/sources/AccountStateMachine.exp
```diff
@@ -5,10 +5,10 @@ publishing vector
 publishing AccountStateMachine
 actor 0x4 created from 0x3::AccountStateMachine
   SUCCESS
-  commit 0x3::AccountStateMachine::AccountStateMachine[0x4] := [00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00]
+  commit 0x3::AccountStateMachine::AccountStateMachine[0x4] := New("[00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00, 00]")
```
vgao1996, line 8: @wrwg let me know if you want me to change the formatting here.

wrwg: Thats fine, it is just kind of debugging output.

language/move-core/types/src/effects.rs
```diff
@@ -7,60 +7,114 @@ use crate::{
...
-                None => return Err(make_err()),
+    use btree_map::Entry::*;
+    use Op::*;
+
+    for (key, op) in other.into_iter() {
+        match map.entry(key) {
+            Occupied(mut entry) => {
+                let r = entry.get_mut();
+                match (r.as_ref(), op) {
+                    (Modify(_) | New(_), New(_)) | (Delete, Delete | Modify(_)) => {
```
wrwg, line 93: I'm wondering if `(New(d1), New(d2)` is really a conflict. You define this such that the right side overrides the left. Why would this then not be resolved as `New(d2)`? Same for other combinations.

vgao1996: It is indeed a conflict. The current semantics of squash is that the resulting `ChangeSet` should be equivalent to applying the two `ChangeSet`s in serial. This means after the first `New`, the slot will be occupied and the second `New` shall fail since it implies no old data exists.

It might be easier to reason about this in terms of version control. You have 3 commits: base, c1 and c2 where c1 and c2 are children of base. Squashing c1 and c2 is actually a two-step process: you first rebase c2 on top of c1 and then squash the contents. In this case, c1 and c2 both modifies the same content so the rebase will fail due to a merge conflict.

vgao1996:
@oxade let me know if the implementation looks good to you. @tnowacki mentioned you guys were fine with this change during the community meeting, at least conceptually.

wrwg (approved):

oxade (approved):
LGTM! Thanks @vgao1996

PR #376, opened 2022-08-12
[vm] Allow to access module metadata

wrwg:
This adds an API to allow the adapter to access module metadata. Metadata can be accessed depending on whether the code is in the cache or not. No guarantees for the availability of metadata are to be expected. Once we have the code loader owned by the adapter, we can remove this API from the VM and let the adapter directly call into the loader.

## Motivation

Better debugging support

Yes

## Test Plan

NA


vgao1996 (approved):
LGTM

movekevin (approved):

PR #338, opened 2022-08-03
[move-analyzer] basic function name completion

0xYao:
## Motivation

The current move-analyzer has limited code completion support, this PR adds primitive types and function name completion.

Note: the `IdentType` in `FunctionIdentTypeMap` is not used in this PR. However, it will be used for function signature help in a future PR.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/move-language/move/blob/main/CONTRIBUTING.md#developer-workflow)?

Yes

## Test Plan

(Share your test plan here. If you changed code, please provide us with clear instructions for verifying that your changes work.)

- Open the extension in debug mode
- There should be a code completion suggestion for primitive types if you start typing a prefix of any primitive types (e.g. u64, bool etc.)
- There should be a code completion suggestion for any function name prefix inside the same file

## PRs to add after the current one

- Add more intellisense support. (e.g. show function signature help, module method completion, auto imports etc.) Each of these is likely going to be a separate PR for easier review.

PR #381, opened 2022-08-13
[move-analyzer] function signature help

0xYao:
## Motivation

This is a follow-up PR from https://github.com/move-language/move/pull/338. Providing signature help should help developers when calling a function inside a file.

Yes.

## Test Plan

- Main test cases have been discussed and covered in `signatureHelp.test.ts`
- Open the extension host in debug mode. You should see a signature help popup when calling a function.
   - There are two key parts of a signature help. The first is the signature label and the second is the position of the active parameter, and you should see the expected signature label and the expected active parameter position when you move the cursor around.

## Demo

https://www.loom.com/share/16e5c241c78a4433bae71046089b73b6


language/move-compiler/src/parser/lexer.rs
```diff
@@ -402,6 +414,34 @@ impl<'input> Lexer<'input> {
         self.token = token;
         self.cur_end = self.cur_start + len
     }
+
+    /// Construct the text info for every token in the lexer by advancing until EOF
```
awelc, line 418: It seems like this is used to parse a single line - is looking for EOF as the end of iteration a correct choice here(@tnowacki)?

language/move-compiler/src/parser/lexer.rs
```diff
@@ -82,6 +82,18 @@ pub enum Tok {
     AtSign,
 }
 
+#[derive(Copy, Clone, Debug, Default)]
+pub struct TextRange {
+    pub start: usize,
+    pub end: usize,
+}
+
+#[derive(Copy, Clone, Debug, Default)]
```
awelc, line 91: Please add documentation to these definitions.

language/move-analyzer/src/symbols.rs
```diff
@@ -2190,6 +2198,228 @@ pub fn on_hover_request(context: &Context, request: &Request, symbols: &Symbols)
     );
 }
 
+/**
+ * Returns a tuple where the first value is the index of the text info of the function name and the second value is the text info of the function
```
awelc, line 2202: Please add explanation of "text info" for the function describes (function name and position/range of the name?)

language/move-analyzer/src/symbols.rs
```diff
@@ -2190,6 +2198,228 @@ pub fn on_hover_request(context: &Context, request: &Request, symbols: &Symbols)
...
+            }
+            _ => {}
+        }
+    }
+
+    None
+}
+
+/**
+ * Find the position of the active parameter given a vector of parsed tokens
```
awelc, line 2241: Please add more explanation of what parsed tokens contains (what kind of line was parsed to obtain the tokens).

language/move-analyzer/src/symbols.rs
```diff
@@ -2190,6 +2198,228 @@ pub fn on_hover_request(context: &Context, request: &Request, symbols: &Symbols)
...
+                parenthesis_delta -= 1;
+            }
+            _ => {}
+        }
+    }
+
+    active_parameter
+}
+
+fn collect_parameter_info_from_ident(ident: &IdentType) -> Vec<ParameterInformation> {
```
awelc, line 2281: Please dd documentation for the function.

awelc (changes requested):
This is a relatively complicated addition and I would really appreciate it if you could add a higher level description of how the algorithm works. I have a high-level idea here but it's difficult to piece together all the details from the code itself - a textual description of the algorithm (what is parsed and when and how the parsed information is used) would go a long way (it does not have to be super long).

Also, I have build and and installed both the language server and the extension from this branch, but things to do not work for me quite the same as in the demo. In particular, while I am getting a match (vi text highlighting) between parameter name (being typed) and parameter type (in the hover box), I do not get the parameter name auto-completion (see attached screenshot). I am using VSCode 1.70.1 and all tests pass with this version (after I modified runTests.rs to include it).

image

awelc:
I have also discovered a problem with the previous PR (https://github.com/move-language/move/pull/338) underlying this one. I added a comment there even though the PR is already merged: https://github.com/move-language/move/pull/338/files#r948387534

Please provide a fix first along with appropriate test before the current PR is merged.

0xYao:
> I have addressed this in #389 (please note that I removed the `IdentType` for functions is now available via `Symbols::file_mods`.

Sounds good, let me push the fixes using the latest changes.

0xYao:
> I do not get the parameter name auto-completion

Sorry, what do you mean by parameter name auto-completion? Is this addressed in your #389 PR?

0xYao:
Given the cursor position of the current line, I need to find out the function name and the corresponding module that it belongs. It seems like I can reuse some of the logic from `Symbolicator::mod_call_symbols`, but maybe it's better to do custom string parsing and then extract the correct information from `file_mods`?

awelc:
> Given the cursor position of the current line, I need to find out the function name and the corresponding module that it belongs.

I thought that you could simply use `Symbols::file_mods`. In most cases there will be one module per file, and even if there isn't, iterating over the list of modules in the set shouldn't be a huge deal (in terms of speed and complexity).  Unless I misunderstand what information you need to obtain and how.

0xYao:
> I thought that you could simply use `Symbols::file_mods`. In most cases there will be one module per file, and even if there isn't, iterating over the list of modules in the set shouldn't be a huge deal (in terms of speed and complexity). Unless I misunderstand what information you need to obtain and how.

Additionally, I also need to find the current enclosing scope given a cursor position, and determine which module the developer is calling the function from.

awelc:
> Additionally, I also need to find the current enclosing scope given a cursor position, and determine which module the developer is calling the function from.

Good point. Here are some thoughts. 

First off, it may make sense to special-case implementation for one-module-per-file - we don't have to worry about multiple modules in this case (which should be the common case).

In terms of figuring out module scope (in terms of file lines), when constructing `ModuleDefs`, we iterate over typed AST unique map that gives us location of modules (I think) in `pos`:
https://github.com/move-language/move/blob/70b34a66473c34ad30d101290b249f2db3c847a2/language/move-analyzer/src/symbols.rs#L708

We currently only store the starting position of the module but this should give us the ending position as well, which we could use to figure out module's scope within a file. To make it fully robust we'd probably have to make an adjustment for a new line (or lines) being inserted that would change start/end of modules in the file, but I am not sure if this is feasible at the moment. 

If this does not work for some reason, then I believe a function call can only be inserted inside another function. If we then remembered both starting line of function definition (we already have this in `FunctionDef`) and the ending line (we could conceivably obtain it from the outermost code block in function's body) we could use `Symbol::file_mods` to iterate over all functions in a given file and figure out if the cursor is located inside one of them. This is potentially more fragile than the previous suggestion, but I haven't checked if module start/end locations are indeed faithfully represented in the AST.

language/move-analyzer/src/symbols.rs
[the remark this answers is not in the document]
0xYao: done

language/move-analyzer/src/symbols.rs
```diff
@@ -2190,6 +2198,228 @@ pub fn on_hover_request(context: &Context, request: &Request, symbols: &Symbols)
     );
 }
 
+/**
+ * Returns a tuple where the first value is the index of the text info of the function name and the second value is the text info of the function
+ */
+fn find_closet_fn_in_text_info_vec<'a>(
```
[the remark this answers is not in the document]
0xYao: fixed

language/move-analyzer/src/symbols.rs
```diff
@@ -2190,6 +2198,228 @@ pub fn on_hover_request(context: &Context, request: &Request, symbols: &Symbols)
...
+    // Plus 1 to include the character at the starting index
+    for i in (0..start_idx + 1).rev() {
+        let text_info = text_info_vec.get(i).unwrap();
+
+        match text_info.text {
+            "(" => {
+                if parenthesis_delta == 0 {
+                    // The identifier to the left of the left parenthesis is the function.
+                    match text_info_vec.get(i - 1) {
+                        Some(text_info) => {
```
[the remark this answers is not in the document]
0xYao: fixed

language/move-compiler/src/parser/lexer.rs
0xYao: @tnowacki a friendly reminder for this^

language/move-compiler/src/parser/lexer.rs
0xYao: done

language/move-analyzer/src/symbols.rs
0xYao: done

language/move-analyzer/src/symbols.rs
0xYao: done

0xYao:
> First off, it may make sense to special-case implementation for one-module-per-file - we don't have to worry about multiple modules in this case (which should be the common case).

I am going with the case where we have one-module-per-file for now. I have added a note in the `on_signature_help_request` function in case we want to address this in the future.

awelc:
> I am going with the case where we have one-module-per-file for now. I have added a note in the `on_signature_help_request` function in case we want to address this in the future.

What are you going to to in the short term then if we have two modules in a file and each module defines a function with the same name? Are you going to list them both in the function signature help?

0xYao:
> What are you going to to in the short term then if we have two modules in a file and each module defines a function with the same name? Are you going to list them both in the function signature help?

The current implementation just returns the signature help of the function in the first module with that function name that we encounter in `file_mods`. However, if this behaviour is not acceptable, I can update the implementation to return the signature help of the function inside the enclosing module only in the short-term.

But feel free to propose an alternative solution.

0xYao:
@awelc can you please review?

language/move-analyzer/src/symbols.rs
```diff
@@ -2206,6 +2210,288 @@ pub fn on_hover_request(context: &Context, request: &Request, symbols: &Symbols)
     );
 }
 
+/// Returns (index of the function, text info of the function name)
+///
+/// The text info result includes the name and the range of the function in a given line.
+///
+/// A non-zero parenthesis_delta value means the current index is in a nested function scope.
```
awelc, line 2217: Please move the comments concerning algorithmic details for all functions (which are very useful, by the way!) to the beginning of the function itself, leaving the documentation comments more succinct.

language/move-analyzer/src/symbols.rs
```diff
@@ -2206,6 +2210,288 @@ pub fn on_hover_request(context: &Context, request: &Request, symbols: &Symbols)
...
+/// For example, consider `fun_1(a, fun_2(b, c, d), e)` and start_idx is at "e", which means
+/// we want to find out which function "e" belongs to. Since we are interating backwards from
```
awelc, line 2219:
```suggestion
/// we want to find out which function "e" belongs to. Since we are iterating backwards from
```

awelc (changes requested):
Thank you for the updates! The newly added comments helped a lot in understanding what's going on!

I still have a few questions/concerns:
- as I indicated before, I cannot reproduce the behavior shown in the video clip attached to the PR - the auto-completion for parameter names does not work for me (is there a special setting that I need to toggle?)
- I wonder if the current version of the algorithm works if a function signature is spread across different lines, and also if the developer wants to put arguments on different lines; I would say that we really need the former while the latter would be nice to have but could be delegated to future work
- I would much prefer to have the multi-module-in-one-file case handled by providing multiple functions in the signature help if multiple of them are a match (e.g. have sufficient number of parameters); as it is now, the risk is that only the "wrong" function signature will be provided which would be quite confusing to the developer

PR #383, opened 2022-08-14
[gas] new gas algebra impl

vgao1996:
This adds a gas algebra module that offers primitives for saturating addition, checked subtraction, type-safe multiplication and unit conversion.

More specifically, this
1. Defines some uninhabitable types representing various units of measurement 
2. Defines a `struct GasQuantity<U>`, which is a wrapper around `u64` but is parametric over the unit types. Following the discussion we had during the community meeting, this type
  a. Always uses saturating addition and will panic if overflow
  b. Only permits saturating multiplication when the units match (e.g. 3 gas units/byte * 5 bytes = 15 gas units)
  c. Only allows checked_sub and does not implement the `Add` trait. 
  d. Does not support division directly, but allows the client to define unit conversion.


language/move-vm/test-utils/src/gas_schedule.rs
```diff
@@ -17,88 +17,39 @@ use move_binary_format::{
...
-    pub fn to_internal_units(&self, units: GasUnits<GasCarrier>) -> InternalGasUnits<GasCarrier> {
-        InternalGasUnits::new(units.get() * self.gas_unit_scaling_factor)
-    }
-
-    pub fn to_external_units(&self, units: InternalGasUnits<GasCarrier>) -> GasUnits<GasCarrier> {
-        GasUnits::new(units.get() / self.gas_unit_scaling_factor)
-    }
+impl ToUnitFractional<GasUnit> for InternalGasUnit {
+    const NOMINATOR: u64 = 1;
+    const DENOMINATOR: u64 = 1000;
```
vgao1996, line 44: A note on `GasStatus`: the existence of two gas units (internal and external) is a legacy of the old days. My plan is to replace it with a test-only gas meter impl that doesn't have this differentiation, which we can use in the Move repo. This gas meter will likely NOT be suitable for production use.

One thing that's a bit tricky to deal with is the Move CLI,  since it still uses `GasStatus` internally and expects clients to pass in an old `CostTable`. IMO we'll need to redesign this part and see if we can come up with a generic approach that works for everyone, perhaps by asking clients to pass in their own gas meter implementation.

I know the repeated simplification and eventual removal of `GasStatus` will be a problem for clients who are currently using it in prod (e.g. @jolestar) or planning to do so , but IMO the correct thing to do really is to implement your own gas meter. In the worst case, you can fork the old `GasStatus` and try to make it work.

jolestar: Do you have a plan to eliminate the `gas_unit_scaling_factor` and `InternalGasUnits`? 

I think the `gas_unit_scaling_factor` is designed for Diem's Currency that does not have enough decimals and introduce the internal gas units to fix the bug https://github.com/diem/diem/issues/7238. 

We set the `gas_unit_scaling_factor` to 1 on Starcoin, so InternalGasUnits == GasUnits.

I support eliminating the `gas_unit_scaling_factor` and `InternalGasUnits`.

vgao1996: > Do you have a plan to eliminate the gas_unit_scaling_factor and InternalGasUnits?

I would say the plan is to remove `GasStatus` so we won't have an official gas meter implementation inside the Move repo any more. Instead, with the changes I recently made, each blockchain can and should implement its own gas meter, in a way that suits its own needs. Whether to have the `gas_unit_scaling_factor` and two gas units is now a design choice you can make on your own. Again, I highly encourage you to replace `GasStatus` with your own gas meter soon.

The new gas meter implementation I mentioned was only meant for testing within the Move repo, where there is no need to calculate any sort of transaction fee, and thus no need to have two gas units to deal with precision issues.

With that being said I think I may end up renaming `InternalGasUnits` to something like `VMGasUnits`. The name `InternalGasUnits` seems to imply there exists some external gas units, which isn't necessarily true. In the mid to long term, it's possible that the concept of `InternalGasUnits` may disappear from the Move repo, but only after I'm done with the VM ownership refactoring and if my plan to further generify gas metering works out.

> I think the gas_unit_scaling_factor is designed for Diem's Currency that does not have enough decimals and introduce the internal gas units to fix the bug https://github.com/diem/diem/issues/7238.

As Tim pointed out, one advantage of having a scaled gas unit is to allow one to "change the scaling factor as time goes on as, e.g., the instructions become less expensive computationally, whereas once a given precision is picked for a currency that can't really ever be changed. And in general, the scaling factor gives us a knob to turn to scale the transaction fees in the system as time goes on". 

We at Aptos also considered whether we should unify the two gas units and eventually decided to keep them, for the reason Tim brought up, but I guess you could live without scaling if you have enough decimals.

vgao1996: Discussed offline.

wrwg (approved):

jolestar (approved):

vgao1996:
@oxade Let me know if this looks good to you!

PR #385, opened 2022-08-16
[book] modify example for freeze of reference

0x6368616e67:
## Motivation

In the  Move Book 
Chapter References [freeze inference](https://move-language.github.io/move/references.html#freeze-inference).
here may want to say the compile will automatic add a `freeze` action between 
`&x ` and `&mut x`
so change the former

`let y: &mut u64 = &mut x;`
to 
`let y: &u64 = &mut x;`


tnowacki (approved):
Thanks for the fix!

PR #387, opened 2022-08-16
[gas] add gas meter callbacks

vgao1996:
This extends the `GasMeter` trait with callbacks for various operations, allowing precise control. As a byproduct, this also adds two traits `TypeView` and `ValueView` which are used to expose certain info to the gas meter in a lazy fashion.

The changes are (hopefully) implemented in a backward-compatible manner, meaning that one should be able to maintain the old gas semantics if implementing the callbacks in a certain way. (See the current `GasStatus` impl for an example.)

**Note: this depends on https://github.com/move-language/move/pull/383, and please be aware that the first commit belongs to that PR.**

### DIFF
--- a/language/move-vm/types/src/gas.rs
+++ b/language/move-vm/types/src/gas.rs
@@ -1,27 +1,199 @@
 // Copyright (c) The Move Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use move_binary_format::{errors::PartialVMResult, file_format_common::Opcodes};
-use move_core_types::gas_algebra::{AbstractMemorySize, InternalGas};
+use crate::views::{TypeView, ValueView};
+use move_binary_format::errors::PartialVMResult;
+use move_core_types::{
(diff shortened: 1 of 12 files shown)


wrwg (approved):

language/move-vm/types/src/gas.rs
```diff
@@ -1,47 +1,369 @@
...
+
+    // TODO(Gas): Expose the elements
+    fn charge_vec_unpack(
+        &mut self,
+        ty: impl TypeView,
+        expect_num_elements: NumArgs,
+    ) -> PartialVMResult<()>;
+
+    // TODO(Gas): Expose the two elements
+    fn charge_vec_swap(&mut self, ty: impl TypeView) -> PartialVMResult<()>;
```
oxade, line 179: Can you make an issue for these TODOs and others?

vgao1996: Will do!

PR #388, opened 2022-08-18
[gas] minor fixes and improvements to gas meter callbacks, views and move-stdlib

vgao1996:
This packs a few fixes and improves to various components:
1. `GasMeter::charge_read_ref` now correctly receives a view into the value being referenced instead of the reference value itself
2. `ValueView` now provides a `visit` function that would allow you to traverse the value recursively. 
  a. This is used to provide a new implementation of abstract memory size calculation
  b. The old `size()` functions are still kept, but they are used only in tests to check that the results produced by both implementations match.
  c. The plan is to (1) remove the `size()` functions once the new impl gets used in the wild for some time, (2) encourage clients to implement their own abstract memory size if they still need it and (3) remove the view-based abstract memory size impl
3. Some move-stdlib gas parameters now have shorter and more explicit names

### DIFF
--- a/language/move-vm/types/src/views.rs
+++ b/language/move-vm/types/src/views.rs
@@ -1,7 +1,10 @@
 // Copyright (c) The Move Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use move_core_types::{gas_algebra::AbstractMemorySize, language_storage::TypeTag};
+use move_core_types::{
+    account_address::AccountAddress, gas_algebra::AbstractMemorySize, language_storage::TypeTag,
+};
+use std::mem::size_of;
 
 /// Trait that provides an abstract view into a Move type.
 ///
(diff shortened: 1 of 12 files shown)


vgao1996:
@tnowacki the sub-points of (2) are what I was referring to as the "phased elimination of abstract memory size" during the meeting

wrwg (approved):

language/move-vm/types/src/views.rs
```diff
@@ -17,11 +20,131 @@ pub trait TypeView {
...
+
     /// Returns the abstract memory size of the value.
     ///
     /// The concept of abstract memory size is not well-defined and is only kept for backward compatibility.
     /// New applications should avoid using this.
-    fn legacy_abstract_memory_size(&self) -> AbstractMemorySize;
+    ///
+    /// TODO(Gas): Encourage clients to replicate this in their own repo and get this removed once
+    ///            they are done.
+    fn legacy_abstract_memory_size(&self) -> AbstractMemorySize {
```
vgao1996, line 32: @jolestar I would recommend you replicate this in the Starcoin repo. Better, you might also want to revise the formula a bit since some of the numbers here picked for no obvious reasons. It might also be a good idea to store some of the parameters on-chain so you can adjust them dynamically.

jolestar: cc @LemonHX

vgao1996: cc @oxade

PR #391, opened 2022-08-18
[Table] use u256 instead of u128 for table handle to increase security

zekun000:
u128 is not enough to avoid collision, increase the handle to u256 represented by two u128. the layout is compatible to [u8;32] in bcs.

### DIFF
--- a/language/extensions/move-table-extension/src/lib.rs
+++ b/language/extensions/move-table-extension/src/lib.rs
@@ -21,7 +21,7 @@ use move_vm_types::{
     loaded_data::runtime_types::Type,
     natives::function::NativeResult,
     pop_arg,
-    values::{GlobalValue, Reference, StructRef, Value},
+    values::{GlobalValue, Reference, Struct, StructRef, Value},
 };
 use sha3::{Digest, Sha3_256};
 use smallvec::smallvec;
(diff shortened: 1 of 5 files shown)


davidiw (approved):

igor-aptos (approved):

language/extensions/move-table-extension/src/lib.rs
```diff
@@ -40,7 +40,7 @@ use std::{
 /// hash over a transaction hash provided by the environment and a table creation counter
 /// local to the transaction.
 #[derive(Copy, Clone, Debug, PartialOrd, Ord, PartialEq, Eq)]
-pub struct TableHandle(pub u128);
+pub struct TableHandle(pub u128, pub u128);
```
vgao1996, line 43: Can you also make this `{ pub low: u128, pub high: u128 }`? Otherwise it seems risky

wrwg: Isn't the memory repr just the opposite (high-low because of big endian).

language/extensions/move-table-extension/src/lib.rs
```diff
@@ -381,25 +385,29 @@ fn native_new_table_handle(
     let mut table_data = table_context.table_data.borrow_mut();
 
     // Take the transaction hash provided by the environment, combine it with the # of tables
-    // produced so far, sha256 this and select 16 bytes from the result. Given the txn hash
+    // produced so far, sha256 this to produce a unique handle. Given the txn hash
     // is unique, this should create a unique and deterministic global id.
```
alinush, line 389: One thing that might be worth thinking about: The assumption that TXN hashes are unique might **not** always hold. For example, in Bitcoin, different coinbase TXNs for different blocks could be identical and therefore have the same hash.

language/extensions/move-table-extension/src/lib.rs
vgao1996: Ah I didn't think twice before writing the stuff down. Was just trying to point out that I would love to see them being explicitly named. Please check the ordering before committing this!

alinush: Yes: naming them would be very useful to prevent developers accidentally flipping them.

zekun000: it's little endian (to comply with bcs)

PR #392, opened 2022-08-19
[table] use address for handle instead of two u128

zekun000:
Use u128 imposes huge usability problem for clients since the json returns back the Object {low: xxx, high: yyy} for the handle instead of 32 bytes.
Use address can mitigate the problem until we have u256 impl.


davidiw (approved):
considering that most would choose to use the address at a certain size due to security reasons, this makes sense. I strongly support this PR.

PR #396, opened 2022-08-19
[table] add proper gas metering and fix return value bug

vgao1996:
This adds proper gas metering to the table extension and fixes a bug with the return values: `Ok(NativeResult::err)` instead of `Err(..)` should be used in case of user errors.

### DIFF
--- a/language/extensions/move-table-extension/src/lib.rs
+++ b/language/extensions/move-table-extension/src/lib.rs
@@ -10,8 +10,12 @@
 use better_any::{Tid, TidAble};
 use move_binary_format::errors::{PartialVMError, PartialVMResult};
 use move_core_types::{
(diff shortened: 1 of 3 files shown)


language/extensions/move-table-extension/src/lib.rs
```diff
@@ -218,157 +220,129 @@ impl TableData {
...
+                };
+                e.insert(table)
+            }
+            Entry::Occupied(e) => e.into_mut(),
+        })
     }
 }
 
 impl Table {
-    /// Inserts a value into a table.
```
vgao1996, line 237: @wrwg I had to delete all these helpers since they interacted poorly with the gas metering code. Found it much easier to just move the logic into the native functions instead of passing the gas parameters around.

wrwg (approved):

Issue #393, opened 2022-08-19
[Feature Request]Support more helper functions in fixed_point32


leofisG:
# 🚀 Feature Request

It will be nice if more utility functions can be added to the `fixed_point32` module in the `stdlib`. Namely:
- round
- min
- max
- floor
- ceil

## Motivation

**Is your feature request related to a problem? Please describe.**

## Pitch

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))
Yes

leofisG:
would love to work on this @sblackshear

PR #397, opened 2022-08-21
[stdlib] Added more helper function for fixed_point32

leofisG:
## Motivation

I added more helper functions to `fixed_point32`, namely:
- min
- max
- create_from_u64
- floor
- ceil
- round

Addressed: #393 

Yes.

## Test Plan

- 100% Test coverage
- Move spec added

### DIFF
--- a/language/move-stdlib/sources/fixed_point32.move
+++ b/language/move-stdlib/sources/fixed_point32.move
@@ -154,6 +154,67 @@ module std::fixed_point32 {
         num.value == 0
     }
 
+    /// Returns the smaller of the two FixedPoint32 numbers.
+    public fun min(num1: FixedPoint32, num2: FixedPoint32): FixedPoint32 {
+        if (num1.value < num2.value) {
+            num1
+        } else {
+            num2
+        }
+    }
+    spec min {
+        pragma opaque;
+        aborts_if false;
+        ensures result == spec_min(num1, num2);
+    }
+    spec fun spec_min(num1: FixedPoint32, num2: FixedPoint32): FixedPoint32 {
+        if (num1.value < num2.value) {
+            num1
+        } else {
+            num2
+        }
+    }
+
+    /// Returns the larger of the two FixedPoint32 numbers.
+    public fun max(num1: FixedPoint32, num2: FixedPoint32): FixedPoint32 {
+        if (num1.value > num2.value) {
+            num1
+        } else {
+            num2
+        }
+    }
+    spec max {
+        pragma opaque;
+        aborts_if false;
+        ensures result == spec_max(num1, num2);
+    }
+    spec fun spec_max(num1: FixedPoint32, num2: FixedPoint32): FixedPoint32 {
+        if (num1.value > num2.value) {
+            num1
+        } else {
+            num2
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


language/move-stdlib/sources/fixed_point32.move
```diff
@@ -154,6 +154,136 @@ module std::fixed_point32 {
...
+            val.value >> 32
+        } else {
+            (val.value - fractional + one) >> 32
+        }
+    }
+
+    /// Returns the value of a FixedPoint32 to the nearest integer.
+    public fun round(num: FixedPoint32): u64 {
+        let floored_num = floor(num) << 32;
+        let bounary = floored_num + ((1 << 32) / 2);
```
chen-robert, line 264: `bounary` should be spelled `boundary`

language/move-stdlib/sources/fixed_point32.move
```diff
@@ -154,6 +154,136 @@ module std::fixed_point32 {
         num.value == 0
     }
 
+    /// Returns the smaller of the two FixedPoint32 numbers.
+    public fun min(num1: FixedPoint32, num2: FixedPoint32): FixedPoint32 {
```
aditya-solana, line 158: If the values are equal then this will return num2. This deviates from [Rust's](https://doc.rust-lang.org/std/cmp/fn.min.html
) and [C++'s](https://en.cppreference.com/w/cpp/algorithm/min) definition of min.

leofisG: @emmazzz @junkil-park does this matter to you? I can switch the two, although the content will be the same anyway.

emmazzz: Doesn't matter to me.

vgao1996: Yeah this shouldn't matter as long as the contents are the same.

language/move-stdlib/sources/fixed_point32.move
```diff
@@ -154,6 +154,136 @@ module std::fixed_point32 {
...
+    spec fun spec_max(num1: FixedPoint32, num2: FixedPoint32): FixedPoint32 {
+        if (num1.value > num2.value) {
+            num1
+        } else {
+            num2
+        }
+    }
+
+    /// Create a fixedpoint value from a u64 value.
+    public fun create_from_u64(val: u64): FixedPoint32 {
```
vgao1996, line 200: How about renaming to "from_u64" for simplicity?

vgao1996 (commented):
Overall looks good to me and thanks for the high quality work!

@wrwg @tnowacki any concerns about changing the stdlib at this time?

language/move-stdlib/sources/fixed_point32.move
leofisG: This is the name that I originally used, however it seems the current name is more consistent with “ create_from_raw_value” and “create_from_rational” that we already have.

vgao1996: I see. Then it's probably better to just follow the existing convention.

vgao1996 (approved):

PR #403, opened 2022-08-23
[move-book-zh] update global storage structure

geometryolife:
## Motivation

Update the contents of the global storage structure in move-book-zh and fix a small typo in book.


geometryolife:
Already done. @sblackshear

sblackshear (approved):

PR #404, opened 2022-08-23
[book and move-book-zh] update modules and scripts

geometryolife:
## Motivation

- Update contents in book
- Uupdate contents in move-book-zh


geometryolife:
Already done. @sblackshear

sblackshear (approved):

PR #406, opened 2022-08-23
[move-docgen] Making Docgen independent of Prover

villesundell:
Docgen was originally developed as a part of the Prover. However, Docgen is now independent of Prover, so moving it from Prover to its own directory seems the next logical step.

## Motivation

Docgen was buried inside Prover for no apparent reason (now).

Yes.

## Test Plan

Unit tests pass: moved Docgen unit tests, tests for `move docgen`, and all the previous tests.


villesundell:
Pinging my mentors @awelc and @tnowacki for review :memo:

wrwg (commented):
This is not really making docgen independent of the prover, because it still can only be run via the main prover driver (as [here](https://github.com/move-language/move/blob/f29fbd1e0bd8426dba7132385d88a5318a2c128b/language/move-prover/src/lib.rs#L94)). The only thing you are doing is moving the crate around, now creating a dependency from the move-prover to `../docgen` instead of `docgen`.

Lets not do this until we have a really cleanup. This would first tear out docgen out of provers `lib.rs` and create its own driver, and then move the crate.

PS. There are historical reasons why the documentation generator (and abigen and errmapen tools) are in the prover directory. The reason is that they heavily depend on move-model, which was originally the internal object model for the prover. The move-model was moved out of the prover a while ago, but those tools are still there. Moving them out requires refactoring of `move-prover/src/lib.rs`.

villesundell:
@wrwg Sorry, I am not following here: if I understand correctly Docgen is nowadays free from Prover dependencies ([docgen.rs](https://github.com/move-language/move/blob/28082f5fc4b1899947a2403f58cb9ffed9a969b7/language/move-prover/move-docgen/src/docgen.rs)) and can be now (#327) used directly with `move docgen` without calling/using Prover?

I even experimented removing the last dev-dependencies in [this commit](https://github.com/villesundell/move/pull/10/commits/a756babb5f1064494ab69a601e7daf027b820d2e) and `move docgen` compiles and works correctly. However test is still using Prover, and that's something I have to change, thank you for the pointer :+1: 

It's true that Prover is still using Docgen, and hence is depending on it, but to me that seems like another tasks to remove Prover's dependence on Docgen. So to me it looks like Docgen is independent of Prover (and hence needs to be moved), and Prover is not independent of Docgen (but must be in the future).

But changing this to WIP at least because of the test refactoring (and for possible refactoring in general).

wrwg:
> @wrwg Sorry, I am not following here: if I understand correctly Docgen is nowadays free from Prover dependencies ([docgen.rs](https://github.com/move-language/move/blob/28082f5fc4b1899947a2403f58cb9ffed9a969b7/language/move-prover/move-docgen/src/docgen.rs)) and can be now (#327) used directly with `move docgen` without calling/using Prover?
> 
> I even experimented removing the last dev-dependencies in [this commit](https://github.com/villesundell/move/pull/10/commits/a756babb5f1064494ab69a601e7daf027b820d2e) and `move docgen` compiles and works correctly. However test is still using Prover, and that's something I have to change, thank you for the pointer 👍
> 
> It's true that Prover is still using Docgen, and hence is depending on it, but to me that seems like another tasks to remove Prover's dependence on Docgen. So to me it looks like Docgen is independent of Prover (and hence needs to be moved), and Prover is not independent of Docgen (but must be in the future).
> 
> But changing this to WIP at least because of the test refactoring (and for possible refactoring in general).

The prover has never been using docgen. Instead the only way to call docgen was over the prover driver (this is the same for some other tools). It is great that you point  out that you made now a direct entrypoint via the Move CLI. Still can you please cut off the remaining, now unnecessary dependency? It's the one and only one AFAICT. If the docgen tests are still depending on the entry over the prover, it shouldn't be too hard to fix this as well. I'm a bit concerned about saying "next PR", because those things can be easily forgotten. It's a similar concern I had with the docgen options, where we have now some kind of duplication. 

BTW for the background, docgen is a tool I've written, so I've some stake here ;-)

Thanks for trying to clean this up.

wrwg:
Ah and one additional remark. We where actually trying to locate all tools like `docgen` in the `tools` folder. Tools which are currently outside are legacy. Perhaps you can move docgen into that sub-directory. Thank you.

villesundell:
@wrwg Ah, okay! So, I see these two steps next for me (probably one commit each):
  * Remove Prover dependency in test
  * Remove Docgen code from Prover

(Current Docgen does not seem to be stand-alone, so probably it's not mature enough for `tools` just yet?)

wrwg:
> (Current Docgen does not seem to be stand-alone, so probably it's not mature enough for `tools` just yet?)

Not sure what you mean -- definitely, if it would not be mature, it would even more belong into tools instead of top-level. In general, we do not have such categories influencing the location of a crate. But there was a decision made to move all tools into the tools directory a while ago.

villesundell:
@wrwg Ah, sorry: at quick glance it seemed to me that `tools` had only the tools which are stand-alone, usable directly from the command line. But now I see that there are some (like `move-package`) which are not stand-alone. So yeah, I agree, maybe `tools` is the best place for `move-docgen` :+1:

PR #407, opened 2022-08-24
[book and move-book-zh] update move coding conventions

geometryolife:
## Motivation

Update the contents of move coding conventions in move-book and move-book-zh.


geometryolife:
Already done. @sblackshear

sblackshear (approved):

PR #408, opened 2022-08-24
feat(stdlib): encode and decode base16

zouguangxian:
## Motivation

provide a human-friendly representation of binary-coded values.

Yes

## Test Plan

`move test` passed in language/move-stdlib/nursery/sources directory.


sblackshear:
I am personally in favor of having more of these utility encoding/decoding libraries inside the Move stdlib, but we need to think carefully about how we want to organize them--one module per format (as this suggests?). All in one big `bytes` module? Something else?

jolestar:
I support `one module per format`

sblackshear (changes requested):
This looks good to me at a high level, and I would like to see it go in. But this needs a lot more tests for both the expected behavior and the abort cases.

sblackshear:
Another approach we could consider here (that would make the testing burden a bit less onerous) is to implement this via `native` functions that wrap the Rust `hex` library.

jolestar:
> Another approach we could consider here (that would make the testing burden a bit less onerous) is to implement this via `native` functions that wrap the Rust `hex` library.

I also think it is possible to let the User Module implement a native function with a WASM runtime in the future. This approach lets adding new native functions more gracefully and does not need a hard fork upgrade.

PR #409, opened 2022-08-24
[move-book and move-book-zh] update signer

geometryolife:
## Motivation

Update contents of signer in move-book and move-book-zh.


geometryolife:
Already done. @sblackshear

sblackshear (approved):

PR #411, opened 2022-08-25
[Move Prover] Support constant vector of addresses

junkil-park:
- Extended the move-model with `AddressArray` to support constant vectors of addresses
- They can be defined in `const A:vector<address> = [...]` at the module level and in `let A = vector<address>[...]` in a function body. This commit supports both cases.
- Added some test cases

## Motivation

To support constant vector of address

Yes

## Test Plan

cargo test

### DIFF
--- a/language/move-model/src/builder/exp_translator.rs
+++ b/language/move-model/src/builder/exp_translator.rs
@@ -1905,31 +1905,53 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
         result
     }
 
-    pub fn translate_from_move_value(&self, loc: &Loc, value: &MoveValue) -> Value {
-        match value {
-            MoveValue::U8(n) => Value::Number(BigInt::from_u8(*n).unwrap()),
-            MoveValue::U64(n) => Value::Number(BigInt::from_u64(*n).unwrap()),
-            MoveValue::U128(n) => Value::Number(BigInt::from_u128(*n).unwrap()),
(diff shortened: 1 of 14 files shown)


language/move-model/src/builder/exp_translator.rs
```diff
@@ -2,6 +2,7 @@
 // Copyright (c) The Move Contributors
 // SPDX-License-Identifier: Apache-2.0
 
+use std::any::Any;
```
meng-xu-cs, line 5: Merge with the import below

meng-xu-cs (commented):
Is there a particular reason why we want to special case `vector<address>` over other vector constants? (e.g., `vector<u64>`, or even `vector<vector<address>>`)?

I can understand `vector<u8>` as we may apply some special theories for strings. But more special casing sounds a bit too ad-hoc to me.

Maybe consider
```diff
pub enum Value {
    Address(BigUint),
    Number(BigInt),
    Bool(bool),
    ByteArray(Vec<u8>),
+   Vector(Vec<Value>),
}
```

wrwg:
Thanks for looking at this! I agree we should not special case more vectors. We should generalize to `Vector(Vec<Value>)` as Meng is saying, otherwise we hit the problem again with `vector[10,11,12]` or `vector[true, false]` etc. Not sure why we introduced `ByteVector` in the first place, probably because of a misunderstanding what the Move compiler generates here.

meng-xu-cs (approved):
Unblocking the landing, will generalize it later.

PR #413, opened 2022-08-25
Enable debug on debug builds

tnowacki:
- Enable debug printing on Rust debug builds

## Motivation

- As discussed this morning, this might be a reasonable way of letting debug prints occur on local deployments even when not in "test" mode
- Not sure if it fully helps but feels like a start

### DIFF
--- a/language/move-stdlib/src/natives/debug.rs
+++ b/language/move-stdlib/src/natives/debug.rs
@@ -35,9 +35,8 @@ fn native_print(
     debug_assert!(ty_args.len() == 1);
     debug_assert!(args.len() == 1);
 
-    // No-op if the feature flag is not present.
-    #[cfg(feature = "testing")]
-    {
+    // No-op if the feature flag is not present. or if in Rust release build
+    if cfg!(feature = "testing") || cfg!(debug_assert) {
         let _ty = ty_args.pop().unwrap();
         let r = pop_arg!(args, Reference);
 
@@ -77,8 +76,7 @@ fn native_print_stack_trace(
     debug_assert!(ty_args.is_empty());
     debug_assert!(args.is_empty());
 
-    #[cfg(feature = "testing")]
-    {
+    if cfg!(feature = "testing") || cfg!(debug_assert) {
         let mut s = String::new();
         context.print_stack_trace(&mut s)?;
         println!("{}", s);


language/move-stdlib/src/natives/debug.rs
```diff
@@ -35,9 +35,8 @@ fn native_print(
     debug_assert!(ty_args.len() == 1);
     debug_assert!(args.len() == 1);
 
-    // No-op if the feature flag is not present.
-    #[cfg(feature = "testing")]
-    {
+    // No-op if the feature flag is not present. or if in Rust release build
```
sblackshear, line 38: If folks get the `move` or `sui` binaries from `cargo install` (recommended/most common install path), they will get Rust release binaries. If they then use these binaries in their local development/debugging flows, won't they get confused when `debug` printing doesn't work?

awelc (approved):
LGTM. This will allow local printing in local deployments spun using a debug build of a given chain, but will become a no-op in actual release builds (dev/test/main net) while still working in test.

sblackshear (changes requested):
See concern about devs using release builds

language/move-stdlib/src/natives/debug.rs
awelc: > If folks get the `move` or `sui` binaries from `cargo install` (recommended/most common install path), they will get Rust release binaries. If they then use these binaries in their local development/debugging flows, won't they get confused when `debug` printing is a no-op?

We could conceivably include `--debug` flag by default with `cargo install` instructions and make a note that if someone wants faster local builds they can omit it but then they will loose the ability to debug-print outside of test code.

sblackshear: Said differently: this means debug printing will only work if you build `move`/`sui` yourself from source, in debug mode. But this is not the flow we recommend/expect.

sblackshear: > We could conceivably include --debug flag by default with cargo install instructions and make a note that if someone wants faster local builds they can omit it but then they will loose the ability to debug-print outside of test code.

I don't asking folks to re-install the `sui`/`move` binaries differently depending on whether they want to use debug printing/not (and asking them to use a slower binary when they want debug printing) is a good experience. Ideally, they should need to install one binary per version and it should just work for everything.

tnowacki: Do you think making another feature flag would work?
We could make the CLI binaries have this flag on? (not sure if there is a way to do this).
I just want to make sure we don't accidentally turn this on for node software

sblackshear: What is "this"--having `debug::print` work as a no-op?

tnowacki: Running `debug::print` as an op.

We have cases where we want it to print
- unit testing
- some sort of local deployment
- maybe some sort of devnet

We need to be able to configure it to print in those cases. And we want to make sure it does not accidentally start spewing in some "release" build for a real testnet or mainnet.

sblackshear: > As discussed this morning, this might be a reasonable way of letting debug prints occur on local deployments even when not in "test" mode

I think this problem was a nice-to-have for the future if we can figure out how to do it, but has to be connected to logging in the host system somehow (and I think a feature flag would be a good way). But I would vote on punting this problem for now and focusing on the Sui-side `debug` publishing issue

PR #415, opened 2022-08-28
Support loop unrolling in Move prover backend

macalinao:
## Motivation

Adds two new options to the Move prover:

- `prover.skip_loop_analysis`, which skips the `LoopAnalysisProcessor` in the proving pipeline
- `backend.loop_unroll`, which is an integer that specifies the `-loopUnroll` argument to Boogie

This allows simple, low iteration count loops to be proven without requiring specifying invariants.

Yes

## Test Plan

Ran this on my own code

### DIFF
--- a/language/move-prover/boogie-backend/src/options.rs
+++ b/language/move-prover/boogie-backend/src/options.rs
@@ -242,7 +242,6 @@ impl BoogieOptions {
             add(&[f.as_str()]);
         }
         add(&[boogie_file]);
-        println!("COMMAND: {:?}", &result);
         Ok(result)
     }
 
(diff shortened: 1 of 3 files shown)


wrwg (approved):
Nice -- maybe loop unrolling can help us with the current staking/governance verification problem?

meng-xu-cs (commented):
I am onboard with the loop-unrolling idea.

There are several places where we can unroll the loop:
1. Move compiler
2. Move bytecode transformation pipeline
3. Boogie backend

This PR is about loop unrolling at 3 which I am okay. But I would prefer loop unrolling to be done by our bytecode transformation pipeline, at step 2.

PR #416, opened 2022-08-28
Simplify serialized representation of ModuleId

macalinao:
## Motivation

ModuleIds are currently ugly to read when serialized into strings.

This renders `ModuleId`s as `short_str_lossless`s, which have a `0x` prefix.

- Before: `{ "address": "0000000000000000000000000000000000000123", "name": "abc" }`
- After: `0x123::abc`

Yes

## Test Plan

Unit tests


wrwg:
This has been attempted before by Greg I think. IIRC some wicked constraints hindered us to do this.

macalinao:
> This has been attempted before by Greg I think. IIRC some wicked constraints hindered us to do this.

what constraints?

wrwg:
> > This has been attempted before by Greg I think. IIRC some wicked constraints hindered us to do this.
> 
> what constraints?

That's why I added Greg and Todd.

language/move-core/types/src/language_storage.rs
```diff
@@ -142,11 +192,23 @@ impl ModuleId {
 
 impl Display for ModuleId {
     fn fmt(&self, f: &mut Formatter) -> std::fmt::Result {
-        write!(f, "{}::{}", self.address, self.name)
+        self.short_str_lossless().fmt(f)
```
jolestar, line 195: We should add `0x` prefix to ModuleId str, but should not use `short_str` in the output format, because `0x1` != `0x00000000000000000000000000000001` in str presentation, this will cause some compatibility issue in offchain usage(web3 js, SDK etc).

So I suggest:

1. When input, we can accept both short and full formats.
2. When output, we just output the full format.

gregnazario:
I think this is only part of the story:

We should make all address outputs consistent on all paths, and I agree with @jolestar 

0x before, and the full address for output (not short str)

But we can take in any of the current inputs (full lower or upper hex, shortened hex with a 0x, full hex with a 0x)

language/move-core/types/src/language_storage.rs
gregnazario: I've fixed the general input of `AccountAddress` accepts `000...01`, `0x1` and `0x00001` for serialization from JSON, and from string previously.

As for the output, it's a mess and comes out in either `000000...01` or `0x1`.  We should probably align similarly to `0x000...01`

macalinao: Ok-- i'll update this. Thanks for the feedback!

wrwg:
I'm concerned to land this now -- given @gregnazario's earlier experience and the risk that if we try to sync Aptos to this PR, tons of things are possible broken (and we are blocked to take changes from the Move repo until those things are fixed). At this point of the dev cycle, we cannot effort such a risk. I suggest to turn this into an issue and let some engineer handle this (across all relevant repos) as a starter task in the future.

macalinao:
> I'm concerned to land this now -- given @gregnazario's earlier experience and the risk that if we try to sync Aptos to this PR, tons of things are possible broken (and we are blocked to take changes from the Move repo until those things are fixed). At this point of the dev cycle, we cannot effort such a risk. I suggest to turn this into an issue and let some engineer handle this (across all relevant repos) as a starter task in the future.

This is fair-- I'll simplify this PR then by focusing only on the Display impl

PR #420, opened 2022-08-30
[move-compiler] implement the `#[verify_only]` attribute

meng-xu-cs:
As title suggests, this is an implementation of the `#[verify_only]`
annotation that can be used in a way that is similar to `#[test_only]`.

The semantics of `#[verify_only]` is the following:

Any AST element (in both source of in the library definitions)
annotated with the `#[verify_only]` attribute will be removed
after parsing, unless the `verify` flag is set when invoking
the compiler.

`#[verify_only]` has no interaction with unit-testing related
attributes such as `#[test]`, `#[test_only]` or `#[abort]`,
`#[expect_failure]` etc. A function that is marked `#[verify_only]` has
no knowledge of that a `#[test_only]` function might exist, unless both
`verify` and `testing` flags are set when invoking the compiler.

A large portion of this PR is in fact an refactoring of the `#[test_only]`
implementation as both `#[test_only]` and `#[verify_only]` involves
filtering AST nodes. This common functionality is abstracted and placed
into the `move-compiler/parser/filter.rs` file, and in particular, as a
`FilterContext` trait which can be customized as needed.

## Motivation

`#[verify_only]` has come up in several discussions related to prover improvements. This is an attempt to implement it and unblock other lines of work.

Yes

## Test Plan

New test cases added

### DIFF
--- a/language/move-compiler/src/command_line/compiler.rs
+++ b/language/move-compiler/src/command_line/compiler.rs
@@ -14,7 +14,7 @@ use crate::{
         CompilationEnv, Flags, IndexedPackagePath, NamedAddressMap, NamedAddressMaps,
         NumericalAddress, PackagePaths,
     },
-    to_bytecode, typing, unit_test,
+    to_bytecode, typing, unit_test, verification,
 };
 use move_command_line_common::files::{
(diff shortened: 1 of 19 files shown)


wrwg (approved):
Awesome!

tnowacki (commented):
Cool, thanks for generalizing the filtering logic!

PR #430, opened 2022-08-31
[move-prover] extract plain trace from boogie error reports

meng-xu-cs:
Boogies does not always give an augmented execution trace. Handle such
cases gracefully instead of giving out a mysterious error.

## Motivation

Graceful error handling

Yes

## Test Plan

CI


junkil-park (approved):
LGTM!

PR #433, opened 2022-09-01
[move-compiler] Support casting without parentheses

macalinao:
## Motivation

Easier to read and type if casts do not require enclosing parentheses.

Yes

## Test Plan

Updated language tests, and ran:

```
cargo test --workspace --exclude move-prover --exclude move-to-yul --exclude evm-exec-utils -- --skip prove
```

### DIFF
--- a/language/move-compiler/src/parser/syntax.rs
+++ b/language/move-compiler/src/parser/syntax.rs
@@ -984,10 +984,6 @@ fn parse_term(context: &mut Context) -> Result<Exp, Diagnostic> {
                     let ty = parse_type(context)?;
                     consume_token(context.tokens, Tok::RParen)?;
                     Exp_::Annotate(Box::new(e), ty)
-                } else if match_token(context.tokens, Tok::As)? {
-                    let ty = parse_type(context)?;
-                    consume_token(context.tokens, Tok::RParen)?;
-                    Exp_::Cast(Box::new(e), ty)
                 } else {
                     if context.tokens.peek() != Tok::RParen {
                         consume_token(context.tokens, Tok::Comma)?;
@@ -1229,7 +1225,7 @@ fn at_end_of_exp(context: &mut Context) -> bool {
         // These are the tokens that can occur after an Exp. If the grammar
         // changes, we need to make sure that these are kept up to date and that
         // none of these tokens can occur at the beginning of an Exp.
-        Tok::Else | Tok::RBrace | Tok::RParen | Tok::Comma | Tok::Colon | Tok::Semicolon
+        Tok::Else | Tok::RBrace | Tok::RParen | Tok::Comma | Tok::Colon | Tok::Semicolon | Tok::As
     )
 }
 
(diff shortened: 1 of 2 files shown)


wrwg:
I'm not sure what the consequences are. There is likely a reason why there are braces (like certain syntax constructs become ambiguous otherwise), but perhaps also not. Do you remember @tnowacki ? 

Unfortunately, these small changes look nice and harmless isolated but may have more indirect impact than we understand.

macalinao:
> I'm not sure what the consequences are. There is likely a reason why there are braces (like certain syntax constructs become ambiguous otherwise), but perhaps also not. Do you remember @tnowacki ?
> 
> Unfortunately, these small changes look nice and harmless isolated but may have more indirect impact than we understand.

Is there a way we can figure out what the impact could be via test cases? If so, what are some test cases we can construct to test this?

wrwg:
> > I'm not sure what the consequences are. There is likely a reason why there are braces (like certain syntax constructs become ambiguous otherwise), but perhaps also not. Do you remember @tnowacki ?
> > Unfortunately, these small changes look nice and harmless isolated but may have more indirect impact than we understand.
> 
> Is there a way we can figure out what the impact could be via test cases? If so, what are some test cases we can construct to test this?

We need @tnowacki input here because he was present when this grammar was designed.

sblackshear:
Agreed that we need @tnowacki's input here; I don't have enough context on the grammar. I do have some ideas for precedence stress tests:

- `my_num / 2 as u64`
- `my_num as u64 as u64`
- `a as u64 / b as u64`

tnowacki (commented):
Thanks for the PR! I had been thinking about this for a while as I know it was annoying (really bit me a few times recently too). 

The solution here provided feels fine in terms of preserving old behavior, but worth discussing the angles...

As it is today, `as` essentially has the lowest possible precedence. 
In Rust, `as` has the lowest precedence for binary operators. Which is nice, but always confused me personally when you have things like `1 + 2 as u64`. Enforcing parens was a clean solution to this, as it made it clear where cast started/stopped. 

With parens being gone in this PR, that is no longer true for some cases. However, all of the cases where this would let it appear, there is something else there to add clarity (I think)

Looking through the cases we have
- Field assignment `f: e`
- let binding `let l = e;`
- Sequencing `e;` or `{ ...; e }`
- Vector arguments `vector[e1, ..., en]`
- Tuples `(e1, ..., en)`
- Type annotations `(e: t)`
- If/while condition `if|while (e)`
- If/else/while/loop/return/abort blocks without braces `if|while|loop|return|abort e` and `if (cond) _ else e`
- Call args `f(e1, ..., en)`
- Various spec cases (but not super relevant here? I hope, since you probably don't need the casts)

In each one of these we basically have something already to remove ambiguity. Which maybe the exception of the control flow cases, but I think those read fine anyway, e.g.
`return u as u128`
`abort u as u64`
`if (cond) v as u8;`

The else case is a bit strange, but as things are with unbraced `else`s
These two bind under the `else`
`let x = if (cond) u as u64 else v as u64`
`let x = if (cond) x else v as u64`
This one applies to all of the `if`
`let x = if (cond) x else { v } as u64`

We will need a lot more tests though so walking through this to summarize:
- We could take Rust's decision and make `as` a very high precedence binary operator. This would be a breaking change (and I always found it confusing personally)
- I like this approach as it preserves old behavior while allowing new cases.
- Nearly all of the new cases have something that takes the place of of the parens, which is awesome!
- We will need lots of tests...

tnowacki (changes requested):
I think I like the approach! But we will need many more tests. Some that I can think of right now:

`+` is a test that should produce no compiler errors. `-` is a test that should produce an error
- `+` All cases where a top level-exp is used (see previous comment)
- `-` Various cases where it is still not allowed. Be sure to included nested cases and various cases with unary/binary operators (@sblackshear's cases were good starting points)
- `+/-` Cases with strange typing. Particularly for cases like 
```
let x: u64 = 0; 
let y: u8 = 1; 
let _: u64 = if (cond) y else x as u64
```
Should _not_ type check, but this one should... I think
```
let x: u64 = 0; 
let y: u8 = 1; 
let _: u64 = if (cond) x else y as u64
```

wrwg:
This PR has been orphan for a while. Everybody seems to want it but concerns are around breaking changes. @macalinao are you intending to continue work on this? Then please feel free to re-open. @tnowacki do you want to take this issue on your todo list? Then we may want to turn this into an issue instead.

tnowacki:
Filed an issue so at least this work won't be lost

PR #434, opened 2022-09-01
[move-core-types] Update ModuleId Display impl to use short_str_lossless

macalinao:
## Motivation

The current `Display` impl for `ModuleId` is ugly.

This was covered by https://github.com/move-language/move/pull/416, but this version has no breaking changes to the serialized representation.

Yes

## Test Plan

Ran Cargo tests

### DIFF
--- a/language/move-bytecode-verifier/transactional-tests/tests/ability_field_requirements/unrestricted_has_resource_field.exp
+++ b/language/move-bytecode-verifier/transactional-tests/tests/ability_field_requirements/unrestricted_has_resource_field.exp
@@ -1,7 +1,7 @@
 processed 1 task
 
 task 0 'publish'. lines 1-10:
-Error: Unable to publish module '00000000000000000000000000000042::Test'. Got VMError: {
+Error: Unable to publish module '0x42::Test'. Got VMError: {
     major_status: FIELD_MISSING_TYPE_ABILITY,
     sub_status: None,
     location: 0x42::Test,
--- a/language/move-bytecode-verifier/transactional-tests/tests/check_bounds/too_few_type_actuals.exp
+++ b/language/move-bytecode-verifier/transactional-tests/tests/check_bounds/too_few_type_actuals.exp
@@ -1,7 +1,7 @@
 processed 1 task
 
 task 0 'publish'. lines 1-11:
-Error: Unable to publish module '00000000000000000000000000000001::M'. Got VMError: {
+Error: Unable to publish module '0x1::M'. Got VMError: {
     major_status: NUMBER_OF_TYPE_ARGUMENTS_MISMATCH,
     sub_status: None,
     location: undefined,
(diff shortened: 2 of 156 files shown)


wrwg:
The change of display could also have a significant effect not only in baseline tests but also in functionality of REST apis etc. of various other repos (aptos, sui, starcoin). Do we have a way to check this? @tnowacki @sblackshear @jolestar what do you think should we take this PR?

sblackshear:
> The change of display could also have a significant effect not only in baseline tests but also in functionality of REST apis etc. of various other repos (aptos, sui, starcoin). Do we have a way to check this? @tnowacki @sblackshear @jolestar what do you think should we take this PR?

I am also worried about the upstream effects on REST--this would almost certainly be a breaking change on all three networks. `Display` gets used to convert `ModuleId`'s to strings in REST responses, and you could imagine that clients might rely on the address part in the string always being 20 bytes (or whatever the address length is). 

My inclination would be to say that this potential breakage isn't worth it to improve the expected value files, but definitely open to other opinions on this...

If others feel strongly about going forward with this, I think we need each of Aptos/Sui/StarCoin to try pulling in the PR and seeing what breaks.

jolestar:
As mentioned at https://github.com/move-language/move/pull/416 @gregnazario 


> We should make all address outputs consistent on all paths, and I agree with @jolestar
> 0x before, and the full address for output (not short str)

macalinao:
> As mentioned at #416
> 
> > We should add 0x prefix to ModuleId str, but should not use short_str in the output format, because 0x1 != 0x00000000000000000000000000000001 in str presentation, this will cause some compatibility issue in offchain usage(web3 js, SDK etc).
> > So I suggest:
> > When input, we can accept both short and full formats.
> > When output, we just output the full format.
> 
> @gregnazario
> 
> > We should make all address outputs consistent on all paths, and I agree with @jolestar
> > 0x before, and the full address for output (not short str)

Would implementing this make the PR get accepted?

Also, what if we just changed SDKs to not rely on address length? It seems weird to always have to worry about leading zeroes, and it makes tools incompatible between chains.

jolestar:
I support unifying all address output format, and this PR is part of that goal.

I think supporting more input formats and consistent output format is a simple strategy to handle this, the SDK should also follow this strategy.

The tool's compatibility between chains is another problem. Not only about leading zeros, we need to find another way.

gregnazario:
> I support unifying all address output format, and this PR is part of that goal.
> 
> I think supporting more input formats and consistent output format is a simple strategy to handle this, the SDK should also follow this strategy.
> 
> The tool's compatibility between chains is another problem. Not only about leading zeros, we need to find another way.

I have "almost" finished the full 0x0000...0001 format https://github.com/move-language/move/pull/441

with the caveat that I need buy in from the stakeholders here, as it's often a complaint by the community about 0s that randomly disappear.

Though the Move EVM stuff is giving me a hard time with hardhat and all

wrwg:
Unfortunately, changing the output here is a large problem because of compatibility. Instead we have started to introduce new functions `canonical_to_string` (for now in AcccountAddress) which we can incrementally call everywhere we want canonic representation. Closing this PR for now.

PR #435, opened 2022-09-01
[move-model] Add support for casts in pure functions

macalinao:
## Motivation

Casts should be allowed in pure functions.

Yes

## Test Plan

Will write prover tests.

### DIFF
--- a/language/move-model/tests/sources/cast_in_spec.exp
+++ b/language/move-model/tests/sources/cast_in_spec.exp
@@ -1,0 +1,1 @@
+All good, no errors!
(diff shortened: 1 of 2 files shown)


macalinao:
question: how would one add a range check to specifications in the generated Boogie? How do I add an "assert"?

wrwg:
Sorry for dropping the ball on this. We should probably allow casts in pure functions. Notice that in a specification function, you cannot have an assert, rather the result of the function is undetermined if parameters are not allowed. So a Move function converted to a specification function with a cast would be a partial function.

PR #442, opened 2022-09-03
[move-compiler] allow prover intrinsic types to be defined in spec

meng-xu-cs:
From the Move Prover's point of view, intrinsic types are types that
only have abstract/logic semantics instead of semantics suitable for
concrete execution. Therefore, these types are typically modeled with
special theories in the backend.

Typical examples of intrinsic types are vector, map, set, and possibly
strings. And usually one intrinsic type can have multiple concrete
implementations.

As a concrete example, both  `HashMap` and `BTreeMap` are implementations
for the `Map` intrinsic type, which holds the following key property:
```
\forall map, k, v: select(store(map, k, v), k) == v
```

This commit adds support in the Move Compiler to allow intrinsic types
to be defined in spec blocks.

An intrinsic types can be defined with the following syntax:
```
spec module {
    struct <TypeName> ("<" <OptionalTypeParams ">")
    ( "has" <OptionalAbilities> )
    ( :: [ <OptionalModeledTypes> ... ] );
}
```

For example:

```
module std::intrinsics {
    spec module {
        /// Defines a `Vec` intrinsic type where the
        /// built-in type `vector<E>` can be refined to.
        struct Vec<E> :: [ vector<E> ];

        /// Defines a `Map` intrinsic type. Currently there
        /// is no built-in model for `Map`.
        struct Map<K, V>;
    }
}
```

And later if there is ever a definition of another vector
or map implementation, we can use `pragma intrinsic = ...` to
connect them. E.g.,

```
module extension::big_vector {
    struct BigVector<E> { ... }
    spec BigVector {
        pragma intrinsic = std::intrinsics::Vector<E>;
    }
}

module extension::map {
    struct HashMap<K, V> { .. }
    spec HashMap {
        pragma intrinsic = std::intrinsics::Map<K, V>;
    }

    struct BTreeMap<K, V> { .. }
    spec BTreeMap {
        pragma intrinsic = std::intrinsics::Map<K, V>;
    }
}
```

## Motivation

Support intrinsic types for Prover

Yes

## Test Plan

- CI
- new test cases added for move-compiler


wrwg:
Let's say we have `spec struct Vec<E> :: [ vector<E> ]`. And then we later have `spec MyVec { pragma intrinsic = Vec<T> }`.  My understanding is that now wherever the concrete program uses `MyVec` we will use instead `Vec`. But what role does the `vector<E>` play?

In any case, just focusing on the substitution, I'm wondering what the consequences are for functions. If I declared an intrinsic substitution of `MyVec`, what happens now with the functions which take this one as a parameter? 

Perhaps we should have a better understanding first of the language semantics before we land the parsing. This problem is definitely harder than I thought.

meng-xu-cs:
> Let's say we have spec struct Vec<E> :: [ vector<E> ]. And then we later have spec MyVec { pragma intrinsic = Vec<T> }. My understanding is that now wherever the concrete program uses MyVec we will use instead Vec. But what role does the `vector<E>` play?

Good question, the reason we have `vector<E>` here is because we cannot do the following:
```
spec vector {
    pragma intrinsic = Vec<E>;
}
```
Therefore, we need this `:: [ vector<E> ]` to achieve the same effect.

> I'm wondering what the consequences are for functions. If I declared an intrinsic substitution of MyVec, what happens now with the functions which take this one as a parameter?

This will be answered in a follow-up PR (will update the link once there). The general design is:

```
module std::intrinsics {
spec module {
    struct Vec<E> :: [ vector<E> ];

    native fn index_of<E>(v: &Vec<E>, e: &E): (bool, num);
    native fn borrow_mut<E>(v: &mut Vec<E>, i: num): &mut E;
    // ... other intrinsic functions ...
}
}

module std::my_vec {
    struct MyVec<E> { .. }
    spec MyVec {
        pragma intrinsic = std::intrinsics::Vec<E>;
    }

    fn my_index_of<E>(v: &MyVec<E>, e: &E): (bool, u64) { .. }
    spec my_index_of {
        pragma intrinsic = std::intrinsics::index_of<E>;
        // move-model will check the signature and ensure that MyVec<E> unifies with Vec<E>
    }

    fn my_borrow_mut<E>(v: &mut MyVec<E>, i: u64): &mut E { .. }
    spec my_borrow_mut {
        pragma intrinsic = std::intrinsics::borrow_mut<E>;
    }
}
```

And later at the Boogie end, we will multiplex the prelude for each type
(i.e., `vector<E>` and `MyVec<E>`):

```
// prelude for `vector<E>`

type {:datatype} vector _;
function {:constructor} vector<T>(v: [int]T, l: int): vector T;

function vector_index_of<T>(v: vector T, e: T): int;
axiom {:ctor "vector"} (forall<T> v: vector T, e: T :: {vector_index_of(v, e)} {...}

// prelude for `MyVec<E>`

type {:datatype} $42_my_vec_MyVec _;
function {:constructor} 42_my_vec_MyVec<T>(v: [int]T, l: int): 42_my_vec_MyVec T;

function 42_my_vec_MyVec_my_index_of<T>(v: 42_my_vec_MyVec T, e: T): int;
axiom {:ctor "42_my_vec_MyVec"} (forall<T> v: 42_my_vec_MyVec T, e: T :: {42_my_vec_MyVec_my_index_of(v, e)} {...}
```

As for `borrow_mut` and `my_borrow_mut`, they will be handled in the borrow-analysis, with the knowledge
that they are bot intrinsic to `borrow_mut`.

> Perhaps we should have a better understanding first of the language semantics before we land the parsing.

We should, and this is going to be a stack of changes to it makes perfect sense to wait for the full stack to grow.

meng-xu-cs:
Closed in favor of #469

PR #443, opened 2022-09-05
[vm] Provide a feature flag to make linking of natives lazy

wrwg:
## Motivation

Adding new native functions to a running chain is cumbersome because a 2-staged roll out is needed: before the Move code with the new function can be loaded into a given node, that node must have upgraded so the native function's implementation is available.  We would rather like to control such upgrades by on-chain flags, that is allow an unbound native function but avoid calling it before we know the native code has been rolled out.

This PR adds a feature flag `lazy_natives` to the VM which if set, allows unbound native functions during bytecode verification time, but instead only errors if such a function is actually called. The change is safe because the linkage check is only postponed, not skipped. There is no performance implication because the current code loader does not cache resolution of natives, so that happens anyway already at execution time again.

Yes

## Test Plan

Existing tests.


wrwg:
Because this is behind a feature flag, we should be able to land this with just a few approvals (it has a bit of an urgency), but feedback is welcome also after landing. I think this might be useful also for other chains.

davidiw (commented):
Have you had a chance to test this? Looks good to me.

wrwg:
> Have you had a chance to test this? Looks good to me.

I'm not familiar with the older e2e test infra in the Move repo. This uses MVIR a language which I don't know and which I found counterintuitive to learn. I think eventually, we should push our new package-system based e2e-move-tests framework from aptos-core down to here. For now, I was planning to fly blind and test on aptos-core side.

davidiw (approved):

wrwg:
I think we should probably make the behavior of lazy binding the default in Move. Will put this on the agenda of the next maintainers meeting. 

```move
native fun stuff_v1();
native fun stuff_v2();

fun stuff() {
  if (chain_version() == v2) { stuff_v2() } else { stuff_v1() }
}
```

PR #444, opened 2022-09-05
[test] Transactional test runner support template vars

jolestar:
## Motivation

Support a variable context and template to get previous tests output as input for new tests.
This feature can support complex Move module tests.

```
//# publish
module A::M {
    struct Foo has key {
        x: u64,
    }

    public fun publish_foo(s: &signer) {
        move_to<Foo>(s, Foo { x: 500 })
    }
}

//# run --signers A
script {
    use A::M;

    fun main(s: signer) {
        M::publish_foo(&s);
    }
}


//# view --address A --resource 0x2a::M::Foo

//# run --signers A --args {{$.view[-1].x}}
script {

    fun main(_s: signer, x: u64) {
        assert!(x == 500, 1001);
    }
}
```

Yes

## Test Plan

Add new transactional test


language/testing-infra/transactional-test-runner/src/framework.rs
```diff
@@ -715,20 +725,27 @@ fn handle_known_task<'a, Adapter: MoveTestAdapter<'a>>(
...
     };
-    assert!(!result_string.is_empty());
 
-    writeln!(
-        output,
-        "\ntask {} '{}'. lines {}-{}:\n{}",
-        task_number, task_name, start_line, stop_line, result_string
-    )
-    .unwrap();
+    if let Some(s) = result_string {
```
runtian-zhou, line 738: We are changing the behavior here when no string is returned? Is there any justification for that?

language/testing-infra/transactional-test-runner/src/framework.rs
```diff
@@ -675,24 +676,32 @@ where
...
+        _ => (Some(first_task), None),
     };
     let (mut adapter, result_opt) =
         Adapter::init(default_syntax, fully_compiled_program_opt, init_opt);
     if let Some(result) = result_opt {
         writeln!(output, "\ninit:\n{}", result)?;
     }
+
+    if let Some(first_task) = first_task {
+        handle_known_task(&mut output, &mut adapter, &mut ctx, first_task);
```
runtian-zhou, line 699: Why is the first task different here?

runtian-zhou (commented):
Looks mostly good to me. Just had some questions about the potential logic difference between the old and new code.

language/testing-infra/transactional-test-runner/src/framework.rs
jolestar: The first task may be the `Init` command.
In the previous version, we pop up the first task and check if it is an `Init` command. If not, we `push_front`.

In the new version, the first task is a LazyTask, we parse it, and it becomes a Task, and we can not `push_front`.

language/testing-infra/transactional-test-runner/src/framework.rs
jolestar: In the previous version, if result_string is None, we just return. But in the new version, we continue to add the template var to context, then return.

wrwg:
Is this PR dormant?

jolestar:
> Is this PR dormant?

I will continue.

language/testing-infra/transactional-test-runner/src/framework.rs
```diff
@@ -137,7 +139,8 @@ pub trait MoveTestAdapter<'a>: Sized {
         args: Vec<<<Self as MoveTestAdapter<'a>>::ExtraValueArgs as ParsableValue>::ConcreteValue>,
         gas_budget: Option<u64>,
         extra: Self::ExtraRunArgs,
-    ) -> Result<(Option<String>, SerializedReturnValues)>;
+    ) -> Result<(Option<String>, SerializedReturnValues, Option<Value>)>;
```
tnowacki, line 142: Can we remove the `Option<Value>` from everything other than view?

language/testing-infra/transactional-test-runner/src/framework.rs
```diff
@@ -147,19 +150,20 @@ pub trait MoveTestAdapter<'a>: Sized {
...
+
     fn view_data(
         &mut self,
         address: AccountAddress,
         module: &ModuleId,
         resource: &IdentStr,
         type_args: Vec<TypeTag>,
-    ) -> Result<String>;
+    ) -> Result<(String, Value)>;
```
tnowacki, line 161: Two things:
1) I think we should maybe make this always return some form of structured value, and move the rendering into the framework
2) I think the structured value should be some enum that is essentially `Value { ModuleMetaData { ... } | Resource(MoveValue) }

tnowacki (changes requested):
Could you speak a bit more to the motivation for this change?

If you recall from the meeting last week, this framework is not designed to be used to write tests for end Move users, and is instead designed for writing tests for core Move components. That is, it is used for testing the verifier, the compiler, a chain's adapter, etc. 
Wanted to make sure we are on the same page for that before proceeding :) 

So the `.move` files need to make it really clear what the file is testing, so we know what we are covering when doing verifier or similar tests. In other words, we need to be able to quickly assess a test to determine what it is test, and to determine its coverage. 

For those reasons, I think pushing this all into json is worse than just copy pasting in the values. I have no idea what something like `{{$.view[-1].x}}` is going to resolve to, and as such I would have a hard time of assessing this test. 

Using something more structured with the Move value would I think make that easier to read. So something like  `--args resource<0x2a::M::Foo>.x` would be at least easier for me to read... but as all these functions should be deterministic... I'm not sure how useful that would really be. 

--- 

Open to discussion here! I am requesting changes because I think as is, it would hurt our ability to assess tests within the repo. Let me know your thoughts!

runtian-zhou:
> it would hurt our ability to assess tests within the repo

I don't see why that would hurt? IMHO the sole purpose of this PR is that we can get values from previous runs as arguments to the next transaction, which I think is a reasonable addition to the system.

tnowacki:
> I don't see why that would hurt?

I'm arguing it makes it harder to read the test. You can start getting really esoteric arguments that are difficult to decipher at a glance (at least I find it difficult to).  These tests are deterministic, just copy paste the values? 

For more complex things, sure I can see some value in being to specify arguments based on existing resources or other metadata. Just so the tests are easier to update or w/e. But we can build a simpler, more structured system for that that doesn't involve manipulating offsets in json objects. 
In the examples shown, @jolestar is trying to populate arguments from prior resources. So we should just build a more straightforward argument type for that case. 
I could then see doing similar things that introspect `SerializedReturnValues` from pervious arguments. 

In short: as far as I understand the motivating examples, there are shorter, cleaner ways to accomplish this PR.

jolestar:
> If you recall from the meeting last week, this framework is not designed to be used to write tests for end Move users, and is instead designed for writing tests for core Move components. That is, it is used for testing the verifier, the compiler, a chain's adapter, etc.

Yes, this test framework is for core Move components, and we used it to test MoveStd and Chain's Move Framework. The Move framework is a Move project, and I think this tool can also be used in the end Move user's projects. 

It is a good tool for testing Move modules, even better than many integration testing tools in the EVM ecosystem. so I hope we can introduce it to move-cli. But we can discuss this in the future, and now we can focus on this PR first.

>For those reasons, I think pushing this all into json is worse than just copy pasting in the values. I have no idea what something like {{$.view[-1].x}} is going to resolve to, and as such I would have a hard time of assessing this test.

>Using something more structured with the Move value would I think make that easier to read. So something like --args resource<0x2a::M::Foo>.x would be at least easier for me to read... but as all these functions should be deterministic... I'm not sure how useful that would really be.

The example is simple, so the value is deterministic. But in some complex tests, its value may depend on the parameters of the genesis, the init time of the genesis, etc. This approach can ensure that the genesis parameter changes do not break the test cases.

For some more complex tests, we can support the command to get the historical state and pass it as an argument to the test. Using historical state is also commonly used in the smart contract, such as Votes based on a historical token balance value.

We have introduced the transactional test to starcoin's move cli as an `integration-test` command, and support the end Move users use it tests Move project. And I think it is also useful for Aptos and Sui App developers.

More detailed information can be found in the following links:

1. Starcoin v1.12.3 Released with Improved Integration Testing for Complex Move SmartContract https://starcoin.medium.com/starcoin-v1-12-3-released-with-improved-integration-testing-for-complex-move-smartcontract-51cbcd4759ff
2. A integration test for DAO proposal which using this feature: https://github.com/starcoinorg/starcoin-framework/blob/main/integration-tests/daospace/dao_proposal.move
3. Some integration tests of the end Move user's projects https://github.com/movefuns/movefuns/tree/main/starcoin/integration-tests https://github.com/Elements-Studio/starswap-core/tree/master/integration-tests

language/testing-infra/transactional-test-runner/src/framework.rs
jolestar: every command needs the context value, not only the `view`, the `view` command just an example.

language/testing-infra/transactional-test-runner/src/framework.rs
jolestar: I would like to use an enum to replace all function's return values, but not all functions return the same results.

tnowacki:
I do not think this tool should be extended into the CLI. We already have something similar with the sandbox, that is right tool to extend, not this one.

jolestar:
> I do not think this tool should be extended into the CLI. We already have something similar with the sandbox, that is right tool to extend, not this one.

Do you mean the `exp-test`? the exp-test is based on the unit test, does not support a transaction flow, and does not support customing chain-specific commands. it is hard to test the DApp Move module.

tnowacki:
> > I do not think this tool should be extended into the CLI. We already have something similar with the sandbox, that is right tool to extend, not this one.
> 
> Do you mean the `exp-test`? the exp-test is based on the unit test, does not support a transaction flow, and does not support customing chain-specific commands. it is hard to test the DApp Move module.

`sandbox` is a bit out of date, but this is a good example:
`language/tools/move-cli/tests/sandbox_tests/offer_smoke/args.txt`

I'd rather see us generalize and open up `sandbox` than try to pull transactional tests into the CLI

jolestar:
> I'd rather see us generalize and open up `sandbox` than try to pull transactional tests into the CLI

Yes, the `sandbox test` is very like the `transactional tests`, but miss some features, and we need to improve it.

1. How to implement chain-specific custom commands? Most Move project tests need the chain-specific env.
2. How to define and manage many tests? I see every Move project only has one sandbox test config.

PR #450, opened 2022-09-07
[vm-status] Output substatus code as hex

gregnazario:
## Motivation

Substatus ends up being some number like 524289, which isn't easy to parse.

This changes it so that it's at least in hex, to separate the high and low bytes.

Yes

## Test Plan

Let's see how the tests run


wrwg (approved):
I'm fine with this but need to see what Mysten and Starcoin say.

jolestar (commented):
I think the digital format is more easy to read and more convenient to search the sub_status from the code.

gregnazario:
Realized we don't need this anymore

PR #461, opened 2022-09-08
[docs] Explain feature request process in CONTRIBUTING.md

sblackshear:
Explain how feature requests should be created and how they progress, add graveyard of features that were not accepted.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -135,3 +135,26 @@ able to be built and passes all checks performed by CI.
 Move uses [GitHub issues](https://github.com/move-language/move/issues) to track
 bugs. Please include necessary information and instructions to reproduce your
 issue.
+
+## Major feature requests
+
+Please begin by checking the following locations for duplicate requests:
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -135,3 +135,23 @@ able to be built and passes all checks performed by CI.
 Move uses [GitHub issues](https://github.com/move-language/move/issues) to track
 bugs. Please include necessary information and instructions to reproduce your
 issue.
+
+## Major feature requests
+
+Please begin by checking the [language feature request graveyard](GRAVEYARD.md) to ensure that your idea is not present.
```
wrwg, line 141: Shouldn't we also mention (resp. give a search URL) for existing proposed and accepted features so we don't get duplicates?

wrwg (approved):

CONTRIBUTING.md
sblackshear: Good call, will add links to the issue board filtered by the appropriate tags.

PR #467, opened 2022-09-09
[move-prover] switch src/dst pair for the havoc operator

### LINKED ISSUE #466: [Move Prover] Loop verification issue with multiple "continue" points
Prover cannot verify the inductive case of the invariant:
```
module 0x42::loop_inv {
    use std::vector;

    fun f() {
        let v1 = vector::empty<u64>();
        let v2 = vector::empty<u64>();
        let index = 0;
        let i = 0;

        while ({
            spec {
                invariant index == len(v1); // fail to verify the inductive case.
            };
            i < 10000
        }) {
            i = i + 1;
            if (i == 100) {
                continue
            };
            vector::push_back(&mut v1, index);
            vector::push_back(&mut v2, index);
            index = index + 1;
        };


        spec {
            assert index == len(v1);
        };
    }
}
```

Prover can verify it if we remove the following if-statement in the loop body:
```
if (i == 100) {
    continue
};
```

meng-xu-cs:
The `havoc` operator has been modeled as:
```
() := havoc($i);
```

i.e., 0 output and 1 input.

This commit switches the src and dest pair into
```
$i := havoc();
```

This is in fact closer to the semantic of `havoc`: it re-assigns the target with an arbitrary value. therefore, the target should be on the receiving end.

This has important effect on our program analysis passes, especially reaching definition analysis and live-var analysis.

## Motivation

Fix #466

Yes

## Test Plan

CI

### DIFF
--- a/language/documentation/examples/diem-framework/move-packages/DPN/sources/CRSN.move
+++ b/language/documentation/examples/diem-framework/move-packages/DPN/sources/CRSN.move
@@ -194,8 +194,9 @@ module DiemFramework::CRSN {
     }
 
     spec force_expire {
-        let addr = signer::address_of(account);
-        ensures global<CRSN>(addr).min_nonce == old(global<CRSN>(addr)).min_nonce + shift_amount;
+        // TODO: this ensures may not hold
+        // let addr = signer::address_of(account);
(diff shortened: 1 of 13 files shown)


wrwg (approved):

junkil-park (approved):
One unit test (`prover unit[default]::functional/loops_with_memory_ops.move`) failed. Other than that, LGTM! Thank you so much!

PR #469, opened 2022-09-10
[move-model] intrinsic declarations in struct spec

meng-xu-cs:
This PR allows intrinsic types and functions to be declared in struct spec block.

This feature is best highlighted in the following code:

```move
module 0x42::M {
    struct MyTable1<phantom K, phantom V> {}

    native fun new<K, V>(): MyTable1<K, V>;
    native fun destroy_empty<K, V>(t: MyTable1<K, V>);
    native fun borrow_mut<K, V>(t: &mut MyTable1<K, V>, k: K): &mut V;
    native fun length<K, V>(t: &MyTable1<K, V>): u64;

    spec native fun spec_len<K, V>(t: MyTable1<K, V>): num;
    spec native fun spec_set<K, V>(t: MyTable1<K, V>, k: K, v: V): MyTable1<K, V>;
    spec native fun spec_get<K, V>(t: MyTable1<K, V>, k: K): V;

    spec MyTable1 {
        pragma intrinsic = map,
            map_new = new,
            map_destroy_empty = destroy_empty,
            map_borrow_mut = borrow_mut,
            map_len = length,
            map_spec_len = spec_len,
            map_spec_get = spec_get,
            map_spec_set = spec_set;
    }
}
```

Pay attention to the spec block of  `MyTable1`. Instead of having
`pragma intrinsic = true;`, we now have
```move
    spec MyTable1 {
        pragma intrinsic = map,  // type mapping
            // a series of function mappings
            map_new = new,
            map_destroy_empty = destroy_empty,
            map_borrow_mut = borrow_mut,
            map_len = length,
            map_spec_len = spec_len,
            map_spec_get = spec_get,
            map_spec_set = spec_set;
    }
```

The move-model is clever enough to detect that the intrinsic declarations
are reasonable, e.g., no unrecognized functions, no duplication, etc,
(except for type checking, which is deferred to boogie).

This is the first step of implementing a better support for intrinsic types.
With this record, we can instantiate the Boogie templates into different
sets of preludes, each prelude maps for an intrinsic declaration.

To illustrate:

if we have two tables, `MyTable1` and `MyTable2` both declared to be
`pragma intrinsic = map, .....`

Then, we will instantiate the `Table.boogie` template twice, one for the type
`MyTable1` and the other for `MyTable2` and include both instantiations in the
prelude.

To be even more specific, this template will be instantiated into two prelude code:
- `map := MyTable1, fun_new := new_tb1, ...`
- `map := MyTable2, fun_new := new_tb2, ...`

```
type {:datatype} {{map}} _ _;

function {:constructor} {{map}}<K, V>(v: [K]V, e: [K]bool, l: int): {{map}} K V;

function {:inline} {{fun_new}}<K, V>(): {{map}} K V {
    {{map}}(DefaultTableArray(), DefaultTableKeyExistsArray(), 0)
}

function {:inline} {{fun_map_get}}<K,V>(t: {{map}} K V, k: K): V {
    v#{{map}}(t)[k]
}
```

This is a way simpler solution that #442.

And, not related to this PR, but features like `trait` can be supported
in a similar way in the prover.

## Motivation

Better support for intrinsic types

Yes

## Test Plan

CI

### DIFF
--- a/language/move-model/src/intrinsics.rs
+++ b/language/move-model/src/intrinsics.rs
@@ -9,7 +9,7 @@ use crate::{
     builder::module_builder::SpecBlockContext,
     model::{IntrinsicId, QualifiedId, SpecFunId},
     pragmas::{INTRINSIC_PRAGMA, INTRINSIC_TYPE_MAP, INTRINSIC_TYPE_MAP_ASSOC_FUNCTIONS},
-    symbol::Symbol,
+    symbol::{Symbol, SymbolPool},
     FunId, Loc, ModuleBuilder, StructId,
 };
 
@@ -17,7 +17,6 @@ use crate::{
 #[derive(Clone, Debug)]
 pub struct IntrinsicDecl {
     move_type: QualifiedId<StructId>,
-    #[allow(dead_code)]
     intrinsic_type: Symbol,
     intrinsic_to_move_fun: BTreeMap<Symbol, QualifiedId<FunId>>,
     move_fun_to_intrinsic: BTreeMap<QualifiedId<FunId>, Symbol>,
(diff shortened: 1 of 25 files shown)


wrwg:
Are we aiming to move forward with this?

meng-xu-cs:
That's my current intention. Maybe wait for the next PR to complete the whole feature?

meng-xu-cs:
@wrwg @junkil-park This PR is ready for review again.

The end result of this PR is that we can now have multiple implementations of the `table` intrinsic type and gets all of them verified in the Prover. This can be show cased in the test case [language/move-prover/tests/sources/functional/verify_custom_table.move](https://github.com/move-language/move/pull/469/files#diff-37166bb2fd78af2682495102c206213ab535bff56390b62d8fa4e7992355e1b1).

In this test case, we mocked another table implementation and checks that the mocked table implementation yields the same verification results as the `Table.move` in `extensions/`

language/move-model/src/intrinsics.rs
```diff
@@ -0,0 +1,337 @@
...
+    builder::module_builder::SpecBlockContext,
+    model::{IntrinsicId, QualifiedId, SpecFunId},
+    pragmas::{INTRINSIC_PRAGMA, INTRINSIC_TYPE_MAP, INTRINSIC_TYPE_MAP_ASSOC_FUNCTIONS},
+    symbol::{Symbol, SymbolPool},
+    FunId, GlobalEnv, Loc, ModuleBuilder, StructId,
+};
+
+/// An information pack that holds the intrinsic declaration
+#[derive(Clone, Debug)]
+pub struct IntrinsicDecl {
```
wrwg, line 20: Nice!

language/move-prover/tests/sources/functional/verify_custom_table.move
```diff
@@ -0,0 +1,187 @@
+// Similar to `verify_table.move`, but against a custom implementation.
+module 0x42::table {
+    struct Table<phantom K: copy + drop, phantom V> has store {}
+    spec Table {
```
wrwg, line 4: Yes!

wrwg (approved):
Let's move forward with this as it allows us to specify multiple data structures in the Aptos framework (Table, SimpleMap) with the same logical theories.

Issue #468, opened 2022-09-10
[Feature Request] Move Package Caching updates local state


gregnazario:
# 🚀 Feature Request

When I build a package, and it is using a git rev / branch, it should do a `git fetch & reset` to update the local cached git repository if you need updates.

## Motivation

It's a bad UX that the git repo never updates based on a branch.  Currently, people keep running into cache issues due to using branch names.

## Pitch

Right now users have to clear the cache, and it's not helpful

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))
Yes

sblackshear:
Thanks for working on this @gregnazario? Can this issue be closed now?

gregnazario:
Yes, now that it's finally not broken, I'll close it.

PR #471, opened 2022-09-10
[move-package] Update cached git packages to latest state automatically

gregnazario:
## Motivation

Currently, the cached packages are never updated, even though they check out the corresponding git repo.  This will now just fetch and reset accordingly to update the local cached version.  Note, this will require a network connection for whenever the compiler is run with a git dependency (though, normally you need it if you don't have it cached).

https://github.com/move-language/move/issues/468

Yes

## Test Plan

Tested with the Move CLI, it now properly downloads to the latest state.

### DIFF
--- a/language/tools/move-package/src/resolution/resolution_graph.rs
+++ b/language/tools/move-package/src/resolution/resolution_graph.rs
@@ -391,7 +391,7 @@ impl ResolvingGraph {
         Self::download_and_update_if_remote(
             dep_name_in_pkg,
             &dep,
-            self.build_options.skip_fetch_latest_git_deps,
+            self.build_options.fetch_latest_git_deps,
         )?;
         let (dep_package, dep_package_dir) =
             Self::parse_package_manifest(&dep, &dep_name_in_pkg, root_path)
@@ -533,7 +533,7 @@ impl ResolvingGraph {
             Self::download_and_update_if_remote(
                 *dep_name,
                 dep,
-                build_options.skip_fetch_latest_git_deps,
+                build_options.fetch_latest_git_deps,
             )?;
 
             let (dep_manifest, _) =
@@ -548,7 +548,7 @@ impl ResolvingGraph {
     fn download_and_update_if_remote(
         dep_name: PackageName,
         dep: &Dependency,
-        skip_fetch_latest_git_deps: bool,
+        fetch_latest_git_deps: bool,
     ) -> Result<()> {
         if let Some(git_info) = &dep.git_info {
             if !git_info.download_to.exists() {
@@ -578,7 +578,7 @@ impl ResolvingGraph {
                             dep_name
                         )
                     })?;
-            } else if !skip_fetch_latest_git_deps {
+            } else if fetch_latest_git_deps {
                 // Check first that it isn't a git rev (if it doesn't work, just continue with the fetch)
                 if let Ok(rev) = Command::new("git")
                     .args([
--- a/language/tools/move-package/src/resolution/resolution_graph.rs
+++ b/language/tools/move-package/src/resolution/resolution_graph.rs
@@ -388,7 +388,11 @@ impl ResolvingGraph {
         dep: Dependency,
         root_path: PathBuf,
     ) -> Result<(Renaming, ResolvingTable)> {
-        Self::download_and_update_if_remote(dep_name_in_pkg, &dep)?;
+        Self::download_and_update_if_remote(
+            dep_name_in_pkg,
+            &dep,
+            self.build_options.skip_fetch_latest_git_deps,
+        )?;
         let (dep_package, dep_package_dir) =
             Self::parse_package_manifest(&dep, &dep_name_in_pkg, root_path)
                 .with_context(|| format!("While processing dependency '{}'", dep_name_in_pkg))?;
@@ -526,7 +530,11 @@ impl ResolvingGraph {
         };
 
         for (dep_name, dep) in manifest.dependencies.iter().chain(additional_deps.iter()) {
-            Self::download_and_update_if_remote(*dep_name, dep)?;
+            Self::download_and_update_if_remote(
+                *dep_name,
+                dep,
+                build_options.skip_fetch_latest_git_deps,
+            )?;
 
             let (dep_manifest, _) =
                 Self::parse_package_manifest(dep, dep_name, root_path.to_path_buf())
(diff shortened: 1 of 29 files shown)


gregnazario:
There is a risk here that it changes the behavior.  You can't use a git repo to build Move while not connected to the internet e.g. on a plane.

I'd like to know the opinions accordingly

language/tools/move-package/src/resolution/resolution_graph.rs
```diff
@@ -565,6 +566,29 @@ impl ResolvingGraph {
                             dep_name
                         )
                     })?;
+            } else {
```
sblackshear, line 569: I think we can make this a bit lazier--if `git_rev` is a hash instead of a branch, updates aren't possible and we can skip the fetch.

language/tools/move-package/src/resolution/resolution_graph.rs
```diff
@@ -565,6 +566,29 @@ impl ResolvingGraph {
...
+                // If the current folder exists, do a fetch and reset to ensure that the branch
+                // is up to date
+                // NOTE: this means that you must run the package system with a working network connection
```
sblackshear, line 572: Might be helpful to make a note of this in the error message for the `git fetch` command

sblackshear:
> There is a risk here that it changes the behavior. You can't use a git repo to build Move while not connected to the internet e.g. on a plane.
> 
> I'd like to know the opinions accordingly

Good point! Couple of ideas on this:
- I think it would make sense to add a `-no-git-dep-updates` (or similar) CLI flag + suggest using this flag in the appropriate error message. This would allow you to keep building (but possibly without the latest git deps) if you don't have a network connection.
- See inline comment for making the fetch lazier--this should (hopefully) reduce the frequency of hitting the "can't build without a network connection" issue

sblackshear:
On a more long-term note (no changes suggested for this PR), I am a bit worried about allowing Move programmers to use deps that can skeakily change under their feet by default--seems like a good way to accidentally introduce breaking changes in the benign case, or malicious updates in the worst case. I think it's important to support auto-updateable deps in some form, but we should think carefully about the safest/most convenient way to do this.

One idea might be relying less on the global (and somewhat magical) Move build cache for dependency management and push for something more like `cargo vendor` https://doc.rust-lang.org/cargo/commands/cargo-vendor.html as the default. Will add this topic to the agenda for the next core contributor's meeting, in any case.

sblackshear (changes requested):
Upon further reflection: instead
- adding an option to skip fetching the latest git deps
- failing with an error message that tells the user to re-run with this option when they are offline

, perhaps we could just use a crate like https://crates.io/crates/online/1.0.0 to check for a network connection and silently skip the fetching if so (I am guessing this is what `cargo` does, since it works fine offline as long as you have the right deps). This will keep the  build system UX the same as it was before, while also fixing the update problem.

gregnazario:
> I'm wondering whether we have any examples of other language frameworks where things are automatically updated with each build. Does cargo does this? Not sure but I think not.
> 
> Best practice IMO should be to always use a static github reference (commit hash or stable branch).

I know that cargo does download a new one if you change it, but I'm unsure of whether you use a branch if it is.

Realistically, if you want a stable change you should use a git rev correct?

sblackshear:
>  Does cargo does this? Not sure but I think not.

Cargo does do this if you use a git dep on repo w/o specified rev or on a branch instead of a specific commit.

> Best practice IMO should be to always use a static github reference (commit hash or stable branch).

I wholeheartedly agree with this as best practices + the common case. But I do think there are use-cases for auto-updates, especially in the early days of a network. As one concrete example: applications depending on the Sui or Aptos Framework code. Projects do not have any agency over which version of the Framework they use--they have to go with whatever is published on-chain. E.g., in Sui right now, we recommend that Move projects use a git dep on our `devnet` branch, which gets updated once/week. Asking folks to change their Move.toml's every time we upgrade the devnet would be cumbersome, particularly because they have no choice but to use the newest version if they want their code to work on devnet.

gregnazario:
> > Does cargo does this? Not sure but I think not.
> 
> Cargo does do this if you use a git dep on repo w/o specified rev or on a branch instead of a specific commit.
> 
> > Best practice IMO should be to always use a static github reference (commit hash or stable branch).
> 
> I wholeheartedly agree with this as best practices + the common case. But I do think there are use-cases for auto-updates, especially in the early days of a network. As one concrete example: applications depending on the Sui or Aptos Framework code. Projects do not have any agency over which version of the Framework they use--they have to go with whatever is published on-chain. E.g., in Sui right now, we recommend that Move projects use a git dep on our `devnet` branch, which gets updated once/week. Asking folks to change their Move.toml's every time we upgrade the devnet would be cumbersome, particularly because they have no choice but to use the newest version if they want their code to work on devnet.

This is the exact case that I'm trying to fix.  Similarly people use the branch, and it gets stuck in time.  We then tell people they have to clean, and then redownload the whole git repos, which is a lot more cumbersome, as it uses a large overhead in networking.

Instead by updating the in place one, we take advantage of what git does best, incremental updates.

wrwg:
> > > Does cargo does this? Not sure but I think not.
> > 
> > 
> > Cargo does do this if you use a git dep on repo w/o specified rev or on a branch instead of a specific commit.
> > > Best practice IMO should be to always use a static github reference (commit hash or stable branch).
> > 
> > 
> > I wholeheartedly agree with this as best practices + the common case. But I do think there are use-cases for auto-updates, especially in the early days of a network. As one concrete example: applications depending on the Sui or Aptos Framework code. Projects do not have any agency over which version of the Framework they use--they have to go with whatever is published on-chain. E.g., in Sui right now, we recommend that Move projects use a git dep on our `devnet` branch, which gets updated once/week. Asking folks to change their Move.toml's every time we upgrade the devnet would be cumbersome, particularly because they have no choice but to use the newest version if they want their code to work on devnet.
> 
> This is the exact case that I'm trying to fix. Similarly people use the branch, and it gets stuck in time. We then tell people they have to clean, and then redownload the whole git repos, which is a lot more cumbersome, as it uses a large overhead in networking.
> 
> Instead by updating the in place one, we take advantage of what git does best, incremental updates.

Fine then with this as a general feature! I did not know about the cargo behavior here. How we use it in aptos should be a separate discussion.

wrwg (approved):

gregnazario:
@sblackshear I addressed your comments

sblackshear (approved):
Thanks for adding the option!

gregnazario:
Looks like some test was doing something wonky with opt out.  So, I've changed it to opt in (by default it doesn't fetch latest dependencies)

PR #472, opened 2022-09-12
[binary format] fix signature token deserialization bug

vgao1996:
This fixes a bug in the binary format deserializer regarding the handling of SignatureToken. The current implementation accepts a struct instantiation with arity 0 and will wrongfully apply the type constructor to the next argument.
```
        [
            SerializedType::STRUCT_INST as u8,
            0x0, /* struct handle idx */
            0x0, /* arity */
            SerializedType::BOOL as u8,
        ]
```

This gets it fixed by checking the arity and adds a few tests to ensure this case is covered.


wrwg (approved):
Great!

sblackshear:
Good catch Victor!!

wrwg:
> Good catch Victor!!

Auditing result ;-)

PR #479, opened 2022-09-13
[compatibility] Relax compatibility for friends

wrwg:
This adds a dynamic global flag which if set, lets the compatibility check treat friend functions like private functions. The feature is not on by default because it assumes that modules are compiled and published on package granularity level. On package level, we can assume that each call site of the friend function has been updated to the changed friend function as well.

Yes

## Test Plan

NA

### DIFF
--- a/language/move-binary-format/src/compatibility.rs
+++ b/language/move-binary-format/src/compatibility.rs
@@ -31,7 +31,11 @@ impl Compatibility {
     }
 
     /// Return compatibility assessment for `new_module` relative to old module `old_module`.
-    pub fn check(old_module: &Module, new_module: &Module) -> Compatibility {
+    pub fn check(
+        treat_friend_as_private: bool,
+        old_module: &Module,
+        new_module: &Module,
+    ) -> Compatibility {
         let mut struct_and_function_linking = true;
         let mut struct_layout = true;
 
@@ -102,6 +106,9 @@ impl Compatibility {
         // friend list. But for simplicity, we decided to go to the more restrictive form now and
         // we may revisit this in the future.
         for (name, old_func) in &old_module.exposed_functions {
+            if treat_friend_as_private && matches!(old_func.visibility, Visibility::Friend) {
+                continue;
+            }
             let new_func = match new_module.exposed_functions.get(name) {
                 Some(new_func) => new_func,
                 None => {
@@ -144,18 +151,17 @@ impl Compatibility {
             }
         }
 
-        // check friend declarations compatibility
-        //
-        // - additions to the list are allowed
-        // - removals are not allowed
-        //
-        // NOTE: we may also relax this checking a bit in the future: we may allow the removal of
-        // a module removed from the friend list if the module does not call any friend function
-        // in this module.
-        let old_friend_module_ids: BTreeSet<_> = old_module.friends.iter().cloned().collect();
-        let new_friend_module_ids: BTreeSet<_> = new_module.friends.iter().cloned().collect();
-        if !old_friend_module_ids.is_subset(&new_friend_module_ids) {
-            struct_and_function_linking = false;
+        if !treat_friend_as_private {
+            // check friend declarations compatibility
+            //
+            // - additions to the list are allowed
+            // - removals are not allowed
+            //
+            let old_friend_module_ids: BTreeSet<_> = old_module.friends.iter().cloned().collect();
+            let new_friend_module_ids: BTreeSet<_> = new_module.friends.iter().cloned().collect();
+            if !old_friend_module_ids.is_subset(&new_friend_module_ids) {
+                struct_and_function_linking = false;
+            }
         }
 
         Compatibility {
--- a/language/move-binary-format/src/unit_tests/compatibility_tests.rs
+++ b/language/move-binary-format/src/unit_tests/compatibility_tests.rs
@@ -82,7 +82,7 @@ const COMPATIBLE: Compatibility = Compatibility {
 fn deprecated_unchanged_script_visibility() {
     let script_module = mk_module(Visibility::DEPRECATED_SCRIPT);
     assert_eq!(
-        Compatibility::check(&script_module, &script_module,),
+        Compatibility::check(false, &script_module, &script_module,),
         COMPATIBLE
     );
 }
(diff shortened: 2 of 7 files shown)


language/move-binary-format/src/compatibility.rs
```diff
@@ -7,7 +7,16 @@ use crate::{
...
     normalized::Module,
 };
-use std::collections::BTreeSet;
+use std::{
+    collections::BTreeSet,
+    sync::atomic::{AtomicBool, Ordering},
+};
+
+/// Whether to treat friends as private during compatibility check.
+static TREAT_FRIEND_AS_PRIVATE: AtomicBool = AtomicBool::new(false);
```
vgao1996, line 16: I understand we are in a rush, but controlling this using an (effectively) mutable global variable really seems like an anti-pattern to me. Can we just pass this in as a parameter?

Notice that we already have an interface that passes in a verifier config
```
pub fn new_with_verifier_config(
        natives: impl IntoIterator<Item = (AccountAddress, Identifier, Identifier, NativeFunction)>,
        verifier_config: VerifierConfig,
    ) -> VMResult<Self> {
        Ok(Self {
            runtime: VMRuntime::new(natives, verifier_config)
                .map_err(|err| err.finish(Location::Undefined))?,
        })
    }
```
We can extend this easily by defining a VM config and change `new_with_verifier_config` to `new_with_vm_config`
```
struct MoveVMConfig {
    verifier_config: VerifierConfig,
    treat_friend_as_private: bool,
}
```

vgao1996 (changes requested):
I understand we are in a rush, but controlling this using an (effectively) mutable and shared global variable really seems like an anti-pattern to me. Can we just pass this in as a parameter?

Notice that we already have an interface that passes in a verifier config
```
pub fn new_with_verifier_config(
        natives: impl IntoIterator<Item = (AccountAddress, Identifier, Identifier, NativeFunction)>,
        verifier_config: VerifierConfig,
    ) -> VMResult<Self> {
        Ok(Self {
            runtime: VMRuntime::new(natives, verifier_config)
                .map_err(|err| err.finish(Location::Undefined))?,
        })
    }
```
We can extend this easily by defining a VM config and change `new_with_verifier_config` to `new_with_vm_config`
```
struct MoveVMConfig {
    verifier_config: VerifierConfig,
    treat_friend_as_private: bool,
}
```

sblackshear:
I'm personally ok with making this the default behavior rather than gating it with a feature flag:

- The VM's `publish` API currently forces you to have all `friend` modules published under the same address
- Sui's publishing construct works at the package level and does not allow X-package friends (and do not plan to change this)

We should check with @jolestar to see about StarCoin's policy on this.

wrwg:
> I understand we are in a rush, but controlling this using an (effectively) mutable and shared global variable really seems like an anti-pattern to me. Can we just pass this in as a parameter?
> 
> Notice that we already have an entry point that passes in a verifier config
> 
> ```
> pub fn new_with_verifier_config(
>         natives: impl IntoIterator<Item = (AccountAddress, Identifier, Identifier, NativeFunction)>,
>         verifier_config: VerifierConfig,
>     ) -> VMResult<Self> {
>         Ok(Self {
>             runtime: VMRuntime::new(natives, verifier_config)
>                 .map_err(|err| err.finish(Location::Undefined))?,
>         })
>     }
> ```
> 
> We can extend this easily by defining a VM config and change `new_with_verifier_config` to `new_with_vm_config`
> 
> ```
> struct MoveVMConfig {
>     verifier_config: VerifierConfig,
>     treat_friend_as_private: bool,
> }
> ```
> 
> Another reason why I think we should do it this way is because we'll need other config options to control the VM's behaviors in some other ways in the future, so why not just lay down the foundation for that now.

Unfortunately, compatibility checking is not happening in the verifier, but in the VM, in `VMRuntime::publish_module_bundle`.

> I understand we are in a rush, but controlling this using an (effectively) mutable and shared global variable really seems like an anti-pattern to me. Can we just pass this in as a parameter?
> 
> Notice that we already have an entry point that passes in a verifier config
> 
> ```
> pub fn new_with_verifier_config(
>         natives: impl IntoIterator<Item = (AccountAddress, Identifier, Identifier, NativeFunction)>,
>         verifier_config: VerifierConfig,
>     ) -> VMResult<Self> {
>         Ok(Self {
>             runtime: VMRuntime::new(natives, verifier_config)
>                 .map_err(|err| err.finish(Location::Undefined))?,
>         })
>     }
> ```
> 
> We can extend this easily by defining a VM config and change `new_with_verifier_config` to `new_with_vm_config`
> 
> ```
> struct MoveVMConfig {
>     verifier_config: VerifierConfig,
>     treat_friend_as_private: bool,
> }
> ```
> 
> Another reason why I think we should do it this way is because we'll need other config options to control the VM's behaviors in some other ways in the future, so why not just lay down the foundation for that now.

Did not know we already have this config, using it now.

wrwg:
> I understand we are in a rush, but controlling this using an (effectively) mutable and shared global variable really seems like an anti-pattern to me. Can we just pass this in as a parameter?
> 
> Notice that we already have an entry point that passes in a verifier config
> 
> ```
> pub fn new_with_verifier_config(
>         natives: impl IntoIterator<Item = (AccountAddress, Identifier, Identifier, NativeFunction)>,
>         verifier_config: VerifierConfig,
>     ) -> VMResult<Self> {
>         Ok(Self {
>             runtime: VMRuntime::new(natives, verifier_config)
>                 .map_err(|err| err.finish(Location::Undefined))?,
>         })
>     }
> ```
> 
> We can extend this easily by defining a VM config and change `new_with_verifier_config` to `new_with_vm_config`
> 
> ```
> struct MoveVMConfig {
>     verifier_config: VerifierConfig,
>     treat_friend_as_private: bool,
> }
> ```
> 
> Another reason why I think we should do it this way is because we'll need other config options to control the VM's behaviors in some other ways in the future, so why not just lay down the foundation for that now.

Wonderful -- did not know we have this config already. Using it now.

wrwg:
> I'm personally ok with making this the default behavior rather than gating it with a feature flag:
> 
> * The VM's `publish` API currently forces you to have all `friend` modules published under the same address
> * Sui's publishing construct works at the package level and does not allow X-package friends (and do not plan to change this)
> 
> We should check with @jolestar to see about StarCoin's policy on this.

Unfortunately, we need the flag for downwards compatible upgrades anyway. I hope we can get rid of it the next time we reset the network.

jolestar:
> I'm personally ok with making this the default behavior rather than gating it with a feature flag:
> 
> * The VM's `publish` API currently forces you to have all `friend` modules published under the same address
> * Sui's publishing construct works at the package level and does not allow X-package friends (and do not plan to change this)
> 
> We should check with @jolestar to see about StarCoin's policy on this.

Starcoin's policy is the same. The `friend` function is only available for the same address's module so that it can be treated as private.

davidiw (approved):

PR #483, opened 2022-09-15
[transactional-test] Record writeset in the golden files.

runtian-zhou:
## Motivation

Record the writesets of execution result in the transactional test. Right now the golden file contains few info in the VM related tests. Adding writesets will allow us to have a strong backward compatibility check as it will force VM to always return the same writeset for existing scripts.

Yes

## Test Plan

`cargo test` and regenerated all golden files


runtian-zhou:
Not sure if we want to enable this recording only for VM related tests..

language/testing-infra/transactional-test-runner/src/vm_test_harness.rs
```diff
@@ -203,7 +206,11 @@ impl<'a> MoveTestAdapter<'a> for SimpleVMTestAdapter<'a> {
                     format_vm_error(&e)
                 )
             })?;
-        Ok((None, serialized_return_values))
+
+        Ok((
+            Some(format!("ChangeSet: {:?}\n", change_set)),
```
runtian-zhou, line 211: This is the only change. Others are just regenerated golden files

wrwg (approved):
Great. Can we also cherrypick this into the aptos branch?

runtian-zhou:
> Great. Can we also cherrypick this into the aptos branch?

I think we should. Also would be a great chance to see any slippage happened between main and aptos branch.

tnowacki (changes requested):
Can you put this behind a flag?

language/testing-infra/transactional-test-runner/src/vm_test_harness.rs
tnowacki, line 211: We should really do some custom display here, feels a bit weird (and unlike the testing framework) to give just the raw debug dump

runtian-zhou:
Gate the change under a flag and enable it for `move-vm` transactional tests only. With that being said, I would suggest enable the writesets for @tnowacki as they will provide testings over critical backward compatibility guarantee that any network would need.

language/testing-infra/transactional-test-runner/src/vm_test_harness.rs
```diff
@@ -367,6 +385,29 @@ static MOVE_STDLIB_COMPILED: Lazy<Vec<CompiledModule>> = Lazy::new(|| {
...
+        false
+    }
+}
+
 pub fn run_test(path: &Path) -> Result<(), Box<dyn std::error::Error>> {
-    run_test_impl::<SimpleVMTestAdapter>(path, Some(&*PRECOMPILED_MOVE_STDLIB))
+    run_test_impl::<SimpleVMTestAdapter<SkipWriteSetMarker>>(path, Some(&*PRECOMPILED_MOVE_STDLIB))
+}
+
+pub fn run_test_with_writesets(path: &Path) -> Result<(), Box<dyn std::error::Error>> {
```
runtian-zhou, line 408: Add a separate harness to enable recording with writesets.

language/testing-infra/transactional-test-runner/src/vm_test_harness.rs
runtian-zhou: Any suggestions for what we should dump here?

language/testing-infra/transactional-test-runner/src/vm_test_harness.rs
```diff
@@ -36,13 +37,19 @@ use move_vm_runtime::{
...
 use once_cell::sync::Lazy;
+use std::marker::PhantomData;
 
 const STD_ADDR: AccountAddress = AccountAddress::ONE;
 
-struct SimpleVMTestAdapter<'a> {
+trait RecordWriteSet {
+    fn record_writeset() -> bool;
+}
+
```
tnowacki, line 47: This feels a bit excessive. But I get why you had to do it. Maybe a const generic would work a bit better... but some more thoughts on this below.

tnowacki (commented):
I think for like 99% of our tests, these change sets will be empty?

So I would have the following suggestions:
- See how much it would change the tests if we always printed it, but committed any printing when it was empty
- Pretty print the changes for the module/resource sections
- Omit the actual data, and only print the identifier/tag + the op. The data there does not need to be tested here 
- We could maybe print out something nice and slick, like a diff for the account: e.g.
```
0x42: 
  + module A
  + module B
  - resource 0x1::coin::Coin
  + resource 0x1::flag::Flag
  +/- resource 0x3::flargle::Blargle 
```

language/testing-infra/transactional-test-runner/src/vm_test_harness.rs
```diff
@@ -203,7 +213,11 @@ impl<'a> MoveTestAdapter<'a> for SimpleVMTestAdapter<'a> {
                     format_vm_error(&e)
                 )
             })?;
-        Ok((None, serialized_return_values))
+
+        Ok((
+            T::record_writeset().then(|| format!("ChangeSet: {:?}\n", change_set)),
```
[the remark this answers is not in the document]
jolestar: Maybe use ChangeSet to json? json is more formal than debug output, avoid breaking the testes when debugging output change.

wrwg:
Is this PR still active?

runtian-zhou:
Lemme close this for now just so that we can revisit the golden file situation later.

PR #484, opened 2022-09-16
[move-unit-test] Create an option to print writeset of execution

runtian-zhou:
## Motivation

Add an option to emit the writeset of each transaction towards the end of testing. This would allow us to compare the output against a golden file for upstream testing.

Yes

## Test Plan

`cargo test` and updated the new baseline files.


runtian-zhou:
One question here is do we want to pipe the writeset output to a separate stream or use the current stream for the golden file? We also might want to cherry pick this PR to aptos branch so that we can generate the golden files in the aptos repo. cc @wrwg

wrwg (approved):
Great!

We had folks frequently complaining about tons of golden tests failing all the time over the Aptos repo, with little practical advantage. When we use that over there, we need to discuss where.

language/tools/move-unit-test/tests/test_sources/native_signer_creation.exp
```diff
@@ -2,6 +2,12 @@ Running Move unit tests
...
...ructTag { address: 00000000000000000000000000000001, module: Identifier("M"), name: Identifier("A"), type_params: [] }: New([0])} }, 05000000000000000000000000000000: AccountChangeSet { modules: {}, resources: {StructTag { address: 00000000000000000000000000000001, module: Identifier("M"), name: Identifier("A"), type_params: [] }: New([0])} }, 06000000000000000000000000000000: AccountChangeSet { modules: {}, resources: {StructTag { address: 00000000000000000000000000000001, module: Identifier("M"), name: Identifier("A"), type_params: [] }: New([0])} }, 07000000000000000000000000000000: AccountChangeSet { modules: {}, resources: {StructTag { address: 00000000000000000000000000000001, module: Identifier("M"), name: Identifier("A"), type_params: [] }: New([0])} }, 08000000000000000000000000000000: AccountChangeSet { modules: {}, resources: {StructTag { address: 00000000000000000000000000000001, module: Identifier("M"), name: Identifier("A"), type_params: [] }: New([0])} }, 09000000000000000000000000000000: AccountChangeSet { modules: {}, resources: {StructTag { address: 00000000000000000000000000000001, module: Identifier("M"), name: Identifier("A"), type_params: [] }: New([0])} }} })
```
tnowacki, line 10: This really feels like a bit of a hack, and I don't find the output here really readable. 
Could we at least change the format here to be something human readable rather than the debug dump?

tnowacki: More on this in #483

PR #493, opened 2022-09-20
[move-vm] Runtime resource checks

runtian-zhou:
## Motivation

Performing runtime checks on struct(resource) that's created in MoveVM. 

Tagging the type of values at runtime to prevent unexpected bytecode that passes the verifier. This mean the following:
1. Each struct will now carry a u64 value that is a hash digest of its type
2. Struct will be tagged either when they are being loaded from the global storage or when created by pack/packgeneric instruction.
3. Tag of a struct can never be modifed.

Currently there are two locations where we check whether type matches:
1. When resource gets serialized back to storage, we will make sure all the inner nested resource matches the type declared in the struct instantiation info.
2. When VM tries to mutate a struct by reference, it will make sure the operand type has the same u64 tag as the reference type.
3. When value gets passed to another function, we will make sure the underlying types matches with the signature of the callee function.

This would NOT be a full mitigation for the unintended reinterprect cast/type conversion as we didn't check every single memory cell. However, I would argue that this PR should provide us resource safety at runtime. The reason is that even if a value could somehow trick VM into a wrong type by misplacing a value to a cell where it shouldn't store, there's no way for one to write this value to storage nor invoke it to another module. Thus it is effectively a "dead" struct sitting on top of a local frame.

Right now the feature is gated under a runtime flag called `strict_runtime_verification` under `VerifierConfig` and is off by default. However, we may want to revisit if we would like to enable this check by default.

Yes

## Test Plan

`cargo test`.

Not sure what is the best way to test this behavior in CI. For a one off test I tried to run the type conversion bug that @chen-robert found with this extra runtime check and the PR is able to capture the problem.


language/move-vm/types/src/values/values_impl.rs
```diff
@@ -230,7 +231,85 @@ impl Container {
...
+                }
+            }
+            (Self::Struct(_, None), _) => Err(PartialVMError::new(
+                StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR,
+            )
+            .with_message(format!("value not tagged"))),
+            (Self::Locals(_), _) | (Self::Vec(_), _) => Err(PartialVMError::new(
+                StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR,
+            )
+            .with_message(format!("unexpected type mismatch"))),
```
vgao1996, line 265: Maybe change the message to something like "should be unreachable"?

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -230,7 +231,85 @@ impl Container {
...
+    }
+
+    fn check_type(&self, ty: &Type) -> PartialVMResult<()> {
+        match (self, ty) {
+            (Self::VecAddress(_), _)
+            | (Self::VecBool(_), _)
+            | (Self::VecU128(_), _)
+            | (Self::VecU64(_), _)
+            | (Self::VecU8(_), _)
+            | (Self::Vec(_), Type::Vector(_)) => Ok(()),
```
vgao1996, line 247: These don't seem correct to me. I think
- For the specialized vectors
  - We should check if `ty` is indeed a vector of this particular type
- For the general vector
  - We should check if 
    - `ty` is indeed a vector type with non-specialized element types
    - All elements have the right type <= I would be concerned if we actually do this though, since it makes passing a vector to a different function O(n)

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -64,7 +64,7 @@ enum ValueImpl {
 enum Container {
     Locals(Rc<RefCell<Vec<ValueImpl>>>),
     Vec(Rc<RefCell<Vec<ValueImpl>>>),
-    Struct(Rc<RefCell<Vec<ValueImpl>>>),
+    Struct(Rc<RefCell<Vec<ValueImpl>>>, Option<Type>),
```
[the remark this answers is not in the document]
vgao1996: I feel like we should move the `Option<Type>` into the `Rc`. Otherwise this is gonna increase the size of the Value a lot.

language/move-core/types/src/value.rs
```diff
@@ -257,11 +263,25 @@ impl MoveStructLayout {
...
             }
         }
     }
+
+    pub fn tag(&self) -> Option<u64> {
+        match self {
+            Self::Runtime(_) => None,
+            Self::CheckedRuntime { fields: _, tag } => Some(*tag),
+            Self::WithFields(_) | Self::WithTypes { .. } => {
+                // It's not possible to implement this without changing the return type, and some
```
sblackshear, line 278: None?

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -262,15 +262,38 @@ impl Interpreter {
...
+                        let err =
+                            PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
+                                .with_message(
+                                    "Unexpected native function not located in a module".to_owned(),
+                                );
+                        self.set_location(err)
+                    }
+                })?;
+
+            v.check_type(&expected_ty)
```
sblackshear, line 293: I think it makes more sense to check the type at each instruction. For calls, that will cover both the argument and return types, whereas this only covers the argument types.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -1000,21 +1025,44 @@ impl Frame {
...
-                    Bytecode::Unpack(_sd_idx) => {
+                    Bytecode::Unpack(sd_idx) => {
                         let struct_ = interpreter.operand_stack.pop_as::<Struct>()?;
+                        let ty = resolver.get_struct_type(*sd_idx);
+                        if let Some(ty1) = struct_.tag() {
+                            if ty1 != ty.get_hash() {
+                                return Err(PartialVMError::new(
+                                    StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR,
+                                )
+                                .with_message(format!("unexpected type mismatch")));
```
sblackshear, line 1045: I think more detail in these messages (e.g., name of the instruction in question, the two types) would be helpful.

language/move-vm/runtime/src/interpreter.rs
runtian-zhou: I feel like this would be a much greater change than this PR: this mean that we want to have a typed operand stack which could be hard to construct with the current design, as the VM right now doesn't carry any type info with instruction which made it quite hard to check. I think maybe we can get started from checking the following three locations:

1. Function Call
2. WriteRef(not in this PR)
3. Write to storage

I think these should offer us enough protection over unexpected type conversion.

runtian-zhou:
Another thing that might need an extra eye on for @vgao1996 is whether this PR changed the gas behavior for deeply nested resources. I don't think that's the case but would love to get your double check as well

language/move-vm/runtime/src/interpreter.rs
runtian-zhou: Done!

runtian-zhou:
Another thing worth noting is whether we want the checks here to be optional? cc @sblackshear

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -1000,21 +1025,44 @@ impl Frame {
...
-                            .operand_stack
-                            .push(Value::struct_(Struct::pack(args)))?;
+                        let value = Value::struct_(Struct::pack_with_tag(
+                            args,
+                            resolver
+                                .instantiate_generic_type(*si_idx, self.ty_args())?
+                                .get_hash(),
+                        ));
+
+                        interpreter.operand_stack.push(value)?;
```
vgao1996, line 1035: We need to evaluate the perf implications this might have very carefully. `instantiate_generic_type` is a very expensive operation and by calling it we're also making pack/unpack no longer O(1)

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -1811,6 +1832,91 @@ impl Vector {
...
+impl ValueImpl {
+    fn check_type(&self, ty: &Type) -> PartialVMResult<()> {
+        match (self, ty) {
+            (ValueImpl::U8(_), Type::U8)
+            | (ValueImpl::U128(_), Type::U128)
+            | (ValueImpl::U64(_), Type::U64)
+            | (ValueImpl::Bool(_), Type::Bool)
+            | (ValueImpl::Address(_), Type::Address) => Ok(()),
+            (ValueImpl::IndexedRef(r), Type::Reference(ty) | Type::MutableReference(ty)) => {
+                r.copy_value().read_ref()?.check_type(ty)
```
vgao1996, line 1856: This is very concerning... `read_ref` makes a deep copy and can get super expensive if your value is big. Can we find a way to check the value type without calling `read_ref`?

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -1811,6 +1832,91 @@ impl Vector {
...
+}
+
+impl Container {
+    fn check_type(&self, ty: &Type) -> PartialVMResult<()> {
+        match (self, ty) {
+            (Self::VecAddress(_), _)
+            | (Self::VecBool(_), _)
+            | (Self::VecU128(_), _)
+            | (Self::VecU64(_), _)
+            | (Self::VecU8(_), _) => Ok(()),
```
vgao1996, line 1887: I don't think this is correct. We should check if the type is indeed a `vector<address> or vector<bool>, ...`

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -1811,6 +1832,91 @@ impl Vector {
...
+            (Self::Vec(vs), Type::Vector(ty)) => {
+                for v in vs.borrow().iter() {
+                    v.check_type(ty)?;
+                }
```
vgao1996, line 1891: This is also very concerning. Here we're looping over the whole vector... Overall I feel like the recursive type check is equivalent to a deep copy, time complexity-wise.

language/move-vm/runtime/src/interpreter.rs
vgao1996, line 293: As I commented under `fn check_type`, this operation performs a deep traversal over the whole value and is very concerning in terms of performance and gas

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -64,7 +64,7 @@ enum ValueImpl {
 enum Container {
     Locals(Rc<RefCell<Vec<ValueImpl>>>),
     Vec(Rc<RefCell<Vec<ValueImpl>>>),
-    Struct(Rc<RefCell<Vec<ValueImpl>>>),
+    Struct(Rc<RefCell<Vec<ValueImpl>>>, Option<u64>),
```
vgao1996, line 67: Again, I think we should move this into the `Rc` or otherwise `Value` is gonna become 16B larger

language/move-vm/runtime/src/interpreter.rs
runtian-zhou: We may be able to cache this computation because the tag should remain unchanged for this instruction

language/move-vm/runtime/src/interpreter.rs
vgao1996: Yeah I guess I'm a little bit torn on this... On one hand, I understand that if you want to be absolutely secure, you need to check the whole value recursively across function boundaries. On the other hand, I'm also really concerned about the perf penalty we're incurring here.

cc @wrwg

vgao1996:
Putting aside the perf/gas concerns, I think the overall logic is now a lot cleaner compared to the first version.

language/move-vm/types/src/values/values_impl.rs
runtian-zhou: This could be a much bigger change cuz I think the vm value is using `Rc<RefCell<Vec<ValueImpl>>>` everywhere.. changing this could yield a bigger refactor.

language/move-vm/types/src/values/values_impl.rs
runtian-zhou: The reason why we are looping over vector is because we don't perform check for write_ref. If we don't do that, we can pass in a struct/vector that contains a resource that experienced a type hack. Lots of Move code right now are using reference to resource as authentication and if we missed that it would be a horrible error. Think about a passing to callee a`Vec<Capability>` where a value underneath is a `Foo`.

runtian-zhou:
> perf/gas concerns

IMHO I would agree with Robert that even if our execution costs 10% more computation it is still a trade-off worth making. Any of an unexpected reinterpret cast type of bug can easily cost us way more than the perf/gas cost itself.

vgao1996:
@runtian-zhou you have a very valid point, but just to be clear that if we make the changes you are suggesting, we'll need to change the gas formulae so that when calling functions, we charge an amount that's proportional to the size of the value. Luckily when I came up with the gas meter trait, I intentionally made this possible/trivial to do via the value view, but never thought we would actually start to use it this soon...

runtian-zhou:
> @runtian-zhou you have a very valid point, but just to be clear that if we make the changes you are suggesting, we'll need to change the gas formulae so that when calling functions, we charge an amount that's proportional to the size of the value. Luckily when I came up with the gas meter trait, I intentionally made this possible/trivial to do via the value view, but never thought we would actually start to use it this soon...

BTW I think the biggest concern you have is for checking values in a vector. I think we don't need to check that if we have a full protection over `write_ref` instruction. So we may even don't need to use that.

vgao1996:
Again, just to be clear that I'm not inherently against these dynamic checks either. (I think they are a very nice thing to have conceptually.) It's just that the degeneration from O(1) to O(n) makes me nervous and I want to make sure we fully understand the tradeoffs. Should probably do some real profiling to find out the real costs...

runtian-zhou:
> Again, just to be clear that I'm not inherently against these dynamic checks either. (I think they are a very nice thing to have conceptually.) It's just that the degeneration from O(1) to O(n) makes me nervous and I want to make sure we fully understand the tradeoffs. Should probably do some real profiling to find out the real costs...

Yea that would be awesome.

language/move-vm/runtime/src/interpreter.rs
runtian-zhou: This is actually the thing might be concerning, cuz we won't be able to avoid this check.

vgao1996:
At meanwhile, let me think about if there's a more efficient method that would give us the same results.

vgao1996:
Overall my biggest two concerns are
1. Perf/gas, which we've already discussed
2. The complexity these dynamic checks themselves introduce and whether they can destabilize the existing (arguably relatively mature) code.

language/move-vm/types/src/loaded_data/runtime_types.rs
```diff
@@ -123,4 +125,10 @@ impl Type {
                 .fold(Self::LEGACY_BASE_MEMORY_SIZE, |acc, ty| acc + ty.size()),
         }
     }
+
+    pub fn get_hash(&self) -> u64 {
```
chen-robert, line 129: How collision resistant is `get_hash`? It's not immediately obvious to me that a `u64` hash is sound.

runtian-zhou: The interesting part is that `Type` was using `CachedStructIndex`, which I believe is a value that would change depending on the loading order of the modules. So IMO a hash collision may be quite hard as attacker won't hold the control to preimage of the hash.

chen-robert (commented):
Would it make sense to do validation on `VecPack` and `VecPushBack` to further mitigate having vectors with invalid types in them?

runtian-zhou:
> Would it make sense to do validation on `VecPack` and `VecPushBack` to further mitigate having vectors with invalid types in them?

I think if we check the property that when we write to a reference, the operand should have the same type as the reference type, we should be good?

language/move-vm/types/src/loaded_data/runtime_types.rs
chen-robert: That's a good point -- it should even be different across different validators, which would make a collision impossible? I think in practice this shouldn't be an issue. Is there a reason you use a hash instead of just the struct index though?

chen-robert:
This would be a potentially better/alternative/additional mitigation to recursing through the entire vector in `check_type`

language/move-vm/types/src/loaded_data/runtime_types.rs
runtian-zhou: Yes I think there's randomness in rust's default hasher. The reason is simply the overhead to carry the full info around is costly. If it is a generic struct it's not going to be a single struct index, but a index with an instantiation vector as well.

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -670,7 +677,30 @@ impl ContainerRef {
                 }
 
                 match self.container() {
-                    Container::Struct(r) => assign!(r, Struct),
+                    Container::Struct(r, ref_tag) => {
+                        let r1 = match c {
+                            Container::Struct(v, tag) => {
+                                if *ref_tag != tag {
```
runtian-zhou, line 683: > This would be a potentially better/alternative/additional mitigation to recursing through the entire vector in `check_type`

I just added the check! It was surprisingly a one line change. I'll see how I can potentially relax the check_type

language/move-vm/types/src/loaded_data/runtime_types.rs
vgao1996: I think we can salt the hash if we are concerned.

zjma: it's not clear if the hash here needs to be cryptographically collision resistant? In other words, if someone finds a collision, what's the worst thing can happen? Or usually if the hash is just used in things like a HashMap for bucketing things then cryptographic collision resistance is not needed.

runtian-zhou: ^ I think the thing here is the preimage of the hash is not controlled by anyone. If ppl find collision, they still need to find a verifier bug in order to utilize it.

chen-robert: The issue is, it represents a bypass of this particular runtime type-safety mitigation -- ie the point of this PR. Any sort of type confusion would require a verifier bug to exploit.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -1000,21 +1025,44 @@ impl Frame {
...
                     }
-                    Bytecode::Unpack(_sd_idx) => {
+                    Bytecode::Unpack(sd_idx) => {
                         let struct_ = interpreter.operand_stack.pop_as::<Struct>()?;
+                        let ty = resolver.get_struct_type(*sd_idx);
```
jolestar, line 1039: I think we should pass the `ty` to `stake.pop_as` and check the type in the `Value::value_as` function, and make sure all type cast has checked the type.

How about we carry the StructType to Value, and check the make sure the `value_as` 
  
https://github.com/move-language/move/blob/98312063e3b11b21c83a9d4f874725a50afa3dcf/language/move-vm/types/src/values/values_impl.rs#L1214-L1221

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -140,6 +140,7 @@ pub enum IntegerValue {
 #[derive(Debug)]
 pub struct Struct {
     fields: Vec<ValueImpl>,
+    ty: Option<u64>,
```
jolestar, line 143: How about carrying the TypeTag here? and avoid the hash function call. this uses more memory but is better for performance.

language/move-vm/types/src/loaded_data/runtime_types.rs
runtian-zhou: My point here is collision is not reproducable: the preimage of the type contain an index that is generated by the order how VM loads up a module. The default hasher IIUC is also non-determinisitic so I don't see how attacker can utilize the fact that the default hasher is not cryptographic secure, other than continuously resubmitting the same transaction in hope that a collision would happen. Even if in some random scenario where such collision happen, it requires all nodes in the validator set to have the same collision in order to get pass this check, plus being able to find a type safety verifier bug. These are the reasons why I think using a non-cryptographic secure function is OK. 

With that being said, maybe it's just safer to use a cryptographic secure function for now and we can see how much of perf penalty this would add. The main reason I was thinking about using default hasher is just it should be quicker and we are not persisting this piece of data anywere.

language/move-vm/types/src/values/values_impl.rs
runtian-zhou: I'm not sure how much more perf we will be saving here as the resolving types in runtime to `TypeTag` is actually quite costly, and less memory efficient.

language/move-vm/runtime/src/interpreter.rs
runtian-zhou: Right now I'm just focusing tagging the struct that gets created at runtime. Other types of type mismatch is not captured with this checks. The main reason is that I think the resource safety is probably the first thing we need to tackle here and didn't want to make this PR too complicated.

jolestar: Got it. But we need to copy this code to `Bytecode::VecPack` and ``Bytecode::VecUnpack`

runtian-zhou: So the interesting part is that we don't need to check the two instructions. The two instructions won't taint the tag. Thus if there's a type mismatch, it will be captured either:
1. when you are in a module trying to store the mismatched value to storage
2. when you try to pass the mismatch module to another function

We are actually using the resource property of Move here so that a value that has a bad type essentially becomes a hot potato here that can't be passed to another function nor stored, making it impossible to benefit from the exploit.

runtian-zhou: With that being said, I think a typed stack as @sblackshear suggests is a really interesting approach. I'm just not sure how fast we could implement that as the bytecode right now has very few information about the type stored.

language/move-vm/types/src/values/values_impl.rs
runtian-zhou: With the new check on references, maybe we can omit this check.

runtian-zhou:
Published a new commit to make these checks optional and configurable at runtime.

language/move-vm/types/src/values/values_impl.rs
runtian-zhou: Fixed... I thought those values to be uninteresting so didn't implement that in my first draft :(

language/move-vm/runtime/src/interpreter.rs
jolestar: If we store the type when push to stack, and check it when pop, maybe do not need the bytecode to store the type information.

sblackshear:
> Again, just to be clear that I'm not inherently against these dynamic checks either. (I think they are a very nice thing to have conceptually.) It's just that the degeneration from O(1) to O(n) makes me nervous and I want to make sure we fully understand the tradeoffs. Should probably do some real profiling to find out the real costs...

To make sure I understand: does this generation only happen when strict mode is on, or do we incur some runtime costs even when it is off?

runtian-zhou:
> > Again, just to be clear that I'm not inherently against these dynamic checks either. (I think they are a very nice thing to have conceptually.) It's just that the degeneration from O(1) to O(n) makes me nervous and I want to make sure we fully understand the tradeoffs. Should probably do some real profiling to find out the real costs...
> 
> 
> 
> To make sure I understand: does this generation only happen when strict mode is on, or do we incur some runtime costs even when it is off?

In this implementation the only extra cost when it is off is the extra memory overhead of the container: it's going to be a few bytes larger because of the memory space reserved for the tag. Otherwise I don't see an extra runtime cost here.

sblackshear:
Closing was a mis-click on my part, please ignore!

chen-robert:
Copying from Slack:

Actually, native functions might be more nontrivial than I initially expected — if we ignore type checks there couldn’t somebody use any struct returned by a native function to bypass any type checks under the current system?

runtian-zhou:
> Copying from Slack:
> 
> Actually, native functions might be more nontrivial than I initially expected — if we ignore type checks there couldn’t somebody use any struct returned by a native function to bypass any type checks under the current system?

That is correct. I don't think there's an easy way to capture this in the Move crate cuz we have no control over external native functions here. With that being said, I think it's the native implementer's responsibility to make sure values are tagged.

tnowacki (commented):
I imagine we will discuss this more thoroughly during our sync on Thursdays. 

But I wanted to at least put my initial thoughts here for others to think about:
- I worry about the invasiveness of this change in the `Value` stack. Even if the actual checks are behind a flag, this will still make changes more difficult and cumbersome in the future.
- I'm not sure we could go through with the changes to MoveStructLayout as this is consumed by a lot of external entities. I'd be worried about adding fields that only a given VM instance can understand.

> the unintended reinterprect cast/type conversion as we didn't check every single memory cell. 

Could you expand on your thoughts for this section, @runtian-zhou?
To me this reads as, "We need to re-implement the type safety checks at runtime in case the type safety checks were wrong". If so, this doesn't feel like it adds much value in the sense of what if the second, dynamic implementation was also wrong? Or was wrong while the type safety checks were correct?
In other words, this feels like it helps only in the very narrow case where the type safety rules were wrong but the reimplementation of them was correct. We have many different static rules, which similarly to type safety, have no tolerance to bugs. Adding this layer of dynamic type checking here does not seem to address the broader concern of "what if the bytecode verifier was wrong?" (and I'm not really sure that broader concern can be answered in any way other than we have zero room to be wrong in the first place)

runtian-zhou:
> * I worry about the invasiveness of this change in the `Value` stack. Even if the actual checks are behind a flag, this will still make changes more difficult and cumbersome in the future.

Maybe we can guard the change under a static feature flag as well?

> I'm not sure we could go through with the changes to MoveStructLayout as this is consumed by a lot of external entities. I'd be worried about adding fields that only a given VM instance can understand.

That I'd be less worried about. The reason is that we are adding a new variant so all the old MoveStructLayout should still be usable. 

> this feels like it helps only in the very narrow case where the type safety rules were wrong but the reimplementation of them was correct.

I totally agree with what you are saying. This PR is intended to add an extra hurdle to some of the most critical semantics of Move, aka resource safety, and only monitor those vulnerabilities. To me this is an extra redundancy to the system, because as we observed in the bug that was fixed, any of errors like this could result in severe impact to the entire Move ecosystem so I would suggest for some properties there's definitely need for a second eye on. To me the likelihood of us implementing the system wrong in both places is way smaller than the likelihood of us implementing just the verifier wrong. I get your concern of this system showing false positive but a false negative would be devastating in this case as anyone can break resource semantics in Move and can literally do anything in any Move based blockchain.

runtian-zhou:
> any way other than we have zero room to be wrong in the first place

I don't particularly think that this is safe to build a security first system. IMHO bugs are simply just inevitable and it's usually the redundancy of the system that prevent things are crashing, rather than making sure that our single point of failure is implemented flawlessly, as is proven in the aviation design :D

runtian-zhou:
> > any way other than we have zero room to be wrong in the first place
> 
> I don't particularly think that this is safe to build a security first system with this assumption. IMHO bugs are simply just inevitable and it's usually the redundancy of the system that prevent things are crashing, rather than making sure that our single point of failure is implemented flawlessly, as is proven in the aviation design :D

language/move-vm/types/src/values/values_impl.rs
vgao1996: > changing this could yield a bigger refactor

You will need to borrow the refcell in order to access the tag, but I don't think it will result in architectural changes. I'm concerned that if we end up adding more info/tags in the future, then it's gonna have a devastating effect on the value stack in terms of performance.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -67,6 +67,8 @@ pub(crate) struct Interpreter {
     operand_stack: Stack,
     /// The stack of active functions.
     call_stack: CallStack,
+    /// Whether to perform runtime type safety checks.
+    strict: bool,
```
wrwg, line 71: strict is somehow not the right notion for this. Both the verifier and the interpreter are strict. Perhaps called it `checked` here and elsewhere?

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -262,18 +265,45 @@ impl Interpreter {
...
+        loader: &Loader,
+    ) -> VMResult<Frame> {
+        let resolver = func.get_resolver(loader);
         let mut locals = Locals::new(func.local_count());
         let arg_count = func.arg_count();
         for i in 0..arg_count {
+            let v = self.operand_stack.pop().map_err(|e| self.set_location(e))?;
+            if self.strict {
+                let expected_ty = resolver
+                    .resolve_signature_token(&func.local_types()[arg_count - i - 1])
```
wrwg, line 281: This is rather expensive. It could be made better performing by precomputing or caching things here. We can also keep for now, but I don't think we want to turn this on in every node then.

language/move-vm/types/src/values/values_impl.rs
wrwg: Definitely check_type + the signature token resolution are rather concerning regards performance.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -262,18 +265,45 @@ impl Interpreter {
...
+                                PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
+                                    .with_message(
+                                        "Unexpected native function not located in a module"
+                                            .to_owned(),
+                                    );
+                            self.set_location(err)
+                        }
+                    })?;
+
+                v.check_type(&expected_ty)
```
wrwg, line 298: This is also rather expensive if we have to do this for every call. I would expect we have some cached type hash info for the expected arguments of this function.

wrwg:
> Copying from Slack:
> 
> Actually, native functions might be more nontrivial than I initially expected — if we ignore type checks there couldn’t somebody use any struct returned by a native function to bypass any type checks under the current system?

I don't know of a native function which returns a struct besides `from_bcs`. So this issue should be contained. I don't know what is with other chains.

chen-robert:
Well, the concern would be, if a malicious user is able to get a hold of any such untyped struct at runtime, you can bypass this mitigation.

language/move-vm/runtime/src/interpreter.rs
runtian-zhou: We should be able to cache this part relatively easily.

runtian-zhou:
> I don't know of a native function which returns a struct besides from_bcs. So this issue should be contained. I don't know what is with other chains.

The only other function in aptos side is `type_info`, which we should be able to hard code.

language/move-bytecode-verifier/src/verifier.rs
```diff
@@ -20,6 +20,7 @@ use move_binary_format::{
 pub struct VerifierConfig {
     pub max_loop_depth: Option<usize>,
     pub treat_friend_as_private: bool,
+    pub strict_runtime_verification: bool,
```
sblackshear, line 23: I agree with Wolfgang's earlier comment on naming--`strict` is the wrong word here, since it sounds like it's enforcing properties that are not enforced in non-strict mode. I would call this `redundant_runtime_verification` or `paranoid_runtime_verification` to make it very clear what the role of these checks are (a pragmatic failsafe against bugs in the verifier). This should also be explained in a comment on the config.

language/move-core/types/src/value.rs
```diff
@@ -69,6 +69,11 @@ impl MoveFieldLayout {
 pub enum MoveStructLayout {
     /// The representation used by the MoveVM
     Runtime(Vec<MoveTypeLayout>),
+    /// The representation used by the MoveVM with extra unique identifier for distinct type.
+    CheckedRuntime {
+        fields: Vec<MoveTypeLayout>,
+        tag: u64,
```
sblackshear, line 75: A comment explaining what this `u64` means would be useful. A descriptively named newtype would be even better.

language/move-bytecode-verifier/src/verifier.rs
runtian-zhou: I added it here just so that we don't need a new Config struct for VM.

vgao1996: We will need a config struct for the VM sooner or later, so feel free to just add one
```
struct VMConfig {
  pub verifier_config: VerifierConfig,
  pub strict_runtime_verification: bool,
}
```

language/move-vm/types/src/loaded_data/runtime_types.rs
sblackshear: I think it is critically important to avoid introducing nondeterminism here, either by using a hash of sufficient size or (better) using a deterministic identifier for a normalized type (e.g., an entry in the type cache). Just because an attacker is unlikely to be able to exploit the nondeterminism to break the type system doesn't mean that it's harmless. For example, the nondeterminism can make an innocent validator appear Byzantine, or allow an attack on liveness.

sblackshear (changes requested):

sblackshear:
Requesting changes for the nondeterminism issue. I understand this will be iterated on in the Aptos release branch, but wanted to share this view.

tnowacki:
> this PR is intended to add an extra hurdle to some of the most critical semantics of Move, aka resource safety, and only monitor those vulnerabilities

I would argue that semantics around use-after-move/drop and reference safety are just as important :) If you start reading random references that you shouldn't be able to, lots of things might break. Similarly if you can accidentally drop a resource. 
This is my point when I say we have zero room to be wrong. We have a large set of necessary static checks, and not all of these checks can be re-implemented dynamically. At least not easily, or without massive performance implications. 

This is what I mean by we are only guarding against a very narrow window of possible bugs here. 

If I understand the code correctly, it does not guard even against all possible type errors? 
Again, might be missing something, but if you do not also guard against type checking at ReadRef/WriteRef you can still do nasty things.
Imagine I had some local operation that let me treat some value `v: t` as `t'`. If I can then write `v` to a reference of type `&mut t'`, I can then bypass these function only guards. 
This specific issue could be resolved by checking ReadRef/WriteRef, but you might be better off deeply checking structs in `check_type`, i.e. recursively calling `check_type` on struct fields.  

To summarize
- I think these checks ignore vulnerabilities in other critical aspects of Move's set of static rules (private functions, not reading up the stack, private fields, no dangling references, reference's having referential transparency etc). 
- Without some checks around references and without recursively checking structs, I think the current set of checks is not sufficient at guarding against bugs in type safety.

tnowacki:
> Maybe we can guard the change under a static feature flag as well? Plus the ValueImpl is a complete private construct local to the vm_type crate so I don't quite see how it can make changes more difficult?

It makes changes to ValueImpl more difficult :) 
The currently value implementation is rather slow and inefficient. It would be nice long term to move to a more efficient (and likely less interpreted) value layout. 
Adding type tags in there makes these changes more difficult (though likely not impossible)

runtian-zhou:
Actually would love to understand from concern for nondeterminism issue better.  I think the concern here is this function might return false negative when we have an attack. False positive is not a concern here IIUC. The worst case scenario here is we indeed had an attack that causes this check to fail. However, some nodes passes the check because of the false positve and if we have more than 2f+1 node that has this hash collision, the network still alive even though we have this check. Is this your concern or is there anything else I missed here? cc @sblackshear

sblackshear:
> Actually would love to understand from concern for nondeterminism issue better. I think the concern here is this function might return false negative when we have an attack. False positive is not a concern here IIUC. The worst case scenario here is we indeed had an attack that causes this check to fail. However, some nodes passes the check because of the false positve and if we have more than 2f+1 node that has this hash collision, the network still alive even though we have this check. Is this your concern or is there anything else I missed here? cc @sblackshear

Any sort of nondeterminism is a potential problem for consensus--I think it's much safer to just avoid it altogether.

I think the bad scenario is (e.g.) when there's a transaction that triggers a collision in half of the validators, but not the other half.

jolestar:
>@tnowacki
> To summarize
> 
> * I think these checks ignore vulnerabilities in other critical aspects of Move's set of static rules (private functions, not reading up the stack, private fields, no dangling references, reference's having referential transparency etc).
> * Without some checks around references and without recursively checking structs, I think the current set of checks is not sufficient at guarding against bugs in type safety.

A similar question was discussed in the telegram group:

>@jolestar My main concern is how strong does our VM runtime defense mechanism need to be? Do we need to assume that all bytecode verifier conditions were skipped and the VM runtime is still strong enough to defend against exploits?

As I understand, the conclusion is that we can't assume that all verifiers fail because verifiers are also part of the VM, so what we can do is mainly strengthen verifiers, and Runtime checks are only a way to avoid the worst possible case.

language/move-vm/types/src/loaded_data/runtime_types.rs
jolestar: the logical probability of this hash being exploited should be very small, but introducing nondeterminism in a consensus system does feel weird, purely by feel.

runtian-zhou:
> I think the bad scenario is (e.g.) when there's a transaction that triggers a collision in half of the validators, but not the other half.

That would be bad. However isn't the intention here is we need to put the network on halt if such error happens? I guess the non-determinism comes only from the possibility of hash collision using not cryptographic safe functions. Will changing it to the safe functions address the concern here?

runtian-zhou:
> * Without some checks around references and without recursively checking structs, I think the current set of checks is not sufficient at guarding against bugs in type safety.

That's a great observation! Ideally it would be nice if we can check all types to be well formed at creation time. However due to the limited time scope that aptos have, we cut corner on checking the fields of a pack instruction, and we suggest that this is enough for us to isolate the type confusion to only impact its own module but not going across module boundaries.

With that being said, I think all we miss on type safety here is just to check the argument types for pack instruction. Once that's there I think we should be good on the type safety of the paranoid mode. I implemented the checks for write_ref already!

runtian-zhou:
The current status of the PR is aptos has landed this change into its own branch. In the meantime I will try to formulate the safety properties of what this PR is trying to achieve and add tests to test how this paranoid mode can identify potential attacks.

wrwg:
We have landed this in the `aptos` branch and will likely resume the discussion soon for the `main` branch. Closing this PR for now.

PR #499, opened 2022-09-24
Trying to get proptest work

wrwg:
## Motivation

(Write your answer here.)

## Test Plan

### DIFF
--- a/language/move-binary-format/src/deserializer.rs
+++ b/language/move-binary-format/src/deserializer.rs
@@ -65,6 +65,15 @@ fn read_u64_internal(cursor: &mut VersionedCursor) -> BinaryLoaderResult<u64> {
     Ok(u64::from_le_bytes(u64_bytes))
 }
 
+fn read_u64_internal_bounded(cursor: &mut VersionedCursor, max: u64) -> BinaryLoaderResult<u64> {
+    let val = read_u64_internal(cursor)?;
+    if val > max {
+        return Err(PartialVMError::new(StatusCode::MALFORMED)
+            .with_message("u64 greater than max requested".to_string()));
+    }
+    Ok(val)
+}
+
 fn read_u128_internal(cursor: &mut VersionedCursor) -> BinaryLoaderResult<u128> {
     let mut u128_bytes = [0; 16];
     cursor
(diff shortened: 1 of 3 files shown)


runtian-zhou (commented):
Looks good to me. Are we concerned about the potential new behavior introduced by the code? I think this should never be triggered but would love to hear your thoughts @wrwg

language/move-binary-format/src/deserializer.rs
```diff
@@ -1471,17 +1480,19 @@ fn load_code(cursor: &mut VersionedCursor, code: &mut Vec<Bytecode>) -> BinaryLo
...
             Opcodes::VEC_IMM_BORROW => Bytecode::VecImmBorrow(load_signature_index(cursor)?),
             Opcodes::VEC_MUT_BORROW => Bytecode::VecMutBorrow(load_signature_index(cursor)?),
             Opcodes::VEC_PUSH_BACK => Bytecode::VecPushBack(load_signature_index(cursor)?),
             Opcodes::VEC_POP_BACK => Bytecode::VecPopBack(load_signature_index(cursor)?),
-            Opcodes::VEC_UNPACK => {
-                Bytecode::VecUnpack(load_signature_index(cursor)?, read_u64_internal(cursor)?)
-            }
+            Opcodes::VEC_UNPACK => Bytecode::VecUnpack(
+                load_signature_index(cursor)?,
+                read_u64_internal_bounded(cursor, VEC_PACK_UNPACK_MAX)?,
```
sblackshear, line 1494: Does this mean that the instruction `VecPack[VEC_PACK_UNPACK_MAX + 1]` could be included in published code + serialized, but would fail to deserialize (assuming you could get the published code to pass the verifier)?

wrwg:
> Looks good to me. Are we concerned about the potential new behavior introduced by the code? I think this should never be triggered but would love to hear your thoughts @wrwg

I'm sharing this because I can't [fix the failing proptest](https://github.com/move-language/move/actions/runs/3116907822/jobs/5055155937) in this PR ;-) No idea what to do. Related to our discussion that I think proptest is a technical debt for us.

jiangying000:
> > Looks good to me. Are we concerned about the potential new behavior introduced by the code? I think this should never be triggered but would love to hear your thoughts @wrwg
> 
> I'm sharing this because I can't [fix the failing proptest](https://github.com/move-language/move/actions/runs/3116907822/jobs/5055155937) in this PR ;-) No idea what to do and looking for help. Related to our discussion that I think proptest is a technical debt for us.

does this mean you intend to get rid of all the `proptest! {`, or just improve and fix the proptest failures ?

PR #501, opened 2022-09-25
[compatibility] Make struct removal violate layout compatibility

wrwg:
## Motivation

Until now, when a struct was removed, this only counted as a `struct_and_function_linkage` compatibility issue. However, it is also a `struct_layout` issue because it opens the door to first remove a struct, then re-add it with a different layout.

Yes

## Test Plan

Existing

### DIFF
--- a/language/move-binary-format/src/compatibility.rs
+++ b/language/move-binary-format/src/compatibility.rs
@@ -19,7 +19,8 @@ use std::collections::BTreeSet;
 pub struct Compatibility {
     /// If false, dependent modules that reference functions or structs in this module may not link
     pub struct_and_function_linking: bool,
-    /// If false, attempting to read structs previously published by this module will fail at runtime
+    /// If false, attempting to read structs previously published by this module will fail at
+    /// runtime, or worse, may allow to re-interpret representations maliciously.
     pub struct_layout: bool,
 }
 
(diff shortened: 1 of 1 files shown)


sblackshear (approved):
Very good catch!

We need to come up with a more direct way to test these. But in the meantime, would you mind adding an expected value test for this scenario (https://github.com/move-language/move/tree/main/language/tools/move-cli/tests/sandbox_tests/republish is a helpful example)?

I recall that there use to be a lot more expected value tests for the compatibility logic, I wonder if some of them were part of the `diem-cli` testsuite and got deleted...

wrwg:
> Very good catch!

Credits go to @chen-robert from OtterSec.

> 
> We need to come up with a more direct way to test these. But in the meantime, would you mind adding an expected value test for this scenario (https://github.com/move-language/move/tree/main/language/tools/move-cli/tests/sandbox_tests/republish is a helpful example)?
> 
> I recall that there use to be a lot more expected value tests for the compatibility logic, I wonder if some of them were part of the `diem-cli` testsuite and got deleted...

Yes those tests are (to some extent) in the Aptos repo.

However, I saw @jolestar has been writing independently some tests. If I land this could you afterwards add the tests, @jolestar? Transactional tests seem the right thing to do here. (Not acquainted myself with that infra.)

chen-robert (commented):
looks good to me -- would also note that there are probably many other examples of this issue in these compatibility checks, will need some more time to review on our side.

jolestar (approved):

jolestar:
> However, I saw @jolestar has been writing independently some tests. If I land this could you afterwards add the tests, @jolestar? Transactional tests seem the right thing to do here. (Not acquainted myself with that infra.)

Yes. I will continue the #500. This PR land should break the test case https://github.com/move-language/move/pull/500/files#diff-78d81d81673c1eeafe39de4b8007dae712660f9e869e24e8c8721c0f8ecd3c31 .

language/move-binary-format/src/compatibility.rs
```diff
@@ -50,17 +51,10 @@ impl Compatibility {
...
-                    // this struct `Parent::T`. That is, code like the function `foo` in
-                    // ```
-                    // struct S { t: Parent::T }
-                    // public fun foo(a: addr): S { move_from<S>(addr) }
-                    // ```
-                    // in module `Child` will continue to run without error. But values of type
-                    // `Parent::T` in `Child` are now "orphaned" in the sense that `Parent` no
-                    // longer exposes any API for reading/writing them.
+                    struct_layout = false;
                     continue;
```
zsluedem, line 58: Just a nitpick. It could `break` or even return the result directly here.

PR #507, opened 2022-09-27
[vm & test] Migrate the treat_friend_as_private to CompatibilityConfig and add tests

jolestar:
## Motivation

1. `treat_friend_as_private` as CompatibilityConfig and rename it to `check_friend_linking` as @sblackshear suggestion, because it only affect the module publish Compat verify.
2. Add `--skip-check-friend-linking` flag to the transactional test `publish` command.
3. Add tests for `--skip-check-friend-linking`

Yes

## Test Plan

CI

### DIFF
--- a/language/move-vm/runtime/src/runtime.rs
+++ b/language/move-vm/runtime/src/runtime.rs
@@ -38,7 +38,6 @@ use tracing::warn;
 /// An instantiation of the MoveVM.
 pub(crate) struct VMRuntime {
     loader: Loader,
-    verifier_config: VerifierConfig,
 }
 
 impl VMRuntime {
@@ -47,8 +46,7 @@ impl VMRuntime {
         verifier_config: VerifierConfig,
     ) -> PartialVMResult<Self> {
         Ok(VMRuntime {
-            loader: Loader::new(NativeFunctions::new(natives)?, verifier_config.clone()),
-            verifier_config,
+            loader: Loader::new(NativeFunctions::new(natives)?, verifier_config),
         })
     }
 
@@ -120,11 +118,8 @@ impl VMRuntime {
                 let old_module = old_module_ref.module();
                 let old_m = normalized::Module::new(old_module);
                 let new_m = normalized::Module::new(module);
-                let compat = Compatibility::check(
-                    self.verifier_config.treat_friend_as_private,
-                    &old_m,
-                    &new_m,
-                );
+                let compat =
+                    Compatibility::check(compat_config.check_friend_linking, &old_m, &new_m);
 
                 if compat_config.check_struct_and_function_linking
                     && !compat.struct_and_function_linking
--- a/language/move-binary-format/src/compatibility.rs
+++ b/language/move-binary-format/src/compatibility.rs
@@ -26,9 +26,13 @@ pub struct Compatibility {
 
 #[derive(Debug, Clone, Copy)]
 pub struct CompatibilityConfig {
+    /// if false, do not ensure the dependent modules that reference functions or structs in this module can link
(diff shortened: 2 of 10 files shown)


language/move-vm/runtime/src/runtime.rs
```diff
@@ -38,7 +38,6 @@ use tracing::warn;
 /// An instantiation of the MoveVM.
 pub(crate) struct VMRuntime {
     loader: Loader,
-    verifier_config: VerifierConfig,
```
sblackshear, line 41: Excellent work getting rid of this!

language/move-vm/runtime/src/runtime.rs
```diff
@@ -120,11 +118,8 @@ impl VMRuntime {
                 let old_module = old_module_ref.module();
                 let old_m = normalized::Module::new(old_module);
                 let new_m = normalized::Module::new(module);
-                let compat = Compatibility::check(
-                    self.verifier_config.treat_friend_as_private,
-                    &old_m,
-                    &new_m,
-                );
+                let compat =
+                    Compatibility::check(compat_config.check_friend_linking, &old_m, &new_m);
```
sblackshear, line 122: Double-checking: should this be `!compat_config.check_friend_linking`, given that we've flipped the meaning of the boolean?

sblackshear: In addition, am wondering: it seems like `CompatibilityConfig::check(&self, old_module, new_module) -> Compatibility` makes more sense as an API than `Compatibility::check(bool, old_module, new_module) -> Compatibility`. Thoughts?

jolestar: > Double-checking: should this be `!compat_config.check_friend_linking`, given that we've flipped the meaning of the boolean?

I also change `Compatibility::check(treat_friend_as_private` to `Compatibility::check(check_friend_linking`. Keep them consistent. https://github.com/move-language/move/pull/507/files#diff-70ffa03682eab402f7a5b28174bb83750c2ba76fab85d04e56846ecb22c6e319L72-R79

jolestar: > In addition, am wondering: it seems like `CompatibilityConfig::check(&self, old_module, new_module) -> Compatibility` makes more sense as an API than `Compatibility::check(bool, old_module, new_module) -> Compatibility`. Thoughts?

Agree, I also want to do it, but afraid of changing too much in one PR.

And the check function should include the result check logic?

```rust
 if compat_config.check_struct_and_function_linking
                    && !compat.struct_and_function_linking
{
    return Err(PartialVMError::new(
        StatusCode::BACKWARD_INCOMPATIBLE_MODULE_UPDATE,
    )
    .finish(Location::Undefined));
}
if compat_config.check_struct_layout && !compat.struct_layout {
    return Err(PartialVMError::new(
        StatusCode::BACKWARD_INCOMPATIBLE_MODULE_UPDATE,
    )
    .finish(Location::Undefined));
}
```

https://github.com/jolestar/move/blob/0d254bcec10d1da705bf254c82b8ac84ed4c2154/language/move-vm/runtime/src/runtime.rs#L124-L137

So it should be `CompatibilityConfig::check(&self, old_module, new_module) -> Result<()>`?

sblackshear (approved):

jolestar:
Just rebase the origin/main to trigger the CI rerun.

language/move-vm/runtime/src/runtime.rs
jolestar: Do this refactor in https://github.com/move-language/move/pull/559

PR #509, opened 2022-09-27
[move-vm] Add strict/paranoid runtime verification

runtian-zhou:
## Motivation

This is a duplicate of #493 . The intention is we will first land this version on Aptos branch while we discuss what would be the right runtime checks to add for main branch.

Yes

## Test Plan

`cargo test`

### DIFF
--- a/language/move-vm/runtime/src/interpreter.rs
+++ b/language/move-vm/runtime/src/interpreter.rs
@@ -4,13 +4,14 @@
 
 use crate::{
     loader::{Function, Loader, Resolver},
+    move_vm::RuntimeConfig,
     native_functions::NativeContext,
     trace,
 };
 use fail::fail_point;
 use move_binary_format::{
     errors::*,
-    file_format::{Bytecode, FunctionHandleIndex, FunctionInstantiationIndex},
+    file_format::{AbilitySet, Bytecode, FunctionHandleIndex, FunctionInstantiationIndex},
 };
 use move_core_types::{
     account_address::AccountAddress,
@@ -67,6 +68,8 @@ pub(crate) struct Interpreter {
     operand_stack: Stack,
     /// The stack of active functions.
     call_stack: CallStack,
+    /// Whether to perform safety checks at runtime.
+    runtime_config: RuntimeConfig,
 }
 
 struct TypeWithLoader<'a, 'b> {
@@ -94,7 +97,7 @@ impl Interpreter {
     ) -> VMResult<Vec<Value>> {
         // We count the intrinsic cost of the transaction here, since that needs to also cover the
         // setup of the function.
-        let mut interp = Self::new();
+        let mut interp = Self::new(loader.runtime_config());
         interp.execute(
             loader, data_store, gas_meter, extensions, function, ty_args, args,
         )
@@ -102,10 +105,11 @@ impl Interpreter {
 
     /// Create a new instance of an `Interpreter` in the context of a transaction with a
     /// given module cache and gas schedule.
-    fn new() -> Self {
+    fn new(runtime_config: RuntimeConfig) -> Self {
         Interpreter {
             operand_stack: Stack::new(),
             call_stack: CallStack::new(),
+            runtime_config,
         }
     }
 
@@ -145,6 +149,17 @@ impl Interpreter {
         ty_args: Vec<Type>,
         args: Vec<Value>,
     ) -> VMResult<Vec<Value>> {
+        // A counter for tracking numbers of struct that don't have any capability.
+        // This counter will be incremented whenever a struct that has empty capability was created by pack and decremented on unpack.
+        //
+        // The invariant here is this counter should always goes to zero at the end of execution.
+        // The reason is that values with such capability can only be created by pack instruction and destructed by unpack
+        // instruction and can never be stored in storage.
+        //
+        // Thus if the counter is not zero towards the end of execution, it means that some hot potato values are silently dropped at runtime.
+
+        let mut hot_potato_counter: u64 = 0;
+
         let mut locals = Locals::new(function.local_count());
         for (i, value) in args.into_iter().enumerate() {
             locals
@@ -152,11 +167,17 @@ impl Interpreter {
                 .map_err(|e| self.set_location(e))?;
         }
 
-        let mut current_frame = Frame::new(function, ty_args, locals);
+        let mut current_frame = Frame::new(function, ty_args, locals, self.runtime_config);
         loop {
             let resolver = current_frame.resolver(loader);
             let exit_code = current_frame //self
-                .execute_code(&resolver, self, data_store, gas_meter)
+                .execute_code(
+                    &resolver,
+                    self,
+                    data_store,
+                    gas_meter,
+                    &mut hot_potato_counter,
+                )
                 .map_err(|err| self.maybe_core_dump(err, &current_frame))?;
             match exit_code {
                 ExitCode::Return => {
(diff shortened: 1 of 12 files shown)


wrwg:
Please change the target where you want to merge into to `aptos`. There should be dropdown at the top of your PR. Once you do the diff will be against that branch.

wrwg:
@chen-robert FYI

runtian-zhou:
The remaining work items are:
1. Get benchmarks from @vgao1996 
2. Force annotating the return value of native functions.

chen-robert:
@runtian-zhou some additional ideas for checks which I think would make sense to get merged in here
1. Function visibility check
2. Check `ty_args` against function type parameter visibility requirements
3. Hot potato `Pack` and `Unpack` tracking

language/move-bytecode-verifier/src/verifier.rs
```diff
@@ -20,6 +20,7 @@ use move_binary_format::{
 pub struct VerifierConfig {
     pub max_loop_depth: Option<usize>,
     pub treat_friend_as_private: bool,
+    pub strict_runtime_verification: bool,
```
wrwg, line 23: As discussed on the other PR lets not call this strict... There is really an association with 'strict' vs 'non-strict' which is a misfit here.

The name suggested by Sam, `paranoid_checks`, is much better and reflects what this feature is really about in the overall Move architecture.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -67,6 +67,8 @@ pub(crate) struct Interpreter {
     operand_stack: Stack,
     /// The stack of active functions.
     call_stack: CallStack,
+    /// Whether to perform runtime type safety checks.
+    check_runtime_types: bool,
```
wrwg, line 71: Should use the same name as above. Also it is not just type checks.

wrwg (commented):
Looks mostly ready.

runtian-zhou:
Right now the two tests are failing:
- language/move-prover/interpreter-testsuite/tests/concrete_check/pack_unpack.move
- language/move-prover/interpreter-testsuite/tests/concrete_check/destroy.move

IMHO the failure here are as expected because the test function is returning struct has zero capabilities and thus causing the struct to be dropped unexpectedly...

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -148,6 +148,17 @@ impl Interpreter {
         ty_args: Vec<Type>,
         args: Vec<Value>,
     ) -> VMResult<Vec<Value>> {
+        // A counter for tracking numbers of struct that don't have any capability.
+        // This counter will be incremented whenever a struct that has empty capability was created by pack and decremented.
+        //
```
wrwg, line 153: ... decremented on unpack

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -148,6 +148,17 @@ impl Interpreter {
...
+        // The invariant here is this counter should always goes to zero at the end of execution.
+        // The reason is that values with such capability can only be created by pack instruction and destructed by unpack
+        // instruction and can never be stored in storage.
+        //
+        // Thus if the counter is not zero towards the end of execution, it means that some hot potato values are silently dropped at runtime.
+
+        let mut hot_potato_counter: u64 = 0;
```
wrwg, line 160: Its fine that we do just this for now, but a more central problem of hot potatoes -- at least in the mind of the Move team when this feature was developed -- is store, not drop. Anyway good to see this.

language/move-vm/runtime/src/loader.rs
```diff
@@ -1376,6 +1376,23 @@ impl<'a> Resolver<'a> {
         Type::Struct(struct_def)
     }
 
+    pub(crate) fn get_struct_ability(&self, idx: StructDefinitionIndex) -> AbilitySet {
```
wrwg, line 1379: This is again very inefficient... Lets keep in mind to optimize

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -1082,6 +1114,14 @@ impl Frame {
                             interpreter.operand_stack.last_n(field_count as usize)?,
                         )?;
                         let args = interpreter.operand_stack.popn(field_count)?;
+
+                        if self.paranoid_checks {
```
wrwg, line 1118: Wondering whether we should have a different flag, so if this starts to produce false positives (the risk here higher as with type checks), we don't have to turn everthing off.

wrwg (commented):
For hot potato checking

language/move-vm/runtime/src/interpreter.rs
runtian-zhou: Key and copy can both break the invariant here: struct has key can be stored directly into storage, struct has copy can be copied and thus causing more unpack than pack.

language/move-vm/runtime/src/loader.rs
runtian-zhou: This doesn't look very inefficient to me? Just reading out a value from a vector?

wrwg (approved):

language/move-vm/runtime/src/interpreter.rs
runtian-zhou: Added a different flag for this.

jolestar:
I like this solution more than #493, I think this solution will make me sleep a little more soundly😂. Expect it to be merged into the main branch.

sblackshear:
To expand on the issue @tnowacki was mentioning on the other PR with an example:

```
let bad_signer = ... // turn bytes into a signer via exploiting a vulnerable instruction (e.g., VecPack). This will still have a vector<u8> type tag, so the attacker must use it carefully
let vec = vector::empty();
vector::push_back(&mut vec, bad_signer); // note: this is a VecPushBack bytecode instruction, not a function call--no checks are performed
let any = any::pack(vec) // this will check the type of `vec`, but not `vec[0]`
let laundered_vec = any::unpack(any); // now vec[0] has the `signer` tag, can be used freely
```

runtian-zhou:
> To expand on the issue @tnowacki was mentioning on the other PR with an example:
> 
> ```move
> let bad_signer = ... // turn bytes into a signer via exploiting a vulnerable instruction (e.g., VecPack). This will still have a vector<u8> type tag, so the attacker must use it carefully
> let vec = vector::empty();
> vector::push_back(&mut vec, bad_signer); // note: this is a VecPushBack bytecode instruction, not a function call--no checks are performed
> let any = any::pack(vec) // this will check the type of `vec`, but not `vec[0]`
> let laundered_vec = any::unpack(any); // now vec[0] has the `signer` tag, can be used freely
> ```

So if `any` is the function in `aptos-framework` it would be fine because we check the inner fields of struct recursively at serialization time.

If we check all type matches the declaration of Pack, we would be able to argue that in the MoveVM with paranoid mode on, no malformed value can ever be created. However, due to the limited timeline that we have I intensionally cut that corner but I would still argue that this would offer us enough safety. In this version, a module can create a struct that has bad field elements in it. But there's no way to exploit from this fact from a practical standpoint given that we check the struct types to match at function call as well.

wrwg:
What is going on with this PR? I thought it was already ready to land the day before yesterday. @davidiw

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -1811,6 +1945,164 @@ impl Vector {
...
+            | (ValueImpl::U64(_), Type::U64)
+            | (ValueImpl::Bool(_), Type::Bool)
+            | (ValueImpl::Address(_), Type::Address) => Ok(()),
+            (ValueImpl::IndexedRef(r), Type::Reference(ty) | Type::MutableReference(ty)) => {
+                r.check_type(ty)
+            }
+            (ValueImpl::ContainerRef(r), Type::Reference(ty) | Type::MutableReference(ty)) => {
+                r.container().check_type(ty)
+            }
+            (ValueImpl::Container(c), ty) => c.check_type(ty),
```
vgao1996, line 1998: Is this correct? Why are we shadowing `ty` here?

vgao1996 (approved):
Stamping to get us unblocked, but I share similar concerns with @sblackshear that this only seems to cover some of the type safety invariants and it's a bit hard to reason what exact subset being covered is

PR #510, opened 2022-09-27
[interpreter] audit operand stack usage

sblackshear:
- Add paranoid check to protect the VM against a native function that return a number of values inconsistent with its declared return types. The verifier can't help us with this; we are at the mercy of native function implementers
- Remove a redundant wrapper function and have `execute_main` take ownership of the `Interpreter` to ensure that it is used correctly
- Add comments about where operand stack sharing happens
- Remove one out-of-date comment about the create_account bytecode!

### DIFF
--- a/language/move-vm/runtime/src/interpreter.rs
+++ b/language/move-vm/runtime/src/interpreter.rs
@@ -92,37 +92,11 @@ impl Interpreter {
         extensions: &mut NativeContextExtensions,
         loader: &Loader,
     ) -> VMResult<Vec<Value>> {
-        // We count the intrinsic cost of the transaction here, since that needs to also cover the
-        // setup of the function.
-        let mut interp = Self::new();
-        interp.execute(
-            loader, data_store, gas_meter, extensions, function, ty_args, args,
-        )
-    }
-
-    /// Create a new instance of an `Interpreter` in the context of a transaction with a
-    /// given module cache and gas schedule.
-    fn new() -> Self {
         Interpreter {
             operand_stack: Stack::new(),
             call_stack: CallStack::new(),
         }
-    }
-
-    /// Internal execution entry point.
-    fn execute(
-        &mut self,
-        loader: &Loader,
-        data_store: &mut impl DataStore,
-        gas_meter: &mut impl GasMeter,
-        extensions: &mut NativeContextExtensions,
-        function: Arc<Function>,
-        ty_args: Vec<Type>,
-        args: Vec<Value>,
-    ) -> VMResult<Vec<Value>> {
-        // No unwinding of the call stack and value stack need to be done here -- the context will
-        // take care of that.
-        self.execute_main(
+        .execute_main(
             loader, data_store, gas_meter, extensions, function, ty_args, args,
         )
     }
@@ -133,10 +107,8 @@ impl Interpreter {
     /// function represented by the frame. Control comes back to this function on return or
     /// on call. When that happens the frame is changes to a new one (call) or to the one
     /// at the top of the stack (return). If the call stack is empty execution is completed.
-    // REVIEW: create account will be removed in favor of a native function (no opcode) and
-    // we can simplify this code quite a bit.
     fn execute_main(
-        &mut self,
+        mut self,
         loader: &Loader,
         data_store: &mut impl DataStore,
         gas_meter: &mut impl GasMeter,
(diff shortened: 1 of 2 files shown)


language/move-vm/runtime/src/interpreter.rs
```diff
@@ -109,32 +106,13 @@ impl Interpreter {
...
     ///
     /// This function sets up a `Frame` and calls `execute_code_unit` to execute code of the
     /// function represented by the frame. Control comes back to this function on return or
     /// on call. When that happens the frame is changes to a new one (call) or to the one
     /// at the top of the stack (return). If the call stack is empty execution is completed.
-    // REVIEW: create account will be removed in favor of a native function (no opcode) and
-    // we can simplify this code quite a bit.
+    /// This should only be called directly from `entrypoint()`, and `self` should not be used after it returns
     fn execute_main(
         &mut self,
```
tnowacki, line 117: Based on the comment above, should this be `fn execute_main(self,` ?

sblackshear: Good point, I can refactor to that.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -330,6 +307,19 @@ impl Interpreter {
         let return_values = result
             .result
             .map_err(|code| PartialVMError::new(StatusCode::ABORTED).with_sub_status(code))?;
+        // Paranoid check to protect us against incorrect native function implementations. A native function that
+        // returns a different number of values than its declared types will trigger this check
+        if return_values.len() != return_type_count {
+            return Err(
+                PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR).with_message(
```
wrwg, line 314: Those internal invariant violation errors of the VM are problematic because they do not lead to `Keep`, and thus no gas charge and potential source of DoS. Since this is kind of the same category as a verifier error (only later and dynamic) perhaps NUMBER_OF_ARGUMENTS_MISMATCH or CONSTRAINT_VIOLATION which leads to `Keep`.

wrwg (approved):

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -330,6 +307,19 @@ impl Interpreter {
         let return_values = result
             .result
             .map_err(|code| PartialVMError::new(StatusCode::ABORTED).with_sub_status(code))?;
+        // Paranoid check to protect us against incorrect native function implementations. A native function that
+        // returns a different number of values than its declared types will trigger this check
+        if return_values.len() != return_type_count {
```
chen-robert, line 312: While we're at it, does it make sense to do some light type-checks of return parameters here as well?

language/move-vm/runtime/src/interpreter.rs
tnowacki: > Those internal invariant violation errors of the VM are problematic because they do not lead to Keep, and thus no gas charge and potential source of DoS.

I know that VMStatus is in the Move repo, but it is adapter specific. We do not use it on the SUI side. 
Nonetheless, we would likely also not charge for this invariant violation ;) 
I'm not sure what else to do here though? It is not clear if you hit this if it is an attack or a VM bug.

jolestar: Only the framework can define `native function`, so I think it is impossible to use this Err for DoS.

sblackshear: I think the attack Wolfgang is concerned about is that if there's a bug in a framework native function that triggers this, that is a potential DOS vector.

sblackshear:
Going to go ahead and land this, but please feel free to open a follow-up if you want a different error code.

language/move-vm/runtime/src/interpreter.rs
chen-robert: I was thinking primitive / struct types -- although do agree that it's not _super_ effective. For example, validating that a function which return `u64` does indeed return `u64`

language/move-vm/runtime/src/interpreter.rs
jolestar: Got it. But if we `Keep` the transaction that triggers this ERR, will make it more difficult for us to fix the bugs. We need to ensure that the transactions are executed on the bug-fixed VM with the same result.

language/move-vm/runtime/src/interpreter.rs
runtian-zhou: I added them in #509!

PR #511, opened 2022-09-28
[rfc][interpreter] give each stack frame its own operand stack

sblackshear:
In the current implementation of the VM, the operand stack is shared by all stack frames. However, the stack balancing pass of the verifier performs checks that prevent a callee frame from touching operand stack values that belong to the caller. Thus, each call stack frame conceptually has its own private operand stack, and an implementation that moves the operand stack into a call stack frame should work just fine.

This PR tries out such an implementation, mostly out of curiosity/sanity-checking that the verifier pass does indeed ensure what we think it does (at least on our e2e and smoke tests). It adds (redundant, paranoid) checks at function returns that ensure the number of values remaining on the callee operand stack is the same as the number of return types in the callee.

These redundant checks/this refactoring should be cheap (modulo extra allocations compared to a shared operand stack, which we should watch out for). But it has several advantages over the shared operand stack implementation in terms of defense in depth.

- A bug in the stack balancing verifier pass that allows the callee to reach into the caller's stack frame (aka the "Darion bug" from back in the day) will be caught at runtime--it will result in popping an empty stack in the callee and hitting a VM_INVARIANT_VIOLATION as a result
- A buggy native function that accidentally returns too many or too few values will *not* be caught by the verifier. But it will be caught at runtime by the paranoid checks. This seems like a fairly easy mistake for a native function implementer to make, particularly in refactoring.
- A bug in the type safety verifier pass that messes up the return type/return value arity will now be caught at runtime by the paranoid checks

I am putting this up as an experiment and a RFC to see if folks think the approach has value, but this would obviously need much more discussion and some perf measurement before going in.

### DIFF
--- a/language/move-vm/runtime/src/interpreter.rs
+++ b/language/move-vm/runtime/src/interpreter.rs
@@ -63,8 +63,6 @@ macro_rules! set_err_info {
 /// An `Interpreter` instance is a stand alone execution context for a function.
 /// It mimics execution on a single thread, with an call stack and an operand stack.
 pub(crate) struct Interpreter {
-    /// Operand stack, where Move `Value`s are stored for stack operations.
-    operand_stack: Stack,
     /// The stack of active functions.
     call_stack: CallStack,
 }
@@ -104,9 +102,18 @@ impl Interpreter {
     /// given module cache and gas schedule.
     fn new() -> Self {
         Interpreter {
-            operand_stack: Stack::new(),
             call_stack: CallStack::new(),
         }
+    }
+
+    /// Get a mutable reference to the operand stack for the currently executing function
+    fn operand_stack(&mut self) -> &mut Stack {
+        self.call_stack.operand_stack()
+    }
+
+    /// Get a reference to the operand stack for the currently executing function
+    fn operand_stack_immut(&self) -> &Stack {
+        self.call_stack.operand_stack_immut()
     }
 
     /// Internal execution entry point.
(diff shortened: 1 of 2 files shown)


sblackshear:
This implementation approach is broken; ignore the tests and think of this more of an exploration of the idea :)

tnowacki (commented):
I have not looked at this too closely yet, so just a conceptual comment really (sorry)

Just wanted to initially comment we should be careful here not to overcorrect too too much. I like the idea of the API ensuring that you don't go outside of the current frame, but... function calls are already the slowest part of the VM, by a lot. I want to make sure that we aren't making that problem even worse.

tnowacki:
> A buggy native function that accidentally returns too many or too few values will not be caught by the verifier. But it will be caught at runtime by the paranoid checks. This seems like a fairly easy mistake for a native function implementer to make, particularly in refactoring.

I'm wondering if we should just do a bunch of paranoid checks like this for native functions in debug builds. We are really just targeting ourselves here and could build out a nice little set of sanity checks for those cases.

runtian-zhou:
So IIUC what we are doing here is giving each function call its own operand stack to prevent overflow/underflow across function calls? How are you thinking about the possibility of imbalanced stack because of control flow instruction inside one function?

sblackshear:
> So IIUC what we are doing here is giving each function call its own operand stack to prevent overflow/underflow across function calls

Overflow and underflow are separate issues that are already caught by existing dynamic checks for popping an empty operand stack and creating an operand stack bigger than the max (respectively). This change protects us against the scenarios mentioned in the summary:

> A bug in the stack balancing verifier pass that allows the callee to reach into the caller's stack frame (aka the "Darion bug" from back in the day) will be caught at runtime--it will result in popping an empty stack in the callee and hitting a VM_INVARIANT_VIOLATION as a result

>  A bug in the type safety verifier pass that messes up the return type/return value arity will now be caught at runtime by the paranoid checks

sblackshear:
> Just wanted to initially comment we should be careful here not to overcorrect too too much. I like the idea of the API ensuring that you don't go outside of the current frame, but... function calls are already the slowest part of the VM, by a lot. I want to make sure that we aren't making that problem even worse.

I agree with this. Asymptotically, this shouldn't be slower, but allocating an operand stack per function call instead of sharing big one seems likely to cause a slowdown. I would definitely want to measure this before considering this change. The idea of this PR was mostly as a sanity check that this check doesn't actually fail on our tests, but it was harder to make the PR work than I expected :).

wrwg:
I have not checked the details of the current implementation, but are we actually copying parameters for a function call? If not this justifies a single operand stack, otherwise I find it technically questionable and most likely a performance issue, because this large vector will not nicely live in the L1/2 caches. Allocating and freeing smaller pieces should be more efficient and perhaps give LLVM even the option to optimize some of them away.

sblackshear:
Definitely-- we would have to measure the performance impact of this and (if negative) weigh it against the safety benefits this more paranoid implementation buys (same as in the other paranoid checks). These tradeoffs are very subjective/interesting!

Abandoning this for now because my first cut at this approach was harder to get compiling then I thought.

PR #520, opened 2022-09-30
[Prover] Bump up the version number of Boogie

junkil-park:
- Upgrade Boogie to 2.15.8.

## Motivation

Upgrade the Prover's dependency

yes

## Test Plan

cargo test


meng-xu-cs:
There are known non-determinism in these test cases:

```
    prover unit[default]::functional/loops_with_memory_ops.move
    prover unit[default]::functional/loops.move
```

I guess updating Boogie / Z3 accidentally causes the random number to go from one side to the other side (i.e., concretizing variable `x` vs variable `y`). Maybe we should disable them or re-write them into deterministic way?

wrwg:
Let's wait with this until we know whether the new boogie fixes our problem. If so, this PR should also increase the minimal boogie number required in the prover driver.

wrwg (approved):
Can we cherry pick this in the `aptos` branch? Otherwise we wont have it in the `aptos` CLI.

Let me know if its unclear how to do this.

junkil-park:
> Can we cherry pick this in the `aptos` branch? Otherwise we wont have it in the `aptos` CLI.
> 
> Let me know if its unclear how to do this.

@wrwg , I put up a PR to cherry pick this: https://github.com/move-language/move/pull/535

PR #525, opened 2022-10-02
Make TypeTag enum use less memory

zekun000:
Currently a single TypeTag consumes 96 bytes because of the inlined StructTag (88 bytes), it starts to bloat up with nested TypeTag::Vector(TypeTag).

This commit reduces the size from 96 to 16 bytes, it doesn't fully solve the problem but reduces the impact by 6x.


dariorussi (approved):
this is a good change, no question; bloating of enums is always a problem in Rust and the dereference here I believe is marginal. In fact this is likely to make things faster.

Failures look legit and this most likely needs to wait for [526](https://github.com/move-language/move/pull/526) and so will change again

vgao1996 (approved):
LGTM

PR #526, opened 2022-10-02
[types] make type tag nesting limited

davidiw:
There's limited value and a lot of wasted resources by leaving this limited to serde's limits. Recursion is dangerous.

## Motivation

(Write your answer here.)

## Test Plan

### DIFF
--- a/language/move-core/types/src/language_storage.rs
+++ b/language/move-core/types/src/language_storage.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 #[cfg(any(test, feature = "fuzzing"))]
 use proptest_derive::Arbitrary;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, Serializer};
 use std::{
     fmt::{Display, Formatter},
     str::FromStr,
@@ -18,10 +18,12 @@ use std::{
 pub const CODE_TAG: u8 = 0;
 pub const RESOURCE_TAG: u8 = 1;
 
+const MAX_TYPE_TAG_NESTING: u8 = 8;
+
 /// Hex address: 0x1
 pub const CORE_CODE_ADDRESS: AccountAddress = AccountAddress::ONE;
 
-#[derive(Serialize, Deserialize, Debug, PartialEq, Hash, Eq, Clone, PartialOrd, Ord)]
+#[derive(Deserialize, Debug, PartialEq, Hash, Eq, Clone, PartialOrd, Ord)]
 pub enum TypeTag {
     // alias for compatibility with old json serialized data.
     #[serde(rename = "bool", alias = "Bool")]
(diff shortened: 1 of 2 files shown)


gregnazario (approved):

language/move-core/types/src/language_storage.rs
```diff
@@ -219,4 +297,47 @@ mod tests {
...
+    }
+
+    fn make_type_tag_vector(type_param: TypeTag) -> TypeTag {
+        TypeTag::Vector(Box::new(type_param))
+    }
+
+    fn make_type_tag_struct(type_param: TypeTag) -> TypeTag {
+        TypeTag::Struct(StructTag {
+            address: AccountAddress::ONE,
+            module: Identifier::from_utf8(("a".as_bytes()).to_vec()).unwrap(),
```
dariorussi, line 338: just for future reference and I can only imagine your "care factor" here, but if you want to save your beautiful fingers a bit more you can do `Identifier::new("a").unwrap()`.
I know, I know....

language/move-core/types/src/language_storage.rs
```diff
@@ -42,6 +44,82 @@ pub enum TypeTag {
...
+                validate_type_tag_recursion(vec![(**type_tag).clone()].as_slice(), 1)
+            }
+            TypeTag::Struct(value) => validate_type_tag_recursion(&value.type_params, 1),
+            _ => Ok(()),
+        }
+        .map_err(serde::ser::Error::custom)?;
+
+        serializer.serialize_newtype_struct("TypeTag", &SafeTypeTag::from(self))
+    }
+}
```
dariorussi, line 98: so let me understand this and see if this is what you intended.
Here we check that the top level struct is properly nested (up to max) and then we go on to serialize `SafeTypeTag` which contains a `StructTag` and we reenter this function for `n-1` levels recursively for all nested level, which are obviously correct because we already checked the top level.
That seems a bit too much?
I understand the alternatives may be a bit on the annoying side of life (though not sure, are they?) but asking to see if that was the intention or not.
Of course given the check we reenter at most sum(1..8) so not a big deal... let me know if you want this in anyway

language/move-core/types/src/language_storage.rs
davidiw: copypasta :)

language/move-core/types/src/language_storage.rs
davidiw: So the issue here is that even if I build out a proper custom serializer, I will inevitably be left playing a very miserable game of unwrap:

```
TypeTag::Vector(ref value) => serializer.serialize_newtype_variant("TypeTag", 6, "vector", &value)
```

which if value is a typetag as well, then it will just end up trying to serialize it within this method again
same with struct tag containing a typetag

I've made a complete wrapper around it, so that we only go into the recursion three times, once to count, once to convert, and once to serialize. I'd prefer not to do further optimizations at this point, because we still have deserialization. We could theoretically remove one of those by doing serialization time conversions.

davidiw:
@dariorussi I've updated the PR to be a bit more optimal.

dariorussi (approved):
fwiw I would have been fine with the other one as well given how small is the number of recursion and the importance of stopping crazy recursion on crazy nested types.
I just wanted to make sure that was the intent.
Love and cheese brother

davidiw:
> fwiw I would have been fine with the other one as well given how small is the number of recursion and the importance of stopping crazy recursion on crazy nested types. I just wanted to make sure that was the intent. Love and cheese brother

Hahaha, next time, next time

PR #527, opened 2022-10-02
[natives][loader] New API for fully annotated MoveTypeLayout

tnowacki:
- Added API that generates a MoveTypeLayout from a loaded Type, but adds all struct tag information
- Exposed new loader API in the native context
- Exposed the function in the session for completeness

## Motivation

- I have some native function code for tests that fully walks a MoveValue, but to do so, I really need the type information for the struct. To get this, I need a MoveTypeLayout that uses MoveStructLayout::WithTypes instead of Runtime
- Happy to put this somewhere other than the loader (the caching of the results is nice but not super important). Maybe @vgao1996 will have some ideas

## Test Plan

- Using it on another branch, and will be used extensively (so we will effectively have end-to-end tests for this function).
- But we should probably have tests for this... just a bit hard at the moment as none of these functions have their own test suite

### DIFF
--- a/language/move-vm/runtime/src/loader.rs
+++ b/language/move-vm/runtime/src/loader.rs
@@ -25,7 +25,7 @@ use move_core_types::{
     identifier::{IdentStr, Identifier},
     language_storage::{ModuleId, StructTag, TypeTag},
     metadata::Metadata,
-    value::{MoveStructLayout, MoveTypeLayout},
+    value::{MoveFieldLayout, MoveStructLayout, MoveTypeLayout},
     vm_status::StatusCode,
 };
 use move_vm_types::{
@@ -217,12 +217,20 @@ impl ModuleCache {
         idx: StructDefinitionIndex,
     ) -> StructType {
         let struct_handle = module.struct_handle_at(struct_def.struct_handle);
+        let field_names = match &struct_def.field_information {
+            StructFieldInformation::Native => vec![],
+            StructFieldInformation::Declared(field_info) => field_info
+                .iter()
+                .map(|f| module.identifier_at(f.name).to_owned())
+                .collect(),
+        };
         let abilities = struct_handle.abilities;
         let name = module.identifier_at(struct_handle.name).to_owned();
         let type_parameters = struct_handle.type_parameters.clone();
         let module = module.self_id();
         StructType {
             fields: vec![],
+            field_names,
             abilities,
             type_parameters,
             name,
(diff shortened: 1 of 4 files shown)


language/move-vm/runtime/src/loader.rs
```diff
@@ -2155,13 +2170,15 @@ struct FieldInstantiation {
...
     struct_layout: Option<MoveStructLayout>,
+    annotated_struct_layout: Option<MoveStructLayout>,
 }
 
 impl StructInfo {
     fn new() -> Self {
         Self {
             struct_tag: None,
             struct_layout: None,
+            annotated_struct_layout: None,
```
tnowacki, line 2181: I'm not super thrilled by this new field, but I think #525 might make the impact a lot smaller

language/move-vm/types/src/loaded_data/runtime_types.rs
```diff
@@ -16,6 +16,7 @@ pub const TYPE_DEPTH_MAX: usize = 256;
 #[derive(Debug, Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub struct StructType {
     pub fields: Vec<Type>,
+    pub field_names: Vec<Identifier>,
```
tnowacki, line 19: Similarly a bit annoyed by this addition, but I didn't see a way around it

vgao1996:
Sounds like a reasonable approach to the problem you're trying to solve... I could imagine us offering some kind of traversal API in the future so we can say goodbye to all of these layouts, but we can work on that later.

I'll take a closer look tomorrow at some of the details

language/move-vm/types/src/loaded_data/runtime_types.rs
vgao1996: I think this is fine.

vgao1996 (approved):
The MoveStructLayout is becoming more and more of an abomination. Let's think about alternative solutions that can address its use cases...

tnowacki:
> The MoveStructLayout is becoming more and more of an abomination. Let's think about alternative solutions that can address its use cases...

Agreed. I'm wondering if we basically want to split it into a runtime layout version, used internally by the VM, and an annotated/decorated version, which is probably more useful for external tools that don't have a full view of the world to rely on (like the VM sort of does internally)

We've been using `MoveTypeLayout::WithTypes` a lot on Sui for more or less annotating the layout of the struct and then syncing that with some sort of Rust type (with the same layout). Worth considering this sort of paradigm when thinking about a more externally facing API vs one used by the VM

PR #528, opened 2022-10-02
[types] limit parsing of nested TypeTag

davidiw:
Limit parsing of nested type tags in the type parser

### DIFF
--- a/language/move-core/types/src/parser.rs
+++ b/language/move-core/types/src/parser.rs
@@ -254,7 +254,11 @@ impl<I: Iterator<Item = Token>> Parser<I> {
         })
     }
 
-    fn parse_type_tag(&mut self) -> Result<TypeTag> {
+    fn parse_type_tag(&mut self, depth: u8) -> Result<TypeTag> {
+        if depth >= crate::safe_serialize::MAX_TYPE_TAG_NESTING {
+            bail!("Exceeded TypeTag nesting limit during parsing: {}", depth);
+        }
+
         Ok(match self.next()? {
             Token::U8Type => TypeTag::U8,
             Token::U64Type => TypeTag::U64,
@@ -264,7 +268,7 @@ impl<I: Iterator<Item = Token>> Parser<I> {
             Token::SignerType => TypeTag::Signer,
             Token::VectorType => {
                 self.consume(Token::Lt)?;
-                let ty = self.parse_type_tag()?;
+                let ty = self.parse_type_tag(depth + 1)?;
                 self.consume(Token::Gt)?;
                 TypeTag::Vector(Box::new(ty))
             }
@@ -278,7 +282,7 @@ impl<I: Iterator<Item = Token>> Parser<I> {
                                 let ty_args = if self.peek() == Some(&Token::Lt) {
                                     self.next()?;
                                     let ty_args = self.parse_comma_list(
-                                        |parser| parser.parse_type_tag(),
+                                        |parser| parser.parse_type_tag(depth + 1),
                                         Token::Gt,
                                         true,
                                     )?;
@@ -343,12 +347,12 @@ pub fn parse_string_list(s: &str) -> Result<Vec<String>> {
 
 pub fn parse_type_tags(s: &str) -> Result<Vec<TypeTag>> {
     parse(s, |parser| {
-        parser.parse_comma_list(|parser| parser.parse_type_tag(), Token::EOF, true)
+        parser.parse_comma_list(|parser| parser.parse_type_tag(0), Token::EOF, true)
     })
 }
 
 pub fn parse_type_tag(s: &str) -> Result<TypeTag> {
-    parse(s, |parser| parser.parse_type_tag())
+    parse(s, |parser| parser.parse_type_tag(0))
 }
 
 pub fn parse_transaction_arguments(s: &str) -> Result<Vec<TransactionArgument>> {
@@ -366,7 +370,7 @@ pub fn parse_transaction_argument(s: &str) -> Result<TransactionArgument> {
 }
 
 pub fn parse_struct_tag(s: &str) -> Result<StructTag> {
-    let type_tag = parse(s, |parser| parser.parse_type_tag())
+    let type_tag = parse(s, |parser| parser.parse_type_tag(0))
         .map_err(|e| format_err!("invalid struct tag: {}, {}", s, e))?;
     if let TypeTag::Struct(struct_tag) = type_tag {
         Ok(*struct_tag)
@@ -465,6 +469,7 @@ mod tests {
             "bool",
             "vector<u8>",
             "vector<vector<u64>>",
+            "vector<vector<vector<vector<vector<vector<vector<u64>>>>>>>",
             "signer",
             "0x1::M::S",
             "0x2::M::S_",
@@ -480,6 +485,14 @@ mod tests {
         ] {
             assert!(parse_type_tag(s).is_ok(), "Failed to parse tag {}", s);
         }
+
+        let s =
+            "vector<vector<vector<vector<vector<vector<vector<vector<vector<vector<u64>>>>>>>>>>";
+        assert!(
+            parse_type_tag(s).is_err(),
+            "Should have failed to parse type tag {}",
+            s
+        );
     }
 
     #[test]
@@ -505,6 +518,7 @@ mod tests {
             "0x1::Diem::Diem<u8 , bool  ,    vector<u8>,address,signer>",
             "0x1::Diem::Diem<vector<0x1::Diem::Struct<0x1::XUS::XUS>>>",
             "0x1::Diem::Diem<0x1::Diem::Struct<vector<0x1::XUS::XUS>, 0x1::Diem::Diem<vector<0x1::Diem::Struct<0x1::XUS::XUS>>>>>",
+            "0x1::Diem::Diem<vector<0x1::XDX::XDX<vector<vector<vector<0x1::XDX::XDX<vector<u64>>>>>>>>",
         ];
         for text in valid {
             let st = parse_struct_tag(text).expect("valid StructTag");
@@ -516,5 +530,12 @@ mod tests {
                 st
             );
         }
+
+        let s = "0x1::Diem::Diem<vector<0x1::XDX::XDX<vector<vector<vector<0x1::XDX::XDX<vector<vector<u64>>>>>>>>>";
+        assert!(
+            parse_struct_tag(s).is_err(),
+            "Should have failed to parse type tag {}",
+            s
+        );
     }
 }


language/move-core/types/src/parser.rs
```diff
@@ -480,6 +485,13 @@ mod tests {
         ] {
             assert!(parse_type_tag(s).is_ok(), "Failed to parse tag {}", s);
         }
+
+        let s = "vector<vector<vector<vector<vector<vector<vector<vector<u64>>>>>>>>";
+        assert!(
+            parse_type_tag(s).is_err(),
+            "Should have failed to parse type tag {}",
+            s
+        );
```
gregnazario, line 494: Btw there's a test above for "failed" type tags

language/move-core/types/src/parser.rs
```diff
@@ -516,5 +529,12 @@ mod tests {
                 st
             );
         }
+
+        let s = "0x1::Diem::Diem<vector<0x1::XDX::XDX<vector<vector<vector<0x1::XDX::XDX<vector<u64>>>>>>>>";
+        assert!(
+            parse_struct_tag(s).is_err(),
```
gregnazario, line 535: Similarly

language/move-core/types/src/parser.rs
```diff
@@ -254,7 +254,11 @@ impl<I: Iterator<Item = Token>> Parser<I> {
         })
     }
 
-    fn parse_type_tag(&mut self) -> Result<TypeTag> {
+    fn parse_type_tag(&mut self, depth: u8) -> Result<TypeTag> {
+        if depth >= crate::language_storage::MAX_TYPE_TAG_NESTING {
+            bail!("Hit TypeTag nesting limit during parsing: {}", depth);
```
gregnazario, line 259: `bail!("TypeTag has more than {} nested types: {}", MAX_TYPE_TAG_NESTING, depth) `

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,145 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
     }
 }
 
+struct TypeTagVisitor(u8);
```
gregnazario, line 106: Maybe should comment that u8 is `depth` or make a struct with `u8` named as depth

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,145 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
...
+                variant.newtype_variant_seed(NestedTypeTag(self.0))?,
+            ))),
+            DeTypeTag::Struct => Ok(TypeTag::Struct(
+                variant.newtype_variant_seed(NestedStructTag(self.0))?,
+            )),
+        }
+    }
+}
+
+struct NestedTypeTag(u8);
```
gregnazario, line 159: Similarly on the depth

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,145 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
...
+
+impl<'de> DeserializeSeed<'de> for NestedTypeTag {
+    type Value = TypeTag;
+
+    fn deserialize<D>(self, deserializer: D) -> Result<Self::Value, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        println!("tt {}", self.0);
```
gregnazario, line 168: Please remove the println!

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,145 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
...
+        if self.0 >= MAX_DE_TYPE_TAG_NESTING {
+            return Err(de::Error::custom(format!(
+                "Hit TypeTag nesting limit during deserialization: {}",
```
gregnazario, line 171: Similarly, I would change the phrasing here.  `Hit` isn't necessarily a very portable wording.

`"TypeTag has more than {} nested types during deserialization: {}"`

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,145 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
...
+impl<'de> DeserializeSeed<'de> for NestedVecTypeTag {
+    type Value = Vec<TypeTag>;
+
+    fn deserialize<D>(self, deserializer: D) -> Result<Self::Value, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        if self.0 >= MAX_DE_TYPE_TAG_NESTING {
+            return Err(de::Error::custom(format!(
+                "Hit TypeTag nesting limit during deserialization: {}",
```
gregnazario, line 212: Similarly

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,145 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
...
+
+use serde::de::{
+    self, DeserializeSeed, Deserializer, EnumAccess, MapAccess, SeqAccess, VariantAccess, Visitor,
+};
+
+impl<'de> Visitor<'de> for TypeTagVisitor {
+    type Value = TypeTag;
+
+    fn expecting(&self, formatter: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
+        formatter.write_str("`bool`, `u8`, `u64`, `u128`, `address`, `signer`, `vector`, `struct`")
```
gregnazario, line 116: you could probably just print the `TYPE_TAG_VARIANTS` here to simplify

language/move-core/types/src/parser.rs
davidiw: that's transaction arguments

language/move-core/types/src/parser.rs
davidiw: that's transaction arguments

davidiw (commented):
addressing other comments...

gregnazario (approved):

language/move-core/types/src/language_storage.rs
```diff
@@ -18,34 +18,58 @@ use std::{
 pub const CODE_TAG: u8 = 0;
 pub const RESOURCE_TAG: u8 = 1;
 
-const MAX_TYPE_TAG_NESTING: u8 = 8;
+pub const MAX_TYPE_TAG_NESTING: u8 = 8;
+
+#[cfg(test)]
+const MAX_DE_TYPE_TAG_NESTING: u8 = MAX_TYPE_TAG_NESTING - 1;
```
igor-aptos, line 24: to confirm, this is there just so we can check that both serialization and deserialization fails, correct?

wanna add a comment for that?

igor-aptos (commented):
looks good to me, don't have approval rights for this repo :)

dariorussi:
quick question: when do you plan/want this to get in?
this is really something I would love @vgao1996 to review.
That can happen after too but I really think he has the best context to evaluate this.
In any case review coming soon....

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,153 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
     }
 }
 
+// u8 -> depth or nesting level
+struct TypeTagVisitor(u8);
+
+use serde::de::{
```
wrwg, line 109: Could you please ensure to format before sending? I was spending a while wondering where your Visitor and other types come from, as I did not see them imported on the top of the file (where expected).

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,153 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
...
+    self, DeserializeSeed, Deserializer, EnumAccess, MapAccess, SeqAccess, VariantAccess, Visitor,
+};
+
+impl<'de> Visitor<'de> for TypeTagVisitor {
```
wrwg, line 113: I would be nice to see some commenting here, like

```
// Manual implementation of what serde would derive automatically, but with an injected bound. We are using this 
// here as currently no configurable type depth is supported by serde. See [link if there is something to see].
```
Or whatever makes sense. Otherwise, someone seeing this code here would just declare us as insane.

Also as you seem to have reverse engineered Serde, it would be nice you give some hints so others don't need to do the same as far as this is possible.

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,153 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
...
+            formatter.write_fmt(format_args!(", `{}`", value))?;
+        }
+        Ok(())
+    }
+
+    fn visit_enum<A>(self, data: A) -> Result<Self::Value, A::Error>
+    where
+        A: EnumAccess<'de>,
+    {
+        let (variant_type, variant) = data.variant::<DeTypeTag>()?;
```
wrwg, line 128: How do we get a match here of `DeTypeTag` vs the actual type? By position in the variant list of `enum`? If so it would be helpful if that has been documented with a short sentence at `DeTypeTag`.

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,153 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
...
+        match variant_type {
+            DeTypeTag::Bool => {
+                variant.unit_variant()?;
```
wrwg, line 131: What is this? Short comment would help the reader/reviewer the first time this appears. Not everyone knows the details of serde. I have my guesses, but...

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,153 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
...
+            DeTypeTag::Address => {
+                variant.unit_variant()?;
+                Ok(TypeTag::Address)
+            }
+            DeTypeTag::Signer => {
+                variant.unit_variant()?;
+                Ok(TypeTag::Signer)
+            }
+            DeTypeTag::Vector => Ok(TypeTag::Vector(Box::new(
+                variant.newtype_variant_seed(NestedTypeTag(self.0))?,
```
wrwg, line 155: Why not just make `self.0` be `self.nest` or something.

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,153 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
...
+struct NestedTypeTag(u8);
+
+impl<'de> DeserializeSeed<'de> for NestedTypeTag {
+    type Value = TypeTag;
+
+    fn deserialize<D>(self, deserializer: D) -> Result<Self::Value, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        if self.0 >= MAX_DE_TYPE_TAG_NESTING {
```
wrwg, line 174: Perhaps move the condition and the error generation into some `check_nest(self.0)?` as it is used more than once.

dariorussi (approved):
it looks good to me but @vgao1996 please review as you have a chance tomorrow

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,153 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
...
+            return Err(de::Error::custom(format!(
+                "Exceeded TypeTag nesting limit during deserialization: {}",
+                self.0
+            )));
+        }
+
+        deserializer.deserialize_seq(VecTypeTagVisitor(self.0))
```
wrwg, line 224: `self.0 + 1`?

If this was actually omitted (and after you already had a review), consider what this means for in-depth defense. Those also have bugs.

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,153 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
...
+
+    fn deserialize<D>(self, deserializer: D) -> Result<Self::Value, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        deserializer.deserialize_struct("StructTag", STRUCT_TAG_FIELDS, StructTagVisitor(self.0))
+    }
+}
+
+const TYPE_TAG_VARIANTS: &[&str] = &["bool", "u8", "u64", "u128", "signer", "vector", "struct"];
```
wrwg, line 242: Move the const decls to the front? This was used up somewhere already IIRC.

language/move-core/types/src/language_storage.rs
```diff
@@ -79,6 +103,153 @@ impl<'a> From<&'a TypeTag> for SafeTypeTag<'a> {
...
+        }
+
+        deserializer.deserialize_seq(VecTypeTagVisitor(self.0))
+    }
+}
+
+// u8 -> depth or nesting level
+struct NestedStructTag(u8);
+
+impl<'de> DeserializeSeed<'de> for NestedStructTag {
```
wrwg, line 231: Move this closer to the struct stuff?

wrwg (commented):
That's an awesome hack and re-engineering of parts of serde! However, I'm bit concerned about the complexity here (and the possibility of new bugs). Lets discuss offline.

language/move-core/types/src/language_storage.rs
davidiw: this is how I keep my job...

language/move-core/types/src/language_storage.rs
davidiw: this one is a bit of a mess, the documentation is rather clear about this, but the lack of examples and the "step function" to arrive at how this works is certainly NOT OBVIOUS

language/move-core/types/src/language_storage.rs
davidiw: I hate extra code :)

language/move-core/types/src/language_storage.rs
davidiw: It is used by two different objects.

language/move-core/types/src/language_storage.rs
davidiw: at this point, we haven't really picked up a new type tag. the type tag will be incremented for each actual TypeTag we visit.

davidiw: in other words... comment please

language/move-core/types/src/language_storage.rs
davidiw: I suspect one could bikeshed a lot about the organization of this file. I might argue putting the public types at the top is more readable. I could also see an argument about relational placement.

wrwg:
I've a potential alternative (simpler) solution. It involves thread locals. I hope I can send for comparison before EOD.

davidiw:
I'd like to move fast here. If we can come up with a generalizable scheme, I'm all game. I do think that this is a rather safe PR and something that would definitely be beneficial upstream. It is a mess that serde doesn't support limited nesting at a struct level.

wrwg (approved):

PR #529, opened 2022-10-03
[types] limit nesting of types via state

wrwg:
## Motivation

This is an alternative version of what #528 and the previous PR for limitation of nesting of type tags provides.

Yes

## Test Plan

### DIFF
--- a/language/move-core/types/src/language_storage.rs
+++ b/language/move-core/types/src/language_storage.rs
@@ -6,6 +6,7 @@ use crate::{
     account_address::AccountAddress,
     identifier::{IdentStr, Identifier},
     parser::{parse_struct_tag, parse_type_tag},
+    safe_serialize,
 };
 #[cfg(any(test, feature = "fuzzing"))]
 use proptest_derive::Arbitrary;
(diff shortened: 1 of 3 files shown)


dariorussi:
I'll just leave a note:
this is pretty cool and reminds me of back in the days when we had a custom function to serialize/deserialize Move Values and it was few lines of code and easy to understand. The visitor pattern of serde is pretty difficult to follow.
Later there was this push to make everything "pure serde" and that conversation/train was impossible to stop.

The use of thread local storage is difficult to evaluate for me. And of course I wish (as I am sure we all do) that it was easy to pass a custom value on the stack, basically what the visitor pattern allows but without the insanity of the visitor pattern. Anyway, enough of wishful thinking...

With nesting being a global property, a TLS seems fine, not sure that will always be true and that is one place where it could be difficult. 
With leaf types (as TypeTag or Move Values) that is also fine, I think.

runtian-zhou:
Do we need to land this change on aptos branch? Also will this be a substitute for #525 and #526 on aptos branch?

sblackshear:
This solution seems reasonable to me. Couple of questions about the motivation: I'm assuming that this exists to prevent stack overflows when serializing or deserializing `TypeTag`'s. Is the concern about:
1. Converting `SignatureToken`'s in the bytecode format into `TypeTag`'s (e.g., to write a value into global storage)
2. Accepting user-specified `TypeTag`'s as tx inputs (e.g., `entry fun pay<T>(amount: u64)`)
3. Accepting user-specified `TypeTags` in read API's (e.g., to simulate the result of a function call)

I'm guessing all of the above, but wanted to know which surfaces I might be overlooking. Also, if it's just (1) (or if we want to prevent runtime failures in the case of (1)), we could think about enforcing this in the file format or verifier.

wrwg:
> This solution seems reasonable to me. Couple of questions about the motivation: I'm assuming that this exists to prevent stack overflows when serializing or deserializing `TypeTag`'s. Is the concern about:
> 
> 1. Converting `SignatureToken`'s in the bytecode format into `TypeTag`'s (e.g., to write a value into global storage)
> 2. Accepting user-specified `TypeTag`'s as tx inputs (e.g., `entry fun pay<T>(amount: u64)`)
> 3. Accepting user-specified `TypeTags` in read API's (e.g., to simulate the result of a function call)
> 
> ? (I'm guessing all of the above, but wanted to know which I am wrong about + which surfaces I might be overlooking). Also, if it's just (1) (or if we want to prevent runtime failures in the case of (1)), we could think about enforcing this in the file format or verifier.

This is not just about stackoverflows. We can discuss offline. Closing this one, as we have landed this now (with some mods) in aptos branch. Will send another cherrypick PR later.

PR #531, opened 2022-10-03
[gas] pass more info to the gas meter when calling functions

vgao1996:
This passes some additional info to the gas meter when calling functions to make gas metering more robust
- Number of locals
- (Natives only) Views into the return values

### DIFF
--- a/language/move-vm/runtime/src/interpreter.rs
+++ b/language/move-vm/runtime/src/interpreter.rs
@@ -214,6 +214,7 @@ impl Interpreter {
                             self.operand_stack
                                 .last_n(func.arg_count())
                                 .map_err(|e| set_err_info!(current_frame, e))?,
+                            (func.local_count() as u64).into(),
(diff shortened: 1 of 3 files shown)


tnowacki (commented):
cc @oxade for visibility

vgao1996:
@tnowacki yeah right now this ls coming to the aptos branch, but I plan to merge it to main soon.

dariorussi (approved):
this looks reasonable, can you articulate a bit more what the idea is behind the number of locals?
I assume that the return value of a native function is because in a move function the gas charging would be at creation time whereas for a native function we do not see the create?
In that respect the return value of a native function matters only when the value is moved back, is that fair?

vgao1996:
@dariorussi when we make a new call frame we allocate the space for those locals -- we simply need to charge for doing the work.

wrwg (approved):

PR #532, opened 2022-10-03
[types] limit nesting of types via state

wrwg:
Cherry picked from #529.

## Motivation

This is an alternative version of what #528 and the previous for limitation of nesting provides, having a significant lower footprint.

Yes

## Test Plan

### DIFF
--- a/language/move-core/types/src/language_storage.rs
+++ b/language/move-core/types/src/language_storage.rs
@@ -6,10 +6,11 @@ use crate::{
     account_address::AccountAddress,
     identifier::{IdentStr, Identifier},
     parser::{parse_struct_tag, parse_type_tag},
+    safe_serialize,
 };
 #[cfg(any(test, feature = "fuzzing"))]
 use proptest_derive::Arbitrary;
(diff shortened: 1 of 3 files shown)


davidiw (approved):
unblocking with the caveat we should discuss with upstream to decide on whether or not to embrace the serde or the thread local approach

runtian-zhou:
This seems reasonable to me. Let's try to push this upstream.

dariorussi (approved):

dariorussi:
is this conflicting with @zekun000 commit?

wrwg:
> is this conflicting with @zekun000 commit?

Fixed fixed fixed indeed

wrwg:
> unblocking with the caveat we should discuss with upstream to decide on whether or not to embrace the serde or the thread local approach

We need to land one of those today, so the upstream discussion we can postpone, but my feeling is smaller code impact likely easier to accept. Specifically as we expecting that this is supported by Serde. (By which I mean we should work with the Serde author to support this and perhaps send a PR his direction.)

PR #533, opened 2022-10-03
[bytecode-verifier] fix an off-by-one error

runtian-zhou:
## Motivation

This is a duplicate of #524. Just to make sure the changes on the aptos branch could be included in upstream as much as possible if the community agrees.

There is an off-by-one error in the assertion on 
`assert!(pc <= u16::max_value() - 2 && (pc as usize) < code.len())`

While the RHS, i.e., `(pc as usize) < code.len()` is correct, the LHS has the off-by-one issue.

The deserializer guarantees that code unit length is at max `u16::MAX`, i.e., `code.len() <= u16::MAX`. Therefore, it is possible to construct a code unit with length `u16::MAX` which is filled by `u16::MAX - 1` `nop` operations and one `ret` operation, as shown in the added test case `test_max_number_of_bytecode`.

In this test case:
- The whole `nop + ret` code unit will be considered as a single basic block and the bytecode verifier, when running control-flow sensitive passes, will attempt to get the successor of this basic block, and will hence call the `get_successor` function.
- When calling the function, the `pc` value is guaranteed to be pointing to the last `nop`, right before the `ret`, which is at offset `u16::MAX - 1`. This is in violation of the wrongful assertion that `pc <= u16::max_value() - 2`. This is a proof that the LHS of the assertion is wrong.

In fact, it may be even argued that the `pc < u16::MAX` part in the assertion is redundant.

Yes

## Test Plan

Added test

### DIFF
--- a/language/move-binary-format/src/file_format.rs
+++ b/language/move-binary-format/src/file_format.rs
@@ -1672,6 +1672,12 @@ impl Bytecode {
             pc < u16::MAX && (pc as usize) < code.len(),
             "Program counter out of bounds"
         );
+
+        // Return early to prevent overflow if pc is hiting the end of max number of instructions allowed (u16::MAX).
+        if pc > u16::max_value() - 2 {
+            return vec![];
+        }
+
         let bytecode = &code[pc as usize];
         let mut v = vec![];
 
--- a/language/move-binary-format/src/file_format.rs
+++ b/language/move-binary-format/src/file_format.rs
@@ -1668,9 +1668,8 @@ impl Bytecode {
     /// Return the successor offsets of this bytecode instruction.
     pub fn get_successors(pc: CodeOffset, code: &[Bytecode]) -> Vec<CodeOffset> {
         assert!(
-            // The program counter could be added to at most twice and must remain
-            // within the bounds of the code.
-            pc <= u16::max_value() - 2 && (pc as usize) < code.len(),
+            // The program counter must remain within the bounds of the code
+            pc < u16::MAX && (pc as usize) < code.len(),
             "Program counter out of bounds"
         );
         let bytecode = &code[pc as usize];
--- a/language/move-binary-format/src/unit_tests/binary_tests.rs
+++ b/language/move-binary-format/src/unit_tests/binary_tests.rs
@@ -2,7 +2,7 @@
 // Copyright (c) The Move Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use crate::file_format_common::*;
+use crate::{file_format::Bytecode, file_format_common::*};
 use proptest::prelude::*;
 
 #[test]
@@ -12,6 +12,18 @@ fn binary_len() {
         binary_data.push(1).unwrap();
     }
     assert_eq!(binary_data.len(), 100);
+}
+
+#[test]
+fn test_max_number_of_bytecode() {
+    let mut nops = vec![];
+    for _ in 0..u16::MAX - 1 {
+        nops.push(Bytecode::Nop);
+    }
+    nops.push(Bytecode::Ret);
+
+    let result = Bytecode::get_successors(u16::MAX - 1, &nops);
+    assert!(result.is_empty());
 }
 
 proptest! {
--- a/language/move-bytecode-verifier/bytecode-verifier-tests/src/unit_tests/code_unit_tests.rs
+++ b/language/move-bytecode-verifier/bytecode-verifier-tests/src/unit_tests/code_unit_tests.rs
@@ -58,3 +58,16 @@ fn valid_fallthrough_abort() {
     let result = CodeUnitVerifier::verify_module(&Default::default(), &module);
     assert!(result.is_ok());
 }
+
+#[test]
+fn test_max_number_of_bytecode() {
+    let mut nops = vec![];
+    for _ in 0..u16::MAX - 1 {
+        nops.push(Bytecode::Nop);
+    }
+    nops.push(Bytecode::Ret);
+    let module = dummy_procedure_module(nops);
+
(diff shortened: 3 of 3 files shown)


language/move-binary-format/src/file_format.rs
```diff
@@ -1668,9 +1668,8 @@ impl Bytecode {
     /// Return the successor offsets of this bytecode instruction.
     pub fn get_successors(pc: CodeOffset, code: &[Bytecode]) -> Vec<CodeOffset> {
         assert!(
-            // The program counter could be added to at most twice and must remain
-            // within the bounds of the code.
-            pc <= u16::max_value() - 2 && (pc as usize) < code.len(),
+            // The program counter must remain within the bounds of the code
+            pc < u16::MAX && (pc as usize) < code.len(),
```
sblackshear, line 1672: Won't we get an overflow at 1682 (`let next_pc = pc + 1`) or 1689 `v.push(pc + 1)` without the `- 2`?

I see that you are trying to test this indirectly by invoking the verifier below, but I think it would be better to directly test `get_succesors` with (e.g.) `pc = u16::MAX - 1`.

sblackshear:
I think the PR title should make it clear that you're changing behavior (by tweaking a check)--"update outdated comments and add a new test case" sounds like there's no semantic change.

meng-xu-cs:
@sblackshear PR title and description updated.

language/move-binary-format/src/file_format.rs
sblackshear: Got it, thanks for the detailed explanation!

Definitely important to fix the panic, but I am wondering if we can do it in a more future-proof way. I buy the argument that the usage of `get_successors` in `VMControlGraph::new` is safe, but because `get_successors` is a public function, it's easy to imagine it being called in a different context where the same assumptions don't hold (e.g., it also gets called from `tools/move-coverage/src/summary.rs`, could be called in a future verifier pass). Couple of ideas on this, but no strong prefs:

- Move `get_successors` inside of `VMControlFlowGraph` and make it private
- Move `get_successors` inside of `VMControlFlowGraph` and check `is_end_of_block` as a precondition
- Have `get_successors` return an empty vec if `pc > u16::max_value() - 2`

runtian-zhou:
I think we need to discuss how we should release features like this in the long run. It could be imaginable that some day in the future there will be some Move based blockchain running live and theoretically changes like this could be a breaking change for some network.

sblackshear:
> I think we need to discuss how we should release features like this in the long run. It could be imaginable that some day in the future there will be some Move based blockchain running live and theoretically changes like this could be a breaking change for some network.

I'm not sure I understand this comment. This is (IIUC) not a new feature, it is a fixing a crash bug that should not have previously been there and not changing any behavior on the happy paths. Where is the breaking change?

runtian-zhou:
> > I think we need to discuss how we should release features like this in the long run. It could be imaginable that some day in the future there will be some Move based blockchain running live and theoretically changes like this could be a breaking change for some network.
> 
> I'm not sure I understand this comment. This is (IIUC) not a new feature, it is a fixing a crash bug that should not have previously been there and not changing any behavior on the happy paths. Where is the breaking change?

Sry I was just worrying about the sth different that's not quite relevant to the nature of this PR. IMHO there are two types of changes to the bytecode verifier:
1. Relaxing the bytecode verifier to accept code that it used to reject (This PR is one example of this case).
2. Adding more restriction for the bytecode verifier to reject code that it used to accept. (This one is a breaking change)

IMHO the second case could be a lot more intricate in the future when there are multiple live networks and we need to figure out how to roll out changes like (2) over time. Particularly if there's anything that needs to be landed ASAP, as we've experienced in the previous bug.

sblackshear (changes requested):
Requesting changes for one of the mitigations in https://github.com/move-language/move/pull/533/files#r987122347 (or something else that makes it safe to call `get_successors` from a different context)

runtian-zhou:
Pushed the suggested fix by @sblackshear

wrwg:
> Pushed the suggested fix by @sblackshear

Can't see the commit? Also no force update?

runtian-zhou:
Sry pushed to the wrong branch

language/move-binary-format/src/file_format.rs
```diff
@@ -1668,11 +1668,16 @@ impl Bytecode {
...
             "Program counter out of bounds"
         );
+
+        // No successor if pc is at the last instruction of the block. Return early to prevent overflow.
+        if pc > u16::max_value() - 2 {
```
sblackshear, line 1677: Would you mind adding a unit test that hits this case? LGTM otherwise!

language/move-binary-format/src/file_format.rs
```diff
@@ -1668,11 +1668,16 @@ impl Bytecode {
...
         assert!(
-            // The program counter could be added to at most twice and must remain
-            // within the bounds of the code.
-            pc <= u16::max_value() - 2 && (pc as usize) < code.len(),
+            // The program counter must remain within the bounds of the code
+            pc < u16::MAX && (pc as usize) < code.len(),
             "Program counter out of bounds"
         );
+
+        // No successor if pc is at the last instruction of the block. Return early to prevent overflow.
```
sblackshear, line 1676: Nit: comment is a bit confusing because this isn't actually checking whether the pc is at the last instruction in the block in the general case (only in the case where we have u16_max bytecode instructions).

sblackshear (approved):

language/move-binary-format/src/file_format.rs
runtian-zhou: Added the test

PR #536, opened 2022-10-04
[gas] share even more info with the gas meter

vgao1996:
This adds some additional callbacks to the gas meter trait and fine tune some existing ones so to allow us to keep track of the VM's memory usage precisely.


wrwg (approved):
These are an unexpected amount of changes, based on the convo this morning.

davidiw (approved):

PR #432, opened 2022-09-01
[move-compiler] Support underscores in number literals

macalinao:
## Motivation

Easier to read groupings of decimals.

### Have you read the [Contributing Guidelines on pull requests](https://github.com/move-language/move/blob/main/CONTRIBUTING.md#developer-workflow)?

Yes

## Test Plan

Add tests for parsing numbers

PR #542, opened 2022-10-05
tests for: number lit underscore: https://github.com/move-language/move/pull/432

oxade:
## Motivation

This PR works to close out: https://github.com/move-language/move/pull/432

Yes

## Test Plan
Parser & compiler tests


damirka (approved):
looks good to me

language/move-core/types/src/parser.rs
```diff
@@ -375,8 +375,50 @@ pub fn parse_struct_tag(s: &str) -> Result<StructTag> {
     }
 }
 
+/// Test cases for the parser that should always fail.
+/// Note: we don't have `cfg(test)` here because code gated by `cfg(test)` is
+/// not gated outside of its declaring crate: <https://github.com/rust-lang/cargo/issues/8379>
+pub const PARSE_VALUE_NEGATIVE_TEST_CASES: &[&str] = &[
```
sblackshear, line 381: I think it would be better to copy this everywhere it is used rather than exposing this (rather confusing) public constant. It will also make the relevant tests much easier to read/extend.

In general, I think code reuse in production code is great, but we should prioritize readability and avoiding coupling over code reuse in tests. As a rule of thumb, someone touching test code in component A should be able to easily read and understand the tests without looking at test code in component B, and should be able to freely change test inputs/outputs in component A without worrying about changing tests in component B.

sblackshear (changes requested):

language/move-core/types/src/parser.rs
oxade: This was a carry-over from the original PR.
But I agree with you and will change it.

language/move-command-line-common/src/parser.rs
```diff
@@ -467,7 +490,9 @@ mod tests {
             "3false",
             "3 false",
             "",
-        ] {
+        ];
+
+        for s in [PARSE_VALUE_NEGATIVE_TEST_CASES, &["0XFF", "0X0"]].concat() {
```
sblackshear, line 495: inline `0xFF` and `0X0` into the above?

sblackshear (approved):

language/move-compiler/tests/move_check/parser/num_literal_underscore_trailing.move
```diff
@@ -0,0 +1,6 @@
+module 0x42::M {
+    fun t() {
+        // Trailing underscore not allowed
+        let _ = 0u8_;
```
tnowacki, line 4: It is though right? Trailing underscore just isn't allowed after a type suffix. Like `0_` should be valid?

language/move-compiler/tests/move_check/parser/num_literal_underscore_leading.move
```diff
@@ -0,0 +1,6 @@
+module 0x42::M {
+    fun t() {
+        // Leading underscore not allowed
+        let _ = _0u8;
```
tnowacki, line 4: Would be nice to see this case replicated without a type suffix

language/move-compiler/tests/move_check/parser/num_literal_underscore.move
```diff
@@ -0,0 +1,18 @@
+module 0x42::M {
+    fun t() {
+        // Single umderscore separations allowed
+        let _ = 8_5u128;
```
tnowacki, line 4: need some cases without type suffix

language/move-prover/interpreter-testsuite/tests/concrete_check/load_constant_number_literal.move
```diff
@@ -0,0 +1,22 @@
+module 0x2::A {
+    #[test]
+    public fun load_constant_number_literal_u8() {
```
tnowacki, line 3: What is this test testing?

language/move-compiler/transactional-tests/tests/parser/num_literal_underscore.move
```diff
@@ -0,0 +1,13 @@
+//# run
+script {
```
tnowacki, line 2: need some cases without type suffix

language/move-compiler/tests/move_check/parser/num_literal_underscore_trailing.move
oxade: Yes thats valid and tested in the positive tests

PR #544, opened 2022-10-05
[compiler] Resolve 0x1::vector calls to opcodes

wrwg:
The phase in the compiler which maps calls to vector functions to opcodes only catches cases like `std::vector::fun`. If a user has `0x1::vector::fun` (which he should be free to do) no resolution to opcodes happens, and the legacy native functions in the library are called. Those functions, however, should be deactivated as they represent a further surface of concern of security and gas metering.

This PR lets the vector functions resolve to opcodes if either (a) the symbolic address name is `std` (b) or the address resolves to `0x1`.

### DIFF
--- a/language/move-compiler/src/naming/fake_natives.rs
+++ b/language/move-compiler/src/naming/fake_natives.rs
@@ -5,6 +5,7 @@
 //! as 'native`, but do not appear in the compiled module. For developer sanity, they must be marked
 //! with the `FAKE_NATIVE_ATTR`
 
+use move_core_types::account_address::AccountAddress;
 use std::convert::TryInto;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


language/move-compiler/src/naming/fake_natives.rs
```diff
@@ -86,37 +87,24 @@ pub fn resolve_builtin(
     function: &FunctionName,
 ) -> Option<fn(Vec<IR::Type>) -> IR::Bytecode_> {
     let sp!(_, ModuleIdent_ { address, module }) = module;
-    let addr_name = match address {
-        Address::Numerical(Some(sp!(_, n_)), _) | Address::NamedUnassigned(sp!(_, n_)) => n_,
+    // Only resolve if either (a) the address is named "std" or (b) its value is 0x1
+    match address {
+        Address::Numerical(Some(sp!(_, n)), sp!(_, a))
+            if n.as_str() == "std" || a.into_inner() == AccountAddress::ONE => {}
+        Address::NamedUnassigned(sp!(_, n)) if n.as_str() == "std" => {}
```
davidiw, line 94: well hack upon hack... 

I think the long term best solution is to figure out if we can funnel in the address prefix... but ... yeah... the situation currently is not great.

davidiw (approved):

vgao1996 (approved):

language/move-compiler/src/naming/fake_natives.rs
wrwg: That's actually the cleanest we can get (relative to the language definition). Here at this point we have both the symbolic and the numerical form of the address. We want to support that someone deploys `std` at a different address. And we want to support that users just write `0x1`. Thats what this enables.

The problem that we have on the one hand numeric addresses and on the other symbolic ones which can be aliases for the former is inherent to the address alias system. Only a change in the language can fix this (e.g. disallowing numeric addresses all together.)

PR #546, opened 2022-10-05
[compiler] Cherrypick: Resolve 0x1::vector functions to opcodes

wrwg:
Picks #544

The phase in the compiler which maps calls to vector functions to opcodes only catches cases like `std::vector::fun`. If a user has `0x1::vector::fun` (which he should be free to do) no resolution to opcodes happens, and the legacy native functions in the library are called. Those functions, however, should be deactivated as they represent a further surface of concern of security and gas metering.

This PR lets the vector functions resolve to opcodes if either (a) the symbolic address name is `std` (b) or the address resolves to `0x1`.

### DIFF
--- a/language/move-compiler/src/naming/fake_natives.rs
+++ b/language/move-compiler/src/naming/fake_natives.rs
@@ -5,6 +5,7 @@
 //! as 'native`, but do not appear in the compiled module. For developer sanity, they must be marked
 //! with the `FAKE_NATIVE_ATTR`
 
+use move_core_types::account_address::AccountAddress;
 use std::convert::TryInto;
 
 use crate::{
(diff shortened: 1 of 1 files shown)


language/move-compiler/src/naming/fake_natives.rs
```diff
@@ -86,37 +87,24 @@ pub fn resolve_builtin(
     function: &FunctionName,
 ) -> Option<fn(Vec<IR::Type>) -> IR::Bytecode_> {
     let sp!(_, ModuleIdent_ { address, module }) = module;
-    let addr_name = match address {
-        Address::Numerical(Some(sp!(_, n_)), _) | Address::NamedUnassigned(sp!(_, n_)) => n_,
+    // Only resolve if either (a) the address is named "std" or (b) its value is 0x1
+    match address {
+        Address::Numerical(Some(sp!(_, n)), sp!(_, a))
+            if n.as_str() == "std" || a.into_inner() == AccountAddress::ONE => {}
```
tnowacki, line 93: Ah, interesting case. 

And also interesting that this should have been updated after named addresses were made non-opaque for the packages system. 

Maybe we should also check that `"std"` is bound to `0x1` in the address map? Not really a big deal but just concerned about this being accidental.

wrwg: I thought one idea of aliases is that `std` can be mapped to somewhere else if one desires. In any case, if did not enforce this before, enforcing it now would be a breaking change. So I would just leave as is.

wrwg:
> EDIT: put this comment on the wrong PR

Your original comment though absolutely fitting for U256 ;-) We can't lock down main by incremental implementation of this, either do in one PR, or use a branch.

Still this one here needs some attention.

language/move-compiler/src/naming/fake_natives.rs
```diff
@@ -86,37 +87,26 @@ pub fn resolve_builtin(
...
-        Address::Numerical(Some(sp!(_, n_)), _) | Address::NamedUnassigned(sp!(_, n_)) => n_,
-        _ => return None,
+    // Only resolve if either (a) the address is named "std" or (b) its value is 0x1
+    match address {
+        Address::Numerical(Some(sp!(_, n)), _) | Address::NamedUnassigned(sp!(_, n))
+            if n.as_str() == "std" => {}
+        Address::Numerical(_, sp!(_, a)) if a.into_inner() == AccountAddress::ONE => {}
+        _ => {
+            return None;
+        }
```
tnowacki, line 97: After doing some more digging, some thoughts here:

In my pervious comment, I was wondering if we could check if `std` == `0x1`. Unfortunately that isn't really easy after the address-map became a per-package binding. So we could thread that through to here but that is way more work than we we need right now. 

In an early version of this design, I considered doing something like `#[bytecode_instruction(VecBorrow)]` instead of just `#[bytecode_instruction]` and then not even checking things like the function name, but maybe doing some sanity checking for signature. That seemed like overkill at the time. But it would help here for this case as we wouldn't need these extra checks.
What I'm getting at is the motivation of all of these checks is to prevent unexpected behavior from someone programming against a module. We have the `native` check for that. We have the `#[bytecode_instruction]` for that. 

I don't really care at this point if we also have the address check for that. 

So we can either:
1) Land this code as is, and revisit it if anyone ever wants to put `std` at something other than 0x1
2) Just delete this check

wrwg: I guess let's go with (1) is the safest. I'm not sure about the consequences removing this check, with all this fuzz lately, keeping the check seems better. I believe a user _can_ actually locate `std` elsewhere than `0x1` with this code.

tnowacki (approved):
Fine for now! @wrwg could you issue a follow up just to harden/generalize this check?

Issue #453, opened 2022-09-08
[Accepted feature] native `u16`, `u32`, and `u256` types


sblackshear:
## Rationale
- The type u256 is important for interoperability with EVM chains. For example, NFTs are usually represented in 256 bits. The type is also important for Move on EVM. To have an idiomatic experience of using Move to create EVM contracts, u256 is essential. Moreover, u256 has the property to be large enough to be able to represent a safe hash for any representable value, which makes it a useful primitive for design patterns based on cryptographic concepts.  
- For the other types: Move has `u8` and `u64`, but is missing `u16` and `u32`. It makes sense to fill in these gaps, especially given that BCS uses `u32`'s in some areas (e.g., vector indexes) where we also probably want `u32`'s in Move

## Design
Natively support the new integer types. All operations and notations available for the existing integer types (equality, comparison, casts, etc.) should be supported.

For `u256` specifically, we will need to figure out how to represent the type in serialized form since it is not supported by BCS, Rust, or `serde-reflection`. Extending BCS is one option, but there may be easier choices.

## Challenges
1. We need to add support for u256 to BCS. Since the original bcs repo is under the original Diem organization, we may need to fork it.
2. This will likely require a new file format version. We need to check whether we have other breaking changes coming and decide how to coordinate those.
3. The addition of the new integer types will break many existing core and vm APIs. We need to assess the impacts of this on clients and come up with a recommended upgrade protocol for them. 

## Alternatives considered
- Move frameworks that want u256 for EVM compatibility or other reasons can implement it in pure Move (with or without native function support), as Starcoin has done [here](https://github.com/starcoinorg/starcoin-framework/blob/main/sources/U256.move). We could also consider adding this to the Move stdlib and transparently compiling +, - etc on u256 to use this module. This would keep the rather EVM-specific u256 out of the platform-agnostic bytecode language.

- Support `u256`, but do not add `u16` and `u32` at this time

wrwg:
Perhaps we should add some section about "Challenges" to the template. Here it appears to be that a local change in the Move repo may work out fine, but then break many downstream projects. We need to find some model to deal with this...

vgao1996:
@wrwg @sblackshear  I added a challenge section for this proposal. Feel free to add more if you spot anything something

sblackshear:
> We need to add support for u256 to BCS. Since the original bcs repo is under the original Diem organization, we may need to fork it.

[...]

PR #547, opened 2022-10-05
u16, u32, u256 support

oxade:
## Motivation

Need more integer types. See feature [proposal](https://github.com/move-language/move/issues/453)
This PR adds u16, u32, u256.
Example: `let _: u256 = ((234u16 as u32) as u256)`

Yes

## Test Plan

Every area which uses integers
Too many to list, but please double-check that all your favorite tests are covered.


language/move-core/types/src/u256.rs
```diff
@@ -0,0 +1,17 @@
...
+use primitive_types::U256;
+use serde::{Deserialize, Serialize};
+
+#[allow(non_camel_case_types)]
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq, Eq, Hash)]
+pub struct u256(U256);
+
+impl fmt::Display for u256 {
+    fn fmt(&self, f: &mut fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{}", self.0)
```
tnowacki, line 15: Probably want to delegate for any special cases (around hex and what not)
So `fmt::Display(self.0, f)` or I think `self.0.fmt(f)`

oxade: Agreed. I haven't spent enough time fleshing this out yet. I reckon there'll be lost of things to impl in subsequent PRs.
For now I'm focusing on the bare minimum

tnowacki: I just mean that this specific case for Display can be simpler, while being more robust

language/move-core/types/src/u256.rs
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use std::fmt;
+
+use primitive_types::U256;
```
tnowacki, line 6: Have we finalized this choice?

language/move-core/types/src/u256.rs
tnowacki: But understood that more impls will come later

language/move-core/types/src/u256.rs
oxade: No. The plan was to use something for now, wrap it in a new struct, and impl the necessary traits.
The underlying struct can then be swapped out as needed.

language/move-core/types/src/u256.rs
oxade: > I just mean that this specific case for Display can be simpler, while being more robust
Got it. Will address

tnowacki:
Just thinking out loud here, I'm wondering if we should make a branch. Adding things to MoveValue and other spots without full support in the VM is going to be painful for anyone keeping up to date with the repo on main

oxade:
> Just thinking out loud here, I'm wondering if we should make a branch. Adding things to MoveValue and other spots without full support in the VM is going to be painful for anyone keeping up to date with the repo on main

Thats a good idea, however I've not been able to create a branch in this repo yet. Do you have permissions to do it?

language/move-core/types/src/u256.rs
```diff
@@ -1,18 +1,139 @@
...
+};
 
 // This U256 impl was chosen for now but we are open to changing it as needed
 use primitive_types::U256;
 use serde::{Deserialize, Serialize};
 
 #[allow(non_camel_case_types)]
-#[derive(Clone, Deserialize, Serialize, Debug, PartialEq, Eq, Hash, Copy)]
+#[derive(Clone, Deserialize, Serialize, Debug, PartialEq, Eq, Hash, Copy, PartialOrd, Ord)]
 pub struct u256(U256);
```
vgao1996, line 19: I kinda feel like we should rename the struct to U256 with a capital U. By rust convention, type names starting with a lower letter is reserved for primitive types. It might become problematic if let's say, rust adds its own u256 type one day.

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -549,6 +579,20 @@ impl IndexedRef {
                 r1.borrow()[self.idx] == *r2.borrow()[other.idx].as_value_ref::<u8>()?
             }
 
+            (Locals(r1), VecU16(r2)) | (Struct(r1), VecU16(r2)) => {
```
vgao1996, line 582: cc @runtian-zhou this is gonna conflict hard with your runtime checks

vgao1996 (commented):
Some early feedback

language/move-vm/types/src/values/values_impl.rs
runtian-zhou: It's fine. It's already lots of rebase!

language/move-core/types/src/u256.rs
oxade: Yes I initially used the uppercase U256 but there were conflicts due to `U256` also having to be the Move type name in various parts of the code

language/move-core/types/src/u256.rs
```diff
@@ -0,0 +1,270 @@
...
+const U256_NUM_BYTES: usize = 32;
+pub type U256FromStrError = FromStrRadixErr;
+
+// This U256 impl was chosen for now but we are open to changing it as needed
+use primitive_types::U256 as PrimitiveU256;
+use serde::{ser::SerializeTuple, Deserialize, Deserializer, Serialize, Serializer};
+
+#[allow(non_camel_case_types)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Copy, PartialOrd, Ord)]
+pub struct U256Inner(PrimitiveU256);
```
oxade, line 23: The name `U256Inner` was chosen on a whim. I'm open to other names. Maybe something like `U256Move`? Since we'll have to expose this type to users.

bmwill: We could have this be `U256` and just have the Move VM type be `U256(move_core_types::U256)`.

language/move-binary-format/src/file_format_common.rs
```diff
@@ -121,6 +121,10 @@ pub enum SerializedType {
     VECTOR                  = 0xA,
     STRUCT_INST             = 0xB,
     SIGNER                  = 0xC,
+
```
tnowacki, line 124: Nit remove empty line

language/move-binary-format/src/file_format_common.rs
```diff
@@ -209,6 +213,13 @@ pub enum Opcodes {
     VEC_POP_BACK                = 0x45,
     VEC_UNPACK                  = 0x46,
     VEC_SWAP                    = 0x47,
+
```
tnowacki, line 216: Nit remove empty line

language/move-binary-format/src/serializer.rs
```diff
@@ -720,14 +723,32 @@ fn serialize_ability_sets(binary: &mut BinaryData, sets: &[AbilitySet]) -> Resul
...
 
 /// Serializes a single `Bytecode` instruction.
-fn serialize_instruction_inner(binary: &mut BinaryData, opcode: &Bytecode) -> Result<()> {
+fn serialize_instruction_inner(
+    major_version: u32,
+    binary: &mut BinaryData,
+    opcode: &Bytecode,
+) -> Result<()> {
     let res = match opcode {
+        Bytecode::CastU16
```
tnowacki, line 738: two nits: first, flip the order of `Cast*` and `Ld*` to match the bytecode enum. second, move these close to the match below

language/move-core/types/src/vm_status.rs
```diff
@@ -636,8 +636,11 @@ pub enum StatusCode {
     DUPLICATE_TABLE = 3010,
     UNKNOWN_ABILITY = 3013,
     UNKNOWN_NATIVE_STRUCT_FLAG = 3014,
+    BAD_U16 = 3017,
+    BAD_U32 = 3018,
```
tnowacki, line 640: woah... that was lucky lol

language/move-binary-format/src/deserializer.rs
```diff
@@ -1384,6 +1422,22 @@ fn load_code(cursor: &mut VersionedCursor, code: &mut Vec<Bytecode>) -> BinaryLo
         };
         // conversion
         let bytecode = match opcode {
+            Opcodes::LD_U16
+            | Opcodes::LD_U32
+            | Opcodes::LD_U256
+            | Opcodes::CAST_U16
+            | Opcodes::CAST_U32
+            | Opcodes::CAST_U256
```
tnowacki, line 1430: Nit move this into the match above with the vec opcocdes

language/move-binary-format/src/serializer.rs
tnowacki: Maybe move it out of the match, like we do for vector instructions in the deserializer.

language/move-command-line-common/src/parser.rs
```diff
@@ -350,6 +380,17 @@ pub fn parse_u128(s: &str) -> Result<(u128, NumberFormat), ParseIntError> {
     ))
 }
 
+// Parse a u256 from a decimal or hex encoding
+pub fn parse_u256(
+    s: &str,
+) -> Result<(move_core_types::u256::U256, NumberFormat), U256FromStrError> {
```
tnowacki, line 386: Any reason not to alias this type here? So you can just do `U256`?

language/move-command-line-common/src/values.rs
```diff
@@ -41,10 +41,13 @@ pub enum ValueToken {
 #[derive(Eq, PartialEq, Debug, Clone)]
 pub enum ParsedValue<Extra: ParsableValue = ()> {
     Address(ParsedAddress),
-    InferredNum(u128),
+    InferredNum(move_core_types::u256::U256),
```
tnowacki, line 44: Same here, maybe just import U256?

But you can do this in a future PR :)

language/move-core/types/src/vm_status.rs
oxade: very much so!

language/move-command-line-common/src/values.rs
oxade: Yup!

language/move-core/types/src/language_storage.rs
```diff
@@ -28,10 +28,16 @@ pub enum TypeTag {
     Bool,
     #[serde(rename = "u8", alias = "U8")]
     U8,
+    #[serde(rename = "u16", alias = "U16")]
+    U16,
+    #[serde(rename = "u32", alias = "u32")]
```
bmwill, line 33:
```suggestion
    #[serde(rename = "u32", alias = "U32")]
```

language/move-core/types/Cargo.toml
```diff
@@ -19,6 +19,10 @@ rand = "0.8.3"
 ref-cast = "1.0.6"
 serde = { version = "1.0.124", default-features = false }
 serde_bytes = "0.11.5"
+primitive-types = {version = "0.10.1", features = ["impl-serde"]}
+uint = "0.9.4"
+num = "0.4.0"
+ethnum = "1.0.4"
```
bmwill, line 25: Do we need all of these?

language/move-core/types/src/u256.rs
```diff
@@ -0,0 +1,376 @@
...
+    fmt,
+    mem::size_of,
+    ops::{BitAnd, BitOr, BitXor, Shl, Shr},
+};
+use uint::FromStrRadixErr;
+
+const NUM_BITS_PER_BYTE: usize = 8;
+const U256_NUM_BITS: usize = 256;
+const U256_NUM_BYTES: usize = U256_NUM_BITS / NUM_BITS_PER_BYTE;
+pub type U256FromStrError = FromStrRadixErr;
```
bmwill, line 21: You'll probably want to wrap this in a NewType so as to not leak the dep on uint

language/move-core/types/src/u256.rs
```diff
@@ -0,0 +1,376 @@
...
+pub struct U256(PrimitiveU256);
+
+impl fmt::Display for U256 {
+    fn fmt(&self, f: &mut fmt::Formatter) -> std::fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+impl std::str::FromStr for U256 {
+    type Err = anyhow::Error;
```
bmwill, line 38: You have defined a FromStrError type above, why not use it here?

language/move-core/types/src/u256.rs
```diff
@@ -0,0 +1,376 @@
...
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let b: [u8; U256_NUM_BYTES] = self.to_le_bytes();
+        let mut seq = serializer.serialize_tuple(U256_NUM_BYTES)?;
+        for x in b {
+            seq.serialize_element(&x)?;
+        }
+        seq.end()
```
bmwill, line 66: [u8; 32] does implement Serialize so you may be able to do
```suggestion
        self.to_le_bytes().serialize(serializer)
```

language/move-core/types/src/u256.rs
```diff
@@ -0,0 +1,376 @@
...
+
+impl From<&U256> for EthnumU256 {
+    fn from(n: &U256) -> EthnumU256 {
+        // TODO (ade): use better solution for conversion
+        // Currently using str because EthnumU256 can be little or big endian
+        let num_str = format!("{:X}", n.0);
+        // TODO (ade): remove expect()
+        EthnumU256::from_str_radix(&num_str, 16).expect("Cannot convert to U256")
+    }
+}
```
bmwill, line 283: Whats the reason for implementing these conversions? This inadvertently places these types in the public api and we may not want to do that

language/move-core/types/src/u256.rs
```diff
@@ -0,0 +1,376 @@
...
+
+impl TryFrom<U256> for u8 {
+    type Error = anyhow::Error;
```
bmwill, line 286: I would probably recommend using a concrete error type for this vs using anyhow. (this goes for all of the conversions). You could probably define a single type and use that in all the TryFrom conversion impls.

bmwill (commented):
Awesome work, here are some nits

language/move-compiler/src/typing/expand.rs
```diff
@@ -157,25 +157,38 @@ pub fn exp(context: &mut Context, e: &mut T::Exp) {
...
-                } else {
-                    assert!(v > u8_max);
+                } else if v > u32_max {
                     BT::U64
+                } else if v > u16_max {
+                    BT::U32
+                } else if v > u8_max {
+                    BT::U16
+                } else {
+                    BT::U8
```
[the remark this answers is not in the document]
tnowacki: We should add a test for this
```
let i = 2;
let j = 0;
while (j < 64) {
  i = 2 * i;
  j = j + 1;
};
```

language/tools/move-unit-test/src/test_runner.rs
```diff
@@ -396,6 +400,8 @@ impl SharedTestingConfig {
             .unwrap_or_else(|e| panic!("Unable to build stackless bytecode: {}", e));
 
             if model.has_errors() {
+                let mut stderr = StandardStream::stderr(ColorChoice::Always);
+                model.report_diag(&mut stderr, Severity::Error);
```
tnowacki, line 404: Why the change here?

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -539,6 +578,20 @@ impl IndexedRef {
...
```
tnowacki, line 581: Huh... do these type mismatched cases ever happen? I don't think they should... but thats for another PR

language/move-core/types/src/u256.rs
oxade: `BigInt`: Conversions are done between BigInt and primitives in Move Model
`EthnumU256`: Prover uses EthnumU256 and needs conversions.

I agree the downside to the current approach is leaking types. I can move the conversions closer to the usage areas instead.

language/tools/move-unit-test/src/test_runner.rs
oxade: Debugging errors was hard. This just prints more concrete info.

language/move-vm/types/src/values/values_impl.rs
oxade: I actually dont know. I added them for completeness. 
Whats the ideal approach here?

language/move-compiler/src/typing/expand.rs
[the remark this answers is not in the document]
oxade: Ah yes will change `U8` to `U64`
Good catch

language/move-core/types/src/u256.rs
bmwill: Why can't the prover use this type directly?

bmwill: Where "this type" is the `U256` wrapper that you define in this file.

tnowacki (changes requested):

language/move-prover/tests/sources/regression/vector_200630.move
```diff
@@ -13,9 +13,10 @@ module 0x42::TestVector {
         let elem = vector::borrow_mut(v, i);
         let int_ref = &mut elem.x;
 
-        *int_ref = 42;
+        // Forced u64 type since U256 not fully supported in Prover
+        *int_ref = 42u64;
```
tnowacki, line 17: Why would this not be a u64?

language/move-model/tests/sources/schemas_err.exp
```diff
@@ -52,7 +52,7 @@ error: cyclic schema dependency: Cycle1 -> Cycle2 -> Cycle3 -> Cycle1
 80 │         include Cycle1;
    │                 ^^^^^^
 
-error: expected `bool` but found `u128` in expression
+error: expected `bool` but found `u256` in expression
```
tnowacki, line 55: What caused the change here?

language/move-vm/transactional-tests/tests/entry_points/all_possible_param_types.exp
```diff
@@ -1 +1,10 @@
 processed 1 task
+
+task 0 'run'. lines 1-5:
+Error: Script execution failed with VMError: {
+    major_status: FAILED_TO_DESERIALIZE_ARGUMENT,
```
tnowacki, line 5: This looks wrong I think

language/move-compiler/transactional-tests/tests/operators/arithmetic_operators_u16.exp
```diff
@@ -0,0 +1,109 @@
...
+    major_status: INTERNAL_TYPE_ERROR,
+    sub_status: None,
+    location: script,
+    indices: [],
+    offsets: [(FunctionDefinitionIndex(0), 2)],
+}
+
+task 16 'run'. lines 168-174:
+Error: Script execution failed with VMError: {
+    major_status: INTERNAL_TYPE_ERROR,
```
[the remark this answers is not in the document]
oxade: Yes it seems wrong error type is being thrown. Investigating. 
The fact that there are errors is expected though.

language/move-core/types/src/u256.rs
bmwill: Did the prover already have support for u256? or did this pr add it?

If you want to do that in a followup that's fine, just make sure you include a comment indicating that these deps and From impls will be removed shortly

language/move-model/tests/sources/schemas_err.exp
oxade: Largest container changed

language/move-vm/transactional-tests/tests/entry_points/all_possible_param_types.exp
oxade: It is wrong. Fixing it

oxade: ah I see. This test in particular want strong typing suffix.

language/move-core/types/src/u256.rs
oxade: Good call out. Working on this

language/move-core/types/src/u256.rs
oxade: It used some u256, which I didn't realize till I had implemented something else for other components 
The sensible thing is to harmonize 
I'll mark this as a TODO and have a follow up PR

oxade:
From private chats with @tnowacki, I'll add more tests as follows:
Bytecode verifier:
* stack balancing
* general typing

VM:
* some serialization tests maybe with resources + view
* direct math tests + some edge cases in the VM (not through the source language because of constant folding)

language/move-core/types/src/u256.rs
oxade: Created this to track https://github.com/move-language/move/issues/597

language/move-prover/boogie-backend/src/bytecode_translator.rs
```diff
@@ -1260,6 +1262,26 @@ impl<'env> FunctionTranslator<'env> {
...
+                    CastU32 => {
+                        let src = srcs[0];
+                        let dest = dests[0];
+                        emitln!(
+                            writer,
+                            "call {} := $CastU32({});",
+                            str_local(dest),
+                            str_local(src)
+                        );
+                    }
```
rahxephon89, line 1284: Could you also add the implementation of CastU256? Thanks!

oxade: Done!

language/move-compiler/src/typing/expand.rs
```diff
@@ -157,24 +157,37 @@ pub fn exp(context: &mut Context, e: &mut T::Exp) {
...
             };
             let v = *v;
-            let u8_max = std::u8::MAX as u128;
-            let u64_max = std::u64::MAX as u128;
-            let u128_max = std::u128::MAX;
+            let u8_max = U256::from(std::u8::MAX);
+            let u16_max = U256::from(std::u16::MAX);
+            let u32_max = U256::from(std::u32::MAX);
+            let u64_max = U256::from(std::u64::MAX);
+            let u128_max = U256::from(std::u128::MAX);
```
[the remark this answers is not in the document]
oxade: this would be ideal but the current U256 lib doesnt have const impl for creating nums.

language/move-compiler/src/typing/expand.rs
```diff
@@ -157,24 +157,37 @@ pub fn exp(context: &mut Context, e: &mut T::Exp) {
...
+                let fix_bt = if v > u128_max {
+                    BT::U256
+                } else if v > u64_max {
                     BT::U128
+                } else if v > u32_max {
+                    BT::U64
+                } else if v > u16_max {
+                    BT::U32
+                } else if v > u8_max {
+                    BT::U16
```
tnowacki, line 189: I think this last case is unreachable

language/move-binary-format/src/serializer.rs
```diff
@@ -720,13 +723,34 @@ fn serialize_ability_sets(binary: &mut BinaryData, sets: &[AbilitySet]) -> Resul
...
+        | Bytecode::CastU16
+        | Bytecode::CastU32
+        | Bytecode::CastU256
+            if (major_version < VERSION_6) =>
+        {
+            return Err(anyhow!(
+                "Loading or casting u16, u32, u256 integers not supported in bytecode version {}",
+                major_version
+            ));
+        }
```
vgao1996, line 750: Hmm... I'm wondering if the serializer needs to be versioned or not.

@tnowacki could you remind me what the contract here is? Do only support serializing to the latest version?

vgao1996 (commented):
Review 1/2. Sorry for the delay

language/evm/move-to-yul/src/context.rs
```diff
@@ -769,7 +772,7 @@ impl<'a> Context<'a> {
                 U128 => 16,
                 // TODO: optimize for 20 bytes? Then we need primitives like LoadU160 etc.
                 Address | Signer => 32,
-                Num | Range | EventStore => {
+                Num | Range | EventStore | U16 | U32 | U256 => {
```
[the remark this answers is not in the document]
oxade: I didnt focus on getting the EVM stuff. We can do that in subsequent PR.

language/move-binary-format/src/serializer.rs
tnowacki: @wrwg added this relatively recently. We currently can target serialization formats. But only V5, and now V6, are supported.

We can/will use this in the compiler for a rough and dirty ability to target old file format versions with the most recent compiler.

language/move-binary-format/src/deserializer.rs
```diff
@@ -1382,6 +1423,26 @@ fn load_code(cursor: &mut VersionedCursor, code: &mut Vec<Bytecode>) -> BinaryLo
...
         };
+
+        match opcode {
+            Opcodes::LD_U16
+            | Opcodes::LD_U32
+            | Opcodes::LD_U256
+            | Opcodes::CAST_U16
+            | Opcodes::CAST_U32
+            | Opcodes::CAST_U256
+                if (cursor.version() < VERSION_6) =>
```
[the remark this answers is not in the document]
tnowacki: Correct.

language/move-vm/transactional-tests/tests/instructions/arithmetic_operators_u16.mvir
```diff
@@ -0,0 +1,164 @@
+//# run
+main() {
```
tnowacki, line 2: Can you split these tests into positive and negative tests? It is a bit confusing to have some operations that should fail and some that shouldn't in the same test, makes it hard to see what exactly is getting coverage.

language/move-vm/transactional-tests/tests/instructions/arithmetic_operators_u32.mvir
```diff
@@ -0,0 +1,164 @@
...
+main() {
+label b0:
+    _ = 429496u32 * 429496u32;
+    return;
+}
+
+//# run
+main() {
+label b0:
+    _ = 2147483647 * 2u32;
```
tnowacki, line 87: Is this supposed to be a type error? If so, we should move it to the bytecode verifier.

language/evm/move-to-yul/src/context.rs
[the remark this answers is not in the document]
vgao1996: Seems fine to me.

language/move-binary-format/src/serializer.rs
vgao1996: Sounds good then.

language/move-vm/transactional-tests/tests/instructions/arithmetic_operators_u16.mvir
oxade: @tnowacki I copied the same format which existed for u8, u64, u128, which also mixed the passes and fails.
I 100% agree it should be separated (was hard for me to read too) but this faulty pattern is repeated all over the codebase.
If I were fixing tests, this PR would be much much bigger so now I wanna focus on the core essence of adding the requested integer support.
We can file a cleanup PR as needed.

language/move-vm/transactional-tests/tests/instructions/arithmetic_operators_u32.mvir
oxade: @tnowacki fixed. Last digit should've been 8, and suffixed. Was supposed to be arithmetics tests

oxade:
@tnowacki @vgao1996 please let me know if any other feedback. Hoping to get this merged soon.

oxade:
@tnowacki @vgao1996 all comments and requested tests have been addressed.

language/move-bytecode-verifier/transactional-tests/tests/type_safety/integers_valid.mvir
```diff
@@ -0,0 +1,119 @@
...
+    x = 100u8;
+    x = 1u8 + 1u8;
+    x = 1u8 - 1u8;
+    x = 1u8 * 1u8;
+    x = 1u8 / 1u8;
+    x = 1u8 % 1u8;
+    x = 1u8 & 1u8;
+    x = 1u8 | 1u8;
+    x = 1u8 << 1u8;
+    x = 1u8 >> 1u8;
```
tnowacki, line 16: I think we forgot xor... my bad

tnowacki (approved):
Thanks for the heroic work! And all the extra tests! We can fix some more test exhaustiveness in followup PRs

language/move-prover/boogie-backend/src/bytecode_translator.rs
vgao1996, line 1284: Looks correct to me. cc @wrwg @rahxephon89 @junkil-park  if you wish to double check

language/move-core/types/src/u256.rs
```diff
@@ -0,0 +1,431 @@
...
+
+    /// One value as U256
+    pub const fn one() -> Self {
+        Self(PrimitiveU256::one())
+    }
+
+    /// Max value of U256: 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF
+    pub const fn max() -> Self {
+        Self(PrimitiveU256::max_value())
+    }
```
vgao1996, line 189: Please rename this to `max_value` so to be consistent with the primitive integer types. Similarly we may also want to add an associated constant MAX.

By convention `max` is used for getting the bigger number from two. See https://doc.rust-lang.org/std/primitive.u64.html#impl-Ord-for-u64 for example

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -1349,6 +1352,39 @@ impl Frame {
...
+                            .operand_stack
+                            .push(Value::u32(integer_value.cast_u32()?))?;
+                    }
+                    Bytecode::CastU256 => {
+                        gas_meter.charge_simple_instr(S::CastU16)?;
+                        let integer_value = interpreter.operand_stack.pop_as::<IntegerValue>()?;
+                        interpreter
+                            .operand_stack
+                            .push(Value::u256(integer_value.cast_u256()?))?;
+                    }
```
vgao1996, line 1387: Nit: can you please reorder these so that they appear next to the existing load and cast instructions?

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -1219,6 +1361,16 @@ impl VMValueCast<u128> for IntegerValue {
     }
 }
 
+impl VMValueCast<move_core_types::u256::U256> for IntegerValue {
```
vgao1996, line 1364: Yeah should probably just import `move_core_types::u256::U256` at the top to avoid repeating it many times, but this is a minor problem we can fix later.

vgao1996 (approved):
Overall the PR looks good to me and thank you so much for the quality work! However please allow me a just a little bit more time to sync-up with our team and see if they have any concerns (especially about backward compatibility) before you land this PR. Will get back to you real soon

language/move-prover/boogie-backend/src/bytecode_translator.rs
oxade: We version-gate this. Doesn't that work?

language/move-vm/types/src/values/values_impl.rs
oxade: Yeah there's already `U256` used for `IntegerValue::U256`, so I wanted to disambiguate.
But I will use `u256::U256` instead.

oxade:
> Overall the PR looks good to me and thank you so much for the quality work! However please allow me a just a little bit more time to sync-up with our team and see if they have any concerns (especially about backward compatibility) before you land this PR. Will get back to you real soon

We version-gate this by adding `VERSION_6`. Doesn't that work?
Please elaborate on the desired backwards compat needs/approaches

language/move-prover/boogie-backend/src/bytecode_translator.rs
oxade: I'm sure @rahxephon89 already checked this. See
https://github.com/move-language/move/pull/547#discussion_r998915512

vgao1996 (changes requested):
I think I now understand the problem with backward compatibility. 

Assuming you have an existing cluster of validators running the old version of the Move VM that only recognizes binary format v5 and you now want to upgrade them to v6. 

Right now `VERSION_MAX` is a hard-coded number. The old software only recognizes Move binaries up to v5 and rejects the ones that are v6 or higher, with the new one recognizing v6. This creates a problem during a software upgrade since validators can get out of sync.

I think what's missing here is a way to dynamically specify the maximum version the deserializer (in the VM) should recognize. Let's discuss this in details during tomorrow's meeting. (Don't panic. I think there is a fairly simple solution.) I'm sorry for introducing additional delay here, but backward compatibility is an important issue and I appreciate your patience.

In the meantime @wrwg would also love to double check the serializer and make sure it can now target older binary format versions (e.g. v5, v4) properly.

oxade:
> I think I now understand the problem with backward compatibility.
> 
> Assume you have an existing cluster of validators running the old version of the Move VM that only recognizes binary format v5 and you now want to upgrade them to v6.
> 
> Right now `VERSION_MAX` is a hard-coded number, which means the old software only recognizes Move binaries up to v5 and rejects the ones that are v6 or higher, with the new one recognizing v6. This creates a problem during a software upgrade since validators can get out of sync.
> 
> I think what's missing here is a way to dynamically specify the maximum version the deserializer (in the VM) should recognize. Let's discuss this in details during tomorrow's meeting. (Don't panic. I think there is a fairly simple solution.) I'm sorry for introducing additional delay here, but backward compatibility is an important issue and I appreciate your patience.
> 
> In the meantime @wrwg would also love to double check the serializer and make sure it can now target older binary format versions (e.g. v5, v4) properly.

This seems less like a problem with this PR, but more a problem with
* Move's version handling logic in general. This pattern of gating ser-des is already done in other parts of the codebase and this PR does not invent anything new in that direction.
* Your software upgrade logic which did not account for the limitation of Move's versioning logic.

So here it seems like the PR is being gated for reasons specific to your use case. 
So I believe we can decouple the versioning logic problems of Move from the core of this PR which is to add new integer support. We can tackle the former in the broader sync.

Or is there something else fundamentally aberrant about how we're doing things here?

tnowacki:
> Right now `VERSION_MAX` is a hard-coded number, which means the old software only recognizes Move binaries up to v5 and rejects the ones that are v6 or higher, with the new one recognizing v6. This creates a problem during a software upgrade since validators can get out of sync.

This would be a problem with the adapter using Move, not with Move itself. You should have some flag somewhere that marks the maximum allowed file format version number, and if one is submitted past that, you need to give the same error. 

For what it is worth, this logic did exist at some point in Diem. There was a resource that contained the maximum allowed file format version in the validator/node config.

vgao1996:
> So here it seems like the PR is being blocked for reasons specific to your use case.

> This would be a problem with the adapter using Move, not with Move itself
> So to be clear, this is a problem in the adapter, not in Move.

You are right that it's up to the adapter to implement a feature flag and decide how it wants to store and fetch it, but I strongly disagree that this is totally an adapter problem. The actual deserialization happens inside the Move VM so you can't really implement such versioning scheme without the VM providing a configurable option.

Additionally, this is just an instance of a class of more general problems with the VM. Strictly speaking, all new behaviors should be gated by some config options or otherwise we risk breaking existing downstream clients. Consider this thought experiment: if one day we decide to change the semantics of addition, do we still claim it's the adapters' responsibility to fix it? Is it even possible for them to fix it? IMO having a config option that would allow one to enable or disable the new behavior is the only feasible way to roll out changes like this in the general case. (We can and should delete some of these config options once we get past a support window.) 

> Move's version handling logic in general. This pattern of gating ser-des is already done in other parts of the codebase and this PR does not invent anything new in that direction.
> Your software upgrade logic which did not account for the limitation of Move's versioning logic.

Again, I should stress that this is the first time we introduce a new binary format version under the assumption that we need to take backward compatibility seriously so it's not surprising that we run into new issues like this. The fact that some versioning logic is missing doesn't mean it's correct. It's simply the result of us not having to do it previously (at Meta) due to Diem being a somewhat restricted system and  Starcoin not requiring us to support them at that time.

vgao1996:
Actually I think there is a relatively simple solution to this problem:
1. Add a new interface `pub fn deserialize_with_max_version(binary: &[u8], max_version: u64)` to `CompiledModule` while keeping `pub fn deserialize` by letting it delegate to `pub fn deserialize_with_max_version` internally
2. Add a new Move VM config `max_version: u64`
3. Let the Move VM use `deserialize_with_max_version` while other tools could continue to use `deserialize`

vgao1996:
cc @jolestar for this being a potential breaking change that needs to be applied with care.

tnowacki:
If you are able to pass a config to the VM to do the versioning, you are able to do the versioning on the adapter side too. In other words, the existence of being able to configure the VM implies that you could gate the VM. 

Though I agree it doesn't feel optimal! It would be a bit strange that the adapter can rely on the VM for everything compatibility wise, except for this case. 

That being said, is there any reason to block this PR for that?
Y'all already have your own branch on Aptos. You could just wait until this new deserialization entry point is added to pull this into your branch?
I know things are a bit choppy at the moment and y'all are probably pulling (and maybe releasing) a bit frequently. But as long as all the things are fixed by your next release, it should be fine?

tnowacki:
I think what I'm getting at is I would consider `main` more of a stable/nightly build? And that we should at various points tag commits as a specific release. And only at that point would an adapter want to pull in the new code. 

While we are a bit bumpy on the Move side in actually doing this, we are doing it sort of (at least conceptually). We make batches of changes into the next file format version, and then at some point say we are "releasing" that file format version, and no changes can be made (for that version).

vgao1996:
>  Hence I believe we can decouple the versioning logic problems of Move from the core of this PR which is to add new integer support. We can tackle the former in the broader sync.

> That being said, is there any reason to block this PR for that?
Y'all already have your own branch on Aptos. You could just wait until this new deserialization entry point is added to pull this into your branch?
I know things are a bit choppy at the moment and y'all are probably pulling (and maybe releasing) a bit frequently. But as long as all the things are fixed by your next release, it should be fine?

I think you guys have a very valid point. I'd be perfectly happy to leave out the new deserialization entry point for a future PR, but just want to make sure we're aligned on the general direction about backward compatibility, which I think we are now, aren't we?

vgao1996:
> I think what I'm getting at is I would consider main more of a stable/nightly build? And that we should at various points tag commits as a specific release. And only at that point would an adapter want to pull in the new code.
> While we are a bit bumpy on the Move side in actually doing this, we are doing it sort of (at least conceptually). We make batches of changes into the next file format version, and then at some point say we are "releasing" that file format version, and no changes can be made (for that version).

We share the same view. I don't think it's feasible to require every commit to be complete in terms of backward compatibility.

tnowacki:
> > Hence I believe we can decouple the versioning logic problems of Move from the core of this PR which is to add new integer support. We can tackle the former in the broader sync.
> 
> > That being said, is there any reason to block this PR for that?
> > Y'all already have your own branch on Aptos. You could just wait until this new deserialization entry point is added to pull this into your branch?
> > I know things are a bit choppy at the moment and y'all are probably pulling (and maybe releasing) a bit frequently. But as long as all the things are fixed by your next release, it should be fine?
> 
> I think you guys have a very valid point. I'd be perfectly happy to leave out the new deserialization entry point for a future PR, but just want to make sure we're aligned on the general direction about backward compatibility (that is, the VM is responsible for guaranteeing it for a certain amount of time), which I think we are now, aren't we?

I think so yeah! Should be a very quick followup PR to just expose setting the max version. Though I imagine there might be some tweaking if it doesn't apply cleanly into Aptos (or whatever other adapter)

oxade:
@vgao1996 @tnowacki 
So we're good to land this?

oxade:
@vgao1996 

> Yes! Super excited to have these available in Move!

great, please add this versioning issue to the agenda doc and feel free to make the follow up PR also.

oxade:
Syntax highlighting going here
https://github.com/damirka/vscode-move-syntax/pull/4

PR #548, opened 2022-10-05
[compiler] Fix for #544

wrwg:
Fixing a bug in resolving builtins


runtian-zhou (approved):
Stamp to unblock. Let's add a test for the rewrite rule later.

wrwg:
> Stamp to unblock. Let's add a test for the rewrite rule later.

There is basically no existing infra in the entire repo to test this... But we have a test in aptos coming up, thats how I found the issue.

vgao1996 (approved):

PR #549, opened 2022-10-06
[cherry-pick] Safe type tags

wrwg:
## Motivation

This cherry picks a series of PRs from the `aptos` branch to secure recursive type tags. Type tags are limited in serialization and parsing to not exceed a nesting depth of 8. In addition, it significantly reduces size of TypeTag by boxing the content of a larger enum variant.

For details see the individual commits

## Test Plan

Tests added in the commits

### DIFF
--- a/language/move-core/types/src/safe_serialize.rs
+++ b/language/move-core/types/src/safe_serialize.rs
@@ -1,0 +1,69 @@
+// Copyright (c) The Diem Core Contributors
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Custom serializers which track recursion nesting with a thread local,
+//! and otherwise delegate to the derived serializers.
+//!
+//! This is currently only implemented for type tags, but can be easily
+//! generalized, as the the only type-tag specific thing is the allowed nesting.
+
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::cell::RefCell;
+
+pub(crate) const MAX_TYPE_TAG_NESTING: u8 = 9;
+
+thread_local! {
+    static TYPE_TAG_DEPTH: RefCell<u8> = RefCell::new(0);
+}
+
+pub(crate) fn type_tag_recursive_serialize<S, T>(t: &T, s: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+    T: Serialize,
+{
+    use serde::ser::Error;
+
+    TYPE_TAG_DEPTH.with(|depth| {
+        let mut r = depth.borrow_mut();
+        if *r >= MAX_TYPE_TAG_NESTING {
+            // for testability, we allow one level more
+            return Err(S::Error::custom(
+                "type tag nesting exceeded during serialization",
+            ));
+        }
+        *r += 1;
+        Ok(())
+    })?;
+    let res = t.serialize(s);
+    TYPE_TAG_DEPTH.with(|depth| {
+        let mut r = depth.borrow_mut();
+        *r -= 1;
+    });
+    res
+}
+
+pub(crate) fn type_tag_recursive_deserialize<'de, D, T>(d: D) -> Result<T, D::Error>
+where
+    D: Deserializer<'de>,
+    T: Deserialize<'de>,
+{
+    use serde::de::Error;
+    TYPE_TAG_DEPTH.with(|depth| {
+        let mut r = depth.borrow_mut();
+        // For testability, we allow to serialize one more level than deserialize.
+        if *r >= MAX_TYPE_TAG_NESTING - 1 {
+            return Err(D::Error::custom(
+                "type tag nesting exceeded during deserialization",
+            ));
+        }
+        *r += 1;
+        Ok(())
+    })?;
+    let res = T::deserialize(d);
+    TYPE_TAG_DEPTH.with(|depth| {
+        let mut r = depth.borrow_mut();
+        *r -= 1;
+    });
+    res
+}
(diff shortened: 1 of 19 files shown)


tnowacki:
Loving the boxing! Was thinking about just that when looking at `MoveTypeLayout` the other day. 

Just curious, how did you land on `8` for the maximum type depth?

wrwg:
> Just curious, how did you land on `8` for the maximum type depth?

@davidiw initially came up with this number, and we haven't really discussed it after ;-) I guess a little more than 8 might be better.

tnowacki:
I'd feel better if it was a bit higher than 8, especially since this isn't configurable right now.
I could _maybe_ see someone hitting this at 8. 
Very unlikely at 16.
Only malicious at 32 :p

davidiw:
8 seems ridiculous, but maybe not... 4 seems actually quite reasonable. I would love to see a deeply nested, yet useful type :)

tnowacki:
I do not want to arbitrarily restrict the language without some strong motivations. This will be a big change in the sorts of types that are expressible when interacting with storage. The current max type depth at runtime is set fairly high, where no reasonable code will likely hit it (`TYPE_DEPTH_MAX` is currently set to `256). 

The type depth here is an indirect measure of the resource intensity (memory usage, time it takes to parse, etc) of the TypeTag, so I'm not sure what problem we are trying to solve here. What I mean by this is something like `S<S<S<S<S<S<S<S<S<S<S<S<S<S<S<Z>>>>>>>>>>>>>>>` is going to be less intensive than 
a single struct with an absurdly long identifier. 

I'm assuming the concern is one of resource usage. If so, why not bound the length of the bytes before deserializing them? We could probably set that bound fairly high before it causes issues.

davidiw:
Most serializers / deserializers limit the amount of nesting one can enter. This is a defense in depth measure that will undoubtedly impact any service that allows a user to specify this as is and then internally passes around this data structure. I hadn't realized there might be contention here, we should have chatted more on the weekly sync.

wrwg:
I found the nesting of 8 actually also too low. I accepted it because we could always increase without a breaking change. 
Suggesting as a compromise to make this 12 now, specifically as we have decreased the TypeTag size via boxing, so we should be save.

sblackshear:
I don't have a strong conviction on a particular nesting value, but I do think we should check with @jolestar (on StarCoin) and @0o-de-lally (on 0L) to make sure whatever value we pick is not a breaking change for their deployed code

language/move-core/types/src/safe_serialize.rs
```diff
@@ -0,0 +1,69 @@
...
+//! Custom serializers which track recursion nesting with a thread local,
+//! and otherwise delegate to the derived serializers.
+//!
+//! This is currently only implemented for type tags, but can be easily
+//! generalized, as the the only type-tag specific thing is the allowed nesting.
+
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::cell::RefCell;
+
+pub(crate) const MAX_TYPE_TAG_NESTING: u8 = 13;
```
jolestar, line 14: How was the 13 chosen?

jolestar (approved):

jolestar:
> I don't have a strong conviction on a particular nesting value, but I do think we should check with @jolestar (on StarCoin) and @0o-de-lally (on 0L) to make sure whatever value we pick is not a breaking change for their deployed code (since the previous max was 256).

I think no one would use a type with a nesting depth of more than 13, but I need to scan through the chain data to ensure. 

What happens if the user construct a type in the Move code that is deeper than 13? Can't write to external storage?

language/move-core/types/src/safe_serialize.rs
wrwg: Unlucky number ;-)

wrwg:
> > I don't have a strong conviction on a particular nesting value, but I do think we should check with @jolestar (on StarCoin) and @0o-de-lally (on 0L) to make sure whatever value we pick is not a breaking change for their deployed code (since the previous max was 256).
> 
> I think no one would use a type with a nesting depth of more than 13, but I need to scan through the chain data to ensure.
> 
> What happens if the user construct a type in the Move code that is deeper than 13? Can't write to external storage?

I'm not sure whether we store type tags as BCS in storage, but if so, yes. More importantly, this prevents one from having a deeper nested type tag in transaction payloads. This is the source of attacks here. Not a user writes something but someone creates a malicious type tag in a transaction send to the system. See also our telegram discussion.

tnowacki:
> > > I don't have a strong conviction on a particular nesting value, but I do think we should check with @jolestar (on StarCoin) and @0o-de-lally (on 0L) to make sure whatever value we pick is not a breaking change for their deployed code (since the previous max was 256).
> > 
> > 
> > I think no one would use a type with a nesting depth of more than 13, but I need to scan through the chain data to ensure.
> > What happens if the user construct a type in the Move code that is deeper than 13? Can't write to external storage?
> 
> I'm not sure whether we store type tags as BCS in storage, but if so, yes. More importantly, this prevents one from having a deeper nested type tag in transaction payloads. This is the source of attacks here. Not a user writes something but someone creates a malicious type tag in a transaction send to the system. See also our telegram discussion.

Pinged on telegram for more background

0o-de-lally:
Hi Everyone, as I mentioned to @sblackshear 8 would not be an issue for us today. And anywhere between 8 and 16 seems like a generous upper-bound. That said the contention issues seem real but manageable within those limits.

wrwg:
Closing for now, will come back with another PR.

PR #553, opened 2022-10-07
[verifier] Hardening reference analysis

wrwg:
This aborts the abstract interpreter of the bytecode verifier at the first analysis error. Some analyses aren't robust otherwise. Also adds a test case.


rahxephon89 (approved):

runtian-zhou (approved):
Do we want to get a minimal sample so that we can get a cleaner test case?

PR #556, opened 2022-10-07
[verifier] Hardening reference analysis (cherry-pick #553)

wrwg:
This aborts the abstract interpreter of the bytecode verifier at the first analysis error. Some analyses aren't robust otherwise. Also adds a regression test which caused the discovery of this.

### DIFF
--- a/language/move-bytecode-verifier/src/absint.rs
+++ b/language/move-bytecode-verifier/src/absint.rs
@@ -104,8 +104,9 @@ pub trait AbstractInterpreter: TransferFunctions {
             let post_state = match self.execute_block(block_id, pre_state, function_view) {
                 Err(e) => {
                     block_invariant.post = BlockPostcondition::Error(e);
-                    next_block = function_view.cfg().next_block(block_id);
-                    continue;
+                    // Stop analyzing after the first error occurred, to avoid the risk of
+                    // crashes as followup errors.
+                    break;
                 }
                 Ok(s) => {
                     block_invariant.post = BlockPostcondition::Success;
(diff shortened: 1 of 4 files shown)


language/move-bytecode-verifier/src/absint.rs
```diff
@@ -104,8 +104,9 @@ pub trait AbstractInterpreter: TransferFunctions {
             let post_state = match self.execute_block(block_id, pre_state, function_view) {
                 Err(e) => {
                     block_invariant.post = BlockPostcondition::Error(e);
-                    next_block = function_view.cfg().next_block(block_id);
-                    continue;
+                    // Stop analyzing after the first error occurred, to avoid the risk of
+                    // crashes as followup errors.
+                    break;
```
sblackshear, line 109: Fix looks good.

In the future, I think we should return something like

```
enum AnalysisResult {
  Success,
  Failure(AnalysisError)
}
```

here instead of returning a partially completed invariant map. The partial map is no problem today (since all the call sites of `analyze_function` do anyway is iterate through the resulting invariant map), but it looks like there's no reason to let the caller see the invariant map or to have `BlockPostcondition::Error`. The abstract interpreter is a bit overdesigned for its simple use in the bytecode verifier (my fault).

sblackshear: Filed https://github.com/move-language/move/issues/557 for this

sblackshear (approved):

language/move-bytecode-verifier/src/absint.rs
wrwg: Good point! I did check the call sites to see whether the fix works, but did not recognize the simplification we could do here.

PR #559, opened 2022-10-08
[move-vm] Refactor compatibility

jolestar:
## Motivation

Refactoring the `Compatibility` to making the API more reasonable.

* Add `friend_linking` field to `Compatibility` and remove `check_friend_linking` argument for `check` fn.
* `Compatibility::check` to throw Error based on self config.  

Yes

## Test Plan

CI

### DIFF
--- a/language/move-binary-format/src/compatibility.rs
+++ b/language/move-binary-format/src/compatibility.rs
@@ -2,26 +2,32 @@
 // Copyright (c) The Move Contributors
 // SPDX-License-Identifier: Apache-2.0
 
+use std::collections::BTreeSet;
+
 use crate::{
+    errors::{PartialVMError, PartialVMResult},
     file_format::{AbilitySet, StructTypeParameter, Visibility},
     file_format_common::VERSION_5,
     normalized::Module,
 };
-use std::collections::BTreeSet;
+use move_core_types::vm_status::StatusCode;
 
 /// The result of a linking and layout compatibility check. Here is what the different combinations
 /// mean:
-/// `{ struct: true, struct_layout: true }`: fully backward compatible
-/// `{ struct_and_function_linking: true, struct_layout: false }`: Dependent modules that reference functions or types in this module may not link. However, fixing, recompiling, and redeploying all dependent modules will work--no data migration needed.
-/// `{ type_and_function_linking: true, struct_layout: false }`: Attempting to read structs published by this module will now fail at runtime. However, dependent modules will continue to link. Requires data migration, but no changes to dependent modules.
-/// `{ type_and_function_linking: false, struct_layout: false }`: Everything is broken. Need both a data migration and changes to dependent modules.
+/// `{ struct_and_function_linking: true, struct_layout: true, friend_linking: true }`: fully backward compatible
+/// `{ struct_and_function_linking: true, struct_layout: true, friend_linking: false }`: Backward compatible, exclude the friend module declare and friend functions
+/// `{ struct_and_function_linking: false, struct_layout: true, friend_linking: false }`: Dependent modules that reference functions or types in this module may not link. However, fixing, recompiling, and redeploying all dependent modules will work--no data migration needed.
+/// `{ struct_and_function_linking: true, struct_layout: false, friend_linking: true }`: Attempting to read structs published by this module will now fail at runtime. However, dependent modules will continue to link. Requires data migration, but no changes to dependent modules.
+/// `{ struct_and_function_linking: false, struct_layout: false, friend_linking: false }`: Everything is broken. Need both a data migration and changes to dependent modules.
 #[derive(PartialEq, Eq, Debug, Clone, Copy)]
 pub struct Compatibility {
-    /// If false, dependent modules that reference functions or structs in this module may not link
+    /// If false, dependent modules that reference functions or structs in this module may not link (exclude the friend functions)
     pub struct_and_function_linking: bool,
     /// If false, attempting to read structs previously published by this module will fail at
     /// runtime, or worse, may allow to re-interpret representations maliciously.
     pub struct_layout: bool,
+    /// If false, dependent modules that reference friend functions in this module may not link
+    pub friend_linking: bool,
 }
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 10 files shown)


wrwg (commented):
Will review later, but just want to point that doing something wrong here can jeopardize type safety so we must be extremely careful with this code.

language/move-binary-format/src/compatibility.rs
```diff
@@ -58,27 +64,64 @@ impl CompatibilityConfig {
...
+        }
+    }
+
     pub fn need_check_compat(&self) -> bool {
         // `check_friend_linking` is part of check_struct_and_function_linking, so ignore it at here.
         self.check_struct_and_function_linking || self.check_struct_layout
     }
+
+    /// Return compatibility assessment for `new_module` relative to old module `old_module`.
+    pub fn check(&self, old_module: &Module, new_module: &Module) -> PartialVMResult<()> {
```
sblackshear, line 91: Wondering if we can simplify this even further by unifying the `Compatibility` and `CompatibilityInfo` types. That is:

- Make `CompatibilityConfig::new` a function of `Compatibility`
- change the signature of `Compatibility::check` to `check(&self, old_module, new_module)` (no return value)
- Move the `return Err` login in this function inside `Compatibility::check`
- Get rid of the `CompatibilityInfo` struct + functions altogether; in the new design, you define a compatibility policy with `Compatibility::new`, then check it via `Compatibility::check`

jolestar: So, it looks like this:

```
#[derive(Debug, Clone, Copy)]
pub struct Compatibility {
    pub check_struct_and_function_linking: bool,
    pub check_struct_layout: bool,
    pub check_friend_linking: bool,
}

impl Compatibility{

  pub fun new(check_struct_and_function_linking: bool,
        check_struct_layout: bool,
        check_friend_linking: bool,
    ) -> Self;

  pub fn check(&self, old_module: &Module, new_module: &Module) -> PartialVMResult<()>;

}
```

But when to return `CompatibilityInfo`? add a new function to `Compatibility` like:

```
  pub fn check_for_info(old_module: &Module, new_module: &Module) -> CompatibilityInfo;
```

sblackshear: I thin we should no longer need `CompatibilityConfig`--all of the old clients can just use `Compatibility` directly.

jolestar: Got it `CompatibilityInfo` is a typo of `CompatibilityConfig`. But sometimes we need to get the `Compatibility` result but no error, such as https://github.com/move-language/move/pull/559/files#diff-8a42ae3ffa92bcdc1150d38c74c7d4feaacadab017ecd3e7934b52d7429f1896R350

sblackshear: For that particular call site, I think it's ok to do something like

```
if compat.check(&old_api, &new_api).is_err() {
  // check whether it's a layout or linking issue by constructing a new config and re-checking
  if Compatibility::new(true, false, false).is_err() {
    println!("Layout API for structs of module {} has changed. Need to do a data migration of published structs", module_id)
  }
  if Comptability::new(false, true, true).is_err() {
     println!("Linking API for structs/functions of module {} has changed. Need to redeploy all dependent modules.", module_id)
  }
}
```

jolestar: This is fine, but it feels a bit stupid to check Compatibility twice😂, I'll try it tomorrow.

sblackshear: Yep, this is definitely a bit silly :). But it's on an error path, and in service of making the safety-critical usages of the API as simple as possible.

jolestar: Refactor as the comment.

language/move-binary-format/src/compatibility.rs
```diff
@@ -58,27 +54,34 @@ impl CompatibilityConfig {
         }
     }
 
+    pub fn new(
```
sblackshear, line 57: No need to fix in this PR, but might be worth adding a TODO to expose a no-arg entrypoint for each valid config rather than a single entrypoint with 3 bools. That will let us eliminate the `assert` below (since we simplfy won't expose an entrypoint that allows `check_friend_linking` without `check_struct_and_function_linking`.

language/move-binary-format/src/compatibility.rs
```diff
@@ -174,27 +178,47 @@ impl Compatibility {
...
+        // - additions to the list are allowed
+        // - removals are not allowed
+        //
+        let old_friend_module_ids: BTreeSet<_> = old_module.friends.iter().cloned().collect();
+        let new_friend_module_ids: BTreeSet<_> = new_module.friends.iter().cloned().collect();
+        if !old_friend_module_ids.is_subset(&new_friend_module_ids) {
+            friend_linking = false;
+        }
+
+        //if struct_and_function_linking = false, friend_linking makes no sense.
```
sblackshear, line 200: I'm not sure I fully understand the need for this--is there a test that exercises this path?

sblackshear (commented):
This LGTM (thanks for the simplification), but would appreciate another set of eyes given the sensitivity here (perhaps @wrwg).

language/move-binary-format/src/compatibility.rs
jolestar: Now it is not needed, the previous version used this to avoid return

Compatibility{struct_and_function_linking:false, friend_linking:true }.

language/move-binary-format/src/compatibility.rs
jolestar: How about we change `check_struct_and_function_linking` to `check_struct_and_pub_function_linking`, so `check_struct_and_pub_function_linking` +  `check_friend_linking` = origin `check_struct_and_function_linking`.

And `new(check_struct_and_pub_function_linking:false, check_friend_linking: true)` seem be useless(maybe for only check the `friend`?), but the logic is valid.

language/move-binary-format/src/compatibility.rs
sblackshear: Can we delete this if it is no longer needed?

language/move-binary-format/src/compatibility.rs
```diff
@@ -2,40 +2,36 @@
...
-/// `{ type_and_function_linking: true, struct_layout: false }`: Attempting to read structs published by this module will now fail at runtime. However, dependent modules will continue to link. Requires data migration, but no changes to dependent modules.
-/// `{ type_and_function_linking: false, struct_layout: false }`: Everything is broken. Need both a data migration and changes to dependent modules.
+/// `{ check_struct_and_function_linking: true, check_struct_layout: true, check_friend_linking: true }`: fully backward compatible
+/// `{ check_struct_and_function_linking: true, check_struct_layout: true, check_friend_linking: false }`: Backward compatible, exclude the friend module declare and friend functions
+/// `{ check_struct_and_function_linking: false, check_struct_layout: true, check_friend_linking: false }`: Dependent modules that reference functions or types in this module may not link. However, fixing, recompiling, and redeploying all dependent modules will work--no data migration needed.
```
wrwg, line 20: Can you please point out that anything with check_struct_layout == false cannot ensure type safety?

`... -- no data migration needed. NOTE that if this is false, type safety over a series of upgrades cannot be guaranteed.`

wrwg (approved):
Fine with me if we land this, looks the core logic hasn't changed. Please address remaining reviewer comments (not sure whether everything was addressed).

language/move-binary-format/src/compatibility.rs
jolestar: I add `if check_struct_layout is false` line at the top of the comments.

language/move-binary-format/src/compatibility.rs
jolestar: Done

language/move-binary-format/src/compatibility.rs
jolestar: I do the change, and let `check_struct_and_pub_function_linking` and `check_friend_linking` no overlap.

Issue #557, opened 2022-10-07
[Feature Request] Simply abstract interpreter API's used in bytecode verifier 


sblackshear:
The abstract interpreter's `analyze_function` returns an invariant map annotated with post-states or errors for each block. This is an overkill of information for the client, which only needs to know whether the analysis succeeded or failed (and with which error). We should refactor to get rid of "error postconditions", and simplify the return logic

See https://github.com/move-language/move/pull/556 for more context.

tnowacki:
In terms of "simplification" I'm not sure if we have an issue for it, but we have discussed removing the abstract interpreter from the bytecode verifier. Where instead we just check provided annotations for the initial states of blocks (we technically only need it for loop heads, but... thats its own discussion).

I know simplifying the API is maybe nice without these other changes. Just thought it is worth mentioning somewhere, when discussing the simplifications we can do.

sblackshear:
It will take awhile to figure this out, but definitely worth looking into this as part of our longer-term plans. I'd be in favor of moving ahead with the simplications in the meantime (I think we're looking at ~1-2 PR's for that)

wrwg:
I just can say *super* careful making changes to the verifier. Best *don't* make any changes if it works (and is not a user facing surface you'd like to improve, or something terrible inefficient). Removing absint is something which Aptos would most likely oppose.

PR #561, opened 2022-10-08
[bytecode verifier] simplifications enabled by 'return on first …

sblackshear:
…error' discipline

92d1d29efa9ec458938599c6524a28287e44e2e5 fixed a bug in the reference safety verifier pass by returning the first error encountered instead of collecting all errors, then picking one. This change shows that the verifier is basically trying to implement a Rust `Result` in a very indirect way: aggregate a bunch of errors into an `InvariantMap`, let the client iterate through them and pick one, then (finally) turn it into a Rust `Result`.

This PR changes the abstract interpreter to use a Rust `Result` directly at the point where the error is discovered, via the following changes:
- Return a `Result` instead of an `InvariantMap` from `analyze_function`. All call sites were iterating through the `InvariantMap` to see if there was an error, but the caller will now get the error directly.
- Get rid of `BlockPostcondition`. This is no longer needed now that we return the error directly
- Get rid of the `AnalysisError` associated type and use `PartialVMResult` directly. This abstract interpreter is only intended for usage inside the bytecode verifier.

Any changes to the verifier are very sensitive, so this should be looked at very carefully. These changes do not need to go in quickly because they are not fixing any active problems in the verifier or elsewhere. They are made in the interest of making a key part of the verifier smaller/simpler for the long term.

Closes https://github.com/move-language/move/issues/557

### DIFF
--- a/language/move-bytecode-verifier/src/absint.rs
+++ b/language/move-bytecode-verifier/src/absint.rs
@@ -5,6 +5,7 @@
 use move_binary_format::{
     binary_views::FunctionView,
     control_flow_graph::{BlockId, ControlFlowGraph},
+    errors::PartialVMResult,
     file_format::{Bytecode, CodeOffset},
 };
 use std::collections::BTreeMap;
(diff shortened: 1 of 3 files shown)


language/move-bytecode-verifier/src/absint.rs
```diff
@@ -101,18 +81,9 @@ pub trait AbstractInterpreter: TransferFunctions {
...
-                    break;
-                }
-                Ok(s) => {
-                    block_invariant.post = BlockPostcondition::Success;
-                    s
-                }
-            };
+            // Note: this will stop analysis after the first error occurs, to avoid the risk of
+            // subsequent crashes
+            let post_state = self.execute_block(block_id, pre_state, function_view)?;
```
sblackshear, line 86: This is the meat of the PR--all of the other changes are in service of being able to use a `Result` here.

wrwg (approved):
This looks good to me.

tnowacki (commented):
Thanks!

PR #563, opened 2022-10-08
[Test-only] Add kani harness

rahxephon89:
## Motivation

This draft PR adds proof harness for kani. 

## Test Plan

cargo kani


language/move-bytecode-verifier/kani/src/lib.rs
```diff
@@ -0,0 +1,23 @@
+use move_bytecode_verifier::verify_module;
+use move_binary_format::file_format::Bytecode;
+
+#[cfg(kani)]
+#[kani::proof]
+#[kani::unwind(1)]
+fn test_get_successors() {
```
wrwg, line 7: I guess arbitrary bytecode would not satisfy this? Do we need to run a few phases of the bytecode verifier here to basically constraint the input until the first time the CFG is used?

language/move-bytecode-verifier/kani/src/lib.rs
```diff
@@ -0,0 +1,23 @@
...
+    let mut nops = vec![];
+    for _ in 0..opcode_num {
+        nops.push(kani::any());
+    }
+    Bytecode::get_successors(kani::any::<u16>(), &nops);
+}
+
+/*
+#[cfg(kani)]
+#[kani::proof]
```
wrwg, line 18: Too slow? I assume we need to provide bounds for loop unrolling.

language/move-binary-format/src/kani/mod.rs
```diff
@@ -0,0 +1,541 @@
...
+    impl kani::Arbitrary for ModuleHandle {
+        fn any() -> Self {
+            ModuleHandle {
+                address: kani::any::<AddressIdentifierIndex>(),
+                name: kani::any::<IdentifierIndex>(),
+            }
+        }
+    }
+
+    impl kani::Arbitrary for Bytecode {
```
wrwg, line 72: That is unfortunate -- that derive is really needed.

wrwg (commented):
Great start. I'll try to play with it. Anything I should know to get Kani working let me know.

language/move-bytecode-verifier/kani/src/lib.rs
rahxephon89: When I test it, I changed get_successors back to the previous code with wrong conditions to check whether the tool can catch it but so far there is no luck (32GB memory and 16 core). I am currently trying with different parameters including unwind.

language/move-bytecode-verifier/kani/src/lib.rs
rahxephon89: Yeah, I am trying different bounds now

language/move-bytecode-verifier/kani/src/lib.rs
rahxephon89: update: if I set unwind to 100, error can be found

wrwg:
Closing as this was test only. Feel free to re-open if needed.

PR #564, opened 2022-10-09
[fuzzing] Add fuzzing infra for bytecode verifier + fix bug

wrwg:
This adds a fuzzing crate and a few targets for `CompileModule` to the bytecode verifier. The rather rudimentary `cargo-fuzz` is used ([see here](https://rust-fuzz.github.io/book)) as the more sophisticated tools like `honggfuzz` don't work on Apple silicon. Using this (and other fuzzing tools) requires to derive `arbitray::Arbitray` (not to be mistaken with proptests `Arbitrary`) which allows to derive values from bit images, as the fuzzer produces.

This fuzzer infra was capable to find a recent (already fixed) bug in stack balancing in reference analysis, caused by continuing analysis after an error occured. It also discovered (for now) the following new bug: if the number of locals + parameters is larger than 255 certain analysis will go wrong, as `type LocalIndex = u8`. This was discovered since reference analysis uses `0..self.locals as LocalIndex` to iterate over locals, which reduces to `0..0`. It appeared that bounds check actually does not check whether the number of locals + parameters in range, which was added in this PR.

### DIFF
--- a/language/move-binary-format/src/check_bounds.rs
+++ b/language/move-binary-format/src/check_bounds.rs
@@ -11,14 +11,13 @@ use crate::{
     file_format::{
         AbilitySet, Bytecode, CodeOffset, CodeUnit, CompiledModule, CompiledScript, Constant,
         FieldHandle, FieldInstantiation, FunctionDefinition, FunctionDefinitionIndex,
-        FunctionHandle, FunctionInstantiation, ModuleHandle, Signature, SignatureToken,
+        FunctionHandle, FunctionInstantiation, LocalIndex, ModuleHandle, Signature, SignatureToken,
         StructDefInstantiation, StructDefinition, StructFieldInformation, StructHandle, TableIndex,
     },
     internals::ModuleIndex,
     IndexKind,
 };
 use move_core_types::vm_status::StatusCode;
-use std::u8;
 
 enum BoundsCheckingContext {
     Module,
@@ -347,7 +346,7 @@ impl<'a> BoundsChecker<'a> {
             .len()
             .saturating_add(parameters.len());
 
-        if locals_count > (u8::MAX as usize) + 1 {
+        if locals_count > LocalIndex::MAX as usize {
             return Err(verification_error(
                 StatusCode::TOO_MANY_LOCALS,
                 IndexKind::FunctionDefinition,
@@ -367,7 +366,7 @@ impl<'a> BoundsChecker<'a> {
         check_bounds_impl(self.view.signatures(), code_unit.locals)?;
 
         let locals = self.get_locals(code_unit)?;
-        // Use saturating add for stability; range checked above
+        // Use saturating add for stability
         let locals_count = locals.len().saturating_add(parameters.len());
 
         // if there are locals check that the type parameters in local signature are in bounds.
--- a/language/move-binary-format/src/file_format.rs
+++ b/language/move-binary-format/src/file_format.rs
@@ -43,8 +43,6 @@ use move_core_types::{
 };
 #[cfg(any(test, feature = "fuzzing"))]
 use proptest::{collection::vec, prelude::*, strategy::BoxedStrategy};
-#[cfg(any(test, feature = "fuzzing"))]
-use proptest_derive::Arbitrary;
 use ref_cast::RefCast;
 use serde::{Deserialize, Serialize};
 use std::ops::BitOr;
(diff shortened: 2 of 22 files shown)


language/move-binary-format/src/file_format.rs
```diff
@@ -62,6 +62,7 @@ macro_rules! define_index {
         #[derive(Clone, Copy, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
         #[cfg_attr(any(test, feature = "fuzzing"), derive(Arbitrary))]
         #[cfg_attr(any(test, feature = "fuzzing"), proptest(no_params))]
+        #[cfg_attr(any(test, feature = "arbfuzz"), derive(arbitrary::Arbitrary))]
```
msmouse, line 65: Can we roll up to line 63?

msmouse (approved):

language/move-binary-format/src/file_format.rs
msmouse, line 65: Can we roll up to line 63?

msmouse: scratch that, I'm dumb

rahxephon89 (approved):

language/move-binary-format/src/file_format.rs
wrwg: You are right this reads pretty bad, so I made it `derive(proptest_derive::Arbitrary)` for differentiation. Also now under the same `fuzzing` feature which does not hurt.

chen-robert:
Seems like this uses libfuzzer internally which should be pretty sophisticated actually

wrwg:
> Seems like this uses libfuzzer internally which should be pretty sophisticated actually

The UI is not that nice as honggfuzz, and also, it does not utilize cores available. But in fact you are right, this thing seems to have found issues as good or faster. (In addition to this PR, I run honggfuzz in parallel on a Linux machine.)

wrwg:
> > Seems like this uses libfuzzer internally which should be pretty sophisticated actually
> 
> The UI is not that nice as honggfuzz, and also, it does not utilize cores available. But in fact you are right, this thing seems to have found issues as good or faster. (In addition to this PR, I run honggfuzz in parallel on a Linux machine.)

Still we are looking at a mediocre coverage (15% or something). This kind of tech really need to be combined with symbolic execution (as in concolic execution). It is realistic to go much further than those existing fuzzer tools do today.

language/move-binary-format/src/check_bounds.rs
```diff
@@ -367,9 +367,15 @@ impl<'a> BoundsChecker<'a> {
         check_bounds_impl(self.view.signatures(), code_unit.locals)?;
 
         let locals = self.get_locals(code_unit)?;
-        // Use saturating add for stability; range checked above
+        // Use saturating add for stability
         let locals_count = locals.len().saturating_add(parameters.len());
 
+        // Locals and parameters together must fit into LocalIndex (u8)
+        if locals_count > LocalIndex::MAX as usize {
```
vgao1996, line 374: I think it should be `locals_count > LocalIndex::MAX as usize + 1`. For example, you could have 256 locals with the last one being referred to by index 255. 

There's also a dedicated error code for this: TOO_MANY_LOCALS.

For some reason the original check that was here disappeared...
```
       // check if the number of parameters + locals is less than u8::MAX
        let locals_count = locals.len() + parameters.len();

        if locals_count > (u8::MAX as usize) + 1 {
            return Err(append_err_info(
                VMStatus::new(StatusCode::TOO_MANY_LOCALS),
                IndexKind::FunctionDefinition,
                function_def_idx,
            ));
        }
```

wrwg: The check is actually still there, it is only in the calling function. Thanks for pointing out!

So the bug is actually this function in reference analysis:

```
   fn iter_locals(&self) -> impl Iterator<Item = LocalIndex> {
        0..self.num_locals as LocalIndex
    }
```

It is `0..0` if num_locals is 256... The iteration here should never happen in terms of `u8` but `usize`. I'm not sure whether we have other occasions of this bug, so I think the safest is to restrict the size as `255`. I removed the check I added and instead adapted this check.

dariorussi:
this is fantastic! 
I have not looked at anything yet :) but the results are super nice and I love tools that allow "random" generation and combinations of code.
Hoping to be able to look at this soon

PR #566, opened 2022-10-10
[stdlib] add `type_name` module for type reflection

sblackshear:
Move allows type-safe implementations of key types like fungible tokens using the Coin<T> standard. However, distinguishing coins at the type level only is sometimes limiting--e.g., you might want to write a DEX that has at most one pool for a given pair of currencies (need to compare types + reify the result in a value) or write a lending protocol that supports an extensible pool of currencies (need to reify the currency types + map them to the reserve coin supply).

This PR seeks to add the weakest form of reflection that supports use-cases like the ones above in order to keep static reasoning as simple/predictable as possible. Similar modules exist in the Starcoin and Aptos frameworks, and Sui users have also been requesting this feature.

- Add opaque `TypeName` value that can be produced from a Move type via `get<T>(): TypeName`
- Add functions for converting a `TypeName`'s into its inner representation as an ASCII strings so it can be compared or inspected. This is safe because the Move binary format enforces ASCII encoding for all identifiers

One issue worth discussing is whether we want to use the full-length representation of addresses (e.g., always produce a 16, 20, or 32 byte string, depending on the platform address length) or the "short" representation which strips leading 0's (e.g., the convention `0x1` for the Move stdlib). I think the former is slightly cleaner/more predictable, but I went with the latter because it already seems to be the standard in the `TypeTag` implementation of `Display`, as well as in the Aptos/Starcoin `TypeInfo` module that provides similar functionality to this code.


sblackshear:
See https://github.com/move-language/move/pull/566 in upstream Move, which we will pull in

language/move-stdlib/sources/type_name.move
```diff
@@ -0,0 +1,29 @@
+/// Functionality for converting Move types into values. Use with care!
```
emmazzz, line 1: You need to add license header here.

emmazzz (approved):

crazybits:
is it possible to get T from the type string in running time after add this feature?

e.g.

let T = ("uint64").Type;
let v = get<T>();

language/move-stdlib/src/natives/type_name.rs
```diff
@@ -0,0 +1,59 @@
...
+    gas_params: &GetGasParameters,
+    context: &mut NativeContext,
+    ty_args: Vec<Type>,
+    arguments: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert_eq!(ty_args.len(), 1);
+    debug_assert!(arguments.is_empty());
+
+    let type_tag = context.type_to_type_tag(&ty_args[0])?;
+    let type_name = type_tag.to_string();
```
jolestar, line 33: the `type_name` should be the full address or short address, this is affected by PR #441

wrwg:
> is it possible to get T from the type string in running time after add this feature?
> 
> e.g.
> 
> ```
> let T = ("uint64").Type;
> let v = get<T>();
> ```

This is a different feature which we may import from Aptos (called [`Any`](https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-stdlib/sources/any.move) over there).

language/move-stdlib/src/natives/type_name.rs
wrwg: It is essential that this is a canonical-forever name for type safety. I'd suggest to introduce a new function to_canonical string in type_tag to make this clear over there. Then we still have to choose w/wo leading 0s. I don't mind here.

wrwg (approved):

wrwg (changes requested):
Lets please wait with merge, just checking some more things...

jolestar:
> This is a different feature which we may import from Aptos (called [`Any`](https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/aptos-stdlib/sources/any.move) over there).

Good feature, expect import to stdlib.

language/move-stdlib/src/natives/type_name.rs
sblackshear: Very supportive of `to_canonical_string` here, will do!

sblackshear:
> is it possible to get T from the type string in running time after add this feature?
> 
> e.g.
> 
> ```
> let T = ("uint64").Type;
> let v = get<T>();
> ```

IIUC, this is actually two separate things:
1. The ability to construct a `TypeName` from a string
2. The ability to go back and forth between a `TypeName` value and a Move type

This PR very deliberately avoids (1), but you can have (2) without (1) (as in `Any`).

language/move-stdlib/sources/type_name.move
```diff
@@ -0,0 +1,29 @@
...
+        /// Struct types are represented as fully qualified, `0x`-prefixed short type names
+        /// `0x1::string::String` or
+        /// `0xabcd::module_name1::type_name1<0x1234::module_name2::type_name2<u64>>`
+        /// Note: addresses in short type names are *always* shortened by stripping leading 0's
+        /// e.g., 0xA rather than 0x00000000A,
+        name: String
+    }
+
+    /// Return a value representation of the type `T`.
+    public native fun get<T>(): TypeName;
```
tnowacki, line 18: What is the thought behind the name `get`?
My worry is making it clear that it is allocating something, not possibly fetching from a cache. `new` might be better here from that line of thinking.

language/move-stdlib/sources/type_name.move
```diff
@@ -0,0 +1,29 @@
...
+
+    /// Get the String representation of `self`
+    public fun borrow_string(self: &TypeName): &String {
+        &self.name
```
tnowacki, line 22: I am worried about exposing the `String` here, but if y'all are fine with fine by me. 

I think the root of my trepidation around the strings is it allows for cyclic thinking/checks with types. For example consider the difference between `t == TypeName::get<0x42::m::R>` vs `borrow_string(&t) == &ascii::new(b"0x42::m::R")`. The second can be used anywhere, where the first requires being able to link against `0x42::m`. 

But perhaps this cyclic style is desirable?

language/move-stdlib/sources/type_name.move
sblackshear: The reasoning is that you are "extracting" or "getting" the `TypeName` from `T` rather than constructing it (e.g., from a string). Trying to evoke that these come from real types.

language/move-stdlib/sources/type_name.move
sblackshear: Is your concern about both `borrow_string` and `into_string`, or only the former (guessing both, but wanted to double-check)?

I share the concern, and I think that the safest/most idiomatic usages of `TypeName`'s (and the ones easiest to verify with the prover) will look like your option (1). But I think hiding the string is a bit too paternalistic--if we're going to expose this form of reflection at all, I think we don't buy ourselves much by trying to limit its expressivity in this way.

sblackshear:
Added `to_canonical_string` + used it in the native function

language/move-stdlib/sources/type_name.move
tnowacki: Makes sense, I think should think about this more then as `T::type_name()` sorta, rather than `TypeName::new()`

language/move-stdlib/sources/type_name.move
wrwg: I don't see a problem in exposing a string. We have a very beautiful property specific to Move: the same 'name' (string) is unique for this type (and its layout) forever (assuming only compatible upgrades). There is no reason to hide the string. C++/Rust do this because they cannot guarantee uniqueness over different processes. We can.

wrwg (approved):

language/move-core/types/src/account_address.rs
```diff
@@ -45,7 +45,12 @@ impl AccountAddress {
         Self(buf)
     }
 
-    pub fn short_str_lossless(&self) -> String {
+    /// Return a canonical string representation of the address
+    /// Addresses are 0x-prefixed, hex-encoded lowercase values with leading 0's stripped
+    /// e.g., 0xa, *not* 0x00000000a, 0xA, or 0x00000000A
```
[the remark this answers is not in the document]
sblackshear: Oops, I did not mean to change this--will revert to the old behavior. Thanks for flagging!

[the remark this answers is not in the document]
sblackshear: Fixed--you should not need to change anything on the StarCoin side. Apologies again for that oversight.

PR #567, opened 2022-10-11
[move] Restrict size of type terms

wrwg:
Restrict the number of allowed type terms in instantiation to 128. This should prevent that malicious code produces OOM attacks via type instantiation.

### DIFF
--- a/language/move-vm/runtime/src/loader.rs
+++ b/language/move-vm/runtime/src/loader.rs
@@ -1403,6 +1403,20 @@ impl<'a> Resolver<'a> {
             BinaryType::Module(module) => module.struct_instantiation_at(idx.0),
             BinaryType::Script(_) => unreachable!("Scripts cannot have type instructions"),
         };
+
+        // Before instantiating the type, count the # of nodes of all type arguments plus
+        // existing type instantiation.
+        // If that number is larger than MAX_TYPE_NODES, refuse to construct this type.
+        // This prevents constructing larger and lager types via struct instantiation.
+        // TODO: this should be revisited for performance and unification with MAX_VALUE_DEPTH
+        let mut sum_nodes: usize = 1;
+        for ty in ty_args.iter().chain(struct_inst.instantiation.iter()) {
+            sum_nodes = sum_nodes.saturating_add(self.loader.count_type_nodes(ty));
+            if sum_nodes > MAX_TYPE_NODES {
+                return Err(PartialVMError::new(StatusCode::VM_MAX_TYPE_DEPTH_REACHED));
+            }
+        }
+
         Ok(Type::StructInstantiation(
             struct_inst.def,
             struct_inst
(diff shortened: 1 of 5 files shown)


vgao1996 (approved):

language/move-vm/runtime/src/loader.rs
```diff
@@ -1403,6 +1403,20 @@ impl<'a> Resolver<'a> {
...
+        // Before instantiating the type, count the # of nodes of all type arguments plus
+        // existing type instantiation.
+        // If that number is larger than MAX_TYPE_NODES, refuse to construct this type.
+        // This prevents constructing larger and lager types via struct instantiation.
+        // TODO: this should be revisited for performance and unification with MAX_VALUE_DEPTH
+        let mut sum_nodes: usize = 1;
+        for ty in ty_args.iter().chain(struct_inst.instantiation.iter()) {
+            sum_nodes = sum_nodes.saturating_add(self.loader.count_type_nodes(ty));
+            if sum_nodes > MAX_TYPE_NODES {
+                return Err(PartialVMError::new(StatusCode::VM_MAX_TYPE_DEPTH_REACHED));
```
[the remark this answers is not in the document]
wrwg: I do not know what the consequences of a new code are, so I don't really want to do this. We can do this with the right solution in place.

language/move-vm/runtime/src/loader.rs
```diff
@@ -2339,6 +2354,15 @@ impl Loader {
...
     }
 
+    fn count_type_nodes(&self, ty: &Type) -> usize {
+        match ty {
+            Type::Vector(ty) => self.count_type_nodes(&ty).saturating_add(1),
+            Type::StructInstantiation(_, ty_args) => ty_args.iter().fold(1, |x, y| x.saturating_add(self.count_type_nodes(y))),
+            Type::Reference(t) | Type::MutableReference(t) => self.count_type_nodes(t).saturating_add(1),
+            _ => 1
+        }
+    }
```
davidiw, line 2364: This is an unbounded recursion that might be limited by other restrictions but seems risky. Instead could we pass in the value up to this point (e.g., tail recursion of sorts) and terminate if it exceeds a maxium?

wrwg: Made this iterative

language/move-vm/runtime/src/loader.rs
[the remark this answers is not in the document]
wrwg: Introduce new error

language/move-vm/runtime/src/loader.rs
wrwg: For the record, I think the termination check is unnecessary since the point here is that those types which lead to overflow can never be constructed. We already have multiple recursive algorithms which do no checks at all and expect the thing is sized in range. This is just one more. Its purpose is to disallow constructing types which are too large.

vgao1996: Yeah we have some other checks ensuring that VM does not create a type that's over 256 levels deep so this shouldn't be a problem, at least for now

PR #573, opened 2022-10-12
[bug fix] Fixed bug with natives in instantiation loop checker

tnowacki:
- Fixed bug with instantiation loop checker that filtered before enumerating

## Motivation

- Found this annoying bug when adding some native functions

## Test Plan

- I do not think it can be tested without lazy native loading


vgao1996 (approved):
I was a clown.

vgao1996:
@wrwg this is an interesting case. Let's think about what we should do in terms of backward comaptibility

PR #574, opened 2022-10-12
[move-vm] Add deep_check_is_borrowed for GlobalValue

tnowacki:
- Added API function for checking if there are outstanding borrows to a global value

## Motivation

- I am writing a set of native functions that mimic multiple transactions. When ending a transaction, I need to check if there are outstanding borrows on the value, and if there are abort. 

- I am a bit worried that this is the first API function that exposes the dynamic ref checks of `Rc`, but as this is just for GlobalValues (not all references), this doesn't feel that bad. 

## Test Plan

- New function is not used, bit hard in the current setup to test

### DIFF
--- a/language/move-vm/types/src/values/values_impl.rs
+++ b/language/move-vm/types/src/values/values_impl.rs
@@ -2068,6 +2068,67 @@ impl GlobalValueImpl {
             },
         }
     }
+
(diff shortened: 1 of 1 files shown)


tnowacki:
No longer sure I need this, closing for now

PR #576, opened 2022-10-13
[deps] update bcs to 1.4

sblackshear:
And use a workspace dep to enforce the same version everywhere.


bmwill (approved):

wrwg (approved):
Great. We can than hopefully soon remove our fork with the changes which are now in 1.4

PR #578, opened 2022-10-14
[1/2][binary/cfg] Relax backward-jump == loop back-edge requirement

amnn:
## Motivation

When generating a `VMControlFlowGraph`, identify loop heads and back-edges by finding back-edges in the depth-first traversal from the entry block.  This ensures that all the blocks in the CFG that are reachable at runtime are reachable during abstract interpretation.

This is roughly the algorithm proposed for finding widening points in Section 4.1 (Depth-first numbering) of "Efficient chaotic iteration strategies with widenings", Bourdoncle 1993.

Replaces the previous strategy of assuming a backward (in code offset) jump corresponds to a back-edge, which is something the Control Flow verifier was ensuring.  This assumption is no longer required, so this commit also removes references to that dependency.

Yes

## Test Plan

Newly introduced unit tests check that a CFG can be generated, even when the verifier would fail on the bytecode:

```
move/language/move-binary-format$ cargo test
```

### DIFF
--- a/language/move-binary-format/src/control_flow_graph.rs
+++ b/language/move-binary-format/src/control_flow_graph.rs
@@ -107,37 +107,77 @@ impl VMControlFlowGraph {
         let blocks = blocks;
         assert_eq!(entry, code_len);
 
+        // # Loop analysis
+        //
+        // This section identifies loops in the control-flow graph, picks a back edge and loop head
+        // (the basic block the back edge returns to), and decides the order that blocks are
+        // traversed during abstract interpretation (reverse post-order).
+        //
+        // The implementation is based on the algorithm for finding widening points in Section 4.1,
+        // "Depth-first numbering" of Bourdoncle [1993], "Efficient chaotic iteration strategies
+        // with widenings."
+        //
+        // NB. The comments below refer to a block's sub-graph -- the reflexive transitive closure
+        // of its successor edges, modulo cycles.
+
         #[derive(Copy, Clone)]
         enum Exploration {
             InProgress,
             Done,
         }
 
-        let mut loop_heads: Map<BlockId, Set<BlockId>> = Map::new();
         let mut exploration: Map<BlockId, Exploration> = Map::new();
         let mut stack = vec![ENTRY_BLOCK_ID];
+
+        // For every loop in the CFG that is reachable from the entry block, there is an entry in
+        // `loop_heads` mapping to all the back edges pointing to it, and vice versa.
+        //
+        // Entry in `loop_heads` implies loop in the CFG is justified by the comments in the loop
+        // below.  Loop in the CFG implies entry in `loop_heads` is justified by considering the
+        // point at which the first node in that loop, `F` is added to the `exploration` map:
+        //
+        // - By definition `F` is part of a loop, meaning there is a block `L` such that:
+        //
+        //     F - ... -> L -> F
+        //
+        // - `F` will not transition to `Done` until all the nodes reachable from it (including `L`)
+        //   have been visited.
+        // - Because `F` is the first node seen in the loop, all the other nodes in the loop
+        //   (including `L`) will be visited while `F` is `InProgress`.
+        // - Therefore, we will process the `L -> F` edge while `F` is `InProgress`.
+        // - Therefore, we will record a back edge to it.
+        let mut loop_heads: Map<BlockId, Set<BlockId>> = Map::new();
+
+        // Blocks appear in `post_order` after all the blocks in their (non-reflexive) sub-graph.
         let mut post_order = Vec::with_capacity(blocks.len());
 
         while let Some(block) = stack.pop() {
             match exploration.entry(block) {
                 Entry::Vacant(entry) => {
+                    // Record the fact that exploration of this block and its sub-graph has started.
                     entry.insert(Exploration::InProgress);
 
-                    // Push the block back on the stack to finish processing it once the sub-graph
-                    // reachable from it has been traversed.
+                    // Push the block back on the stack to finish processing it, and mark it as done
+                    // once its sub-graph has been traversed.
                     stack.push(block);
 
                     for succ in &blocks[&block].successors {
                         match exploration.get(succ) {
-                            // Frontier detected
+                            // This successor has never been visited before, add it to the stack to
+                            // be explored before `block` gets marked `Done`.
                             None => stack.push(*succ),
 
-                            // Back edge detected
+                            // This block's sub-graph was being explored, meaning it is a (reflexive
+                            // transitive) predecessor of `block` as well as being a successor,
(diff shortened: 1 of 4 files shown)


language/move-binary-format/src/unit_tests/control_flow_graph_tests.rs
```diff
@@ -0,0 +1,72 @@
...
+            Branch(1), // L2
+        ])
+    };
+
+    cfg.display();
+    assert_eq!(cfg.num_blocks(), 3);
+    assert_eq!(traversal(&cfg), vec![0, 2, 1]);
+}
+
+/// Return a vector containing the `BlockId`s from `cfg` in the order suggested by successiely
```
amnn, line 60: Note to self: typo.

language/move-binary-format/src/control_flow_graph.rs
```diff
@@ -112,56 +107,63 @@ impl VMControlFlowGraph {
...
+                    // reachable from it has been traversed.
+                    stack.push(block);
+
+                    for succ in &blocks[&block].successors {
+                        match exploration.get(succ) {
+                            // Frontier detected
+                            None => stack.push(*succ),
+
+                            // Back edge detected
+                            Some(Exploration::InProgress) => {
```
tnowacki, line 136: This intuitively makes sense to me. But is there some way we can sort of formalize the overall approach here in a comment block maybe. 

My proposed description, which might help check my understanding: "In this DFS approach, if you encounter an `Exploration::Done` node, then it means the node was visited via a different path, meaning there is a 'diamond' in the CFG. If however, you encounter an `Exploration:InProgress` node, then it means that one of the blocks predecessors is also a successor, meaning there is a 'loop' in the CFG. We greedily take this edge as the 'back-edge' in the loop"

tnowacki: Though it might be nice to note why you cannot possibly hit a `Done` in a loop, and why you cannot hit an `InProgress` in a diamond. I think it is at least quickly dismissed via contradiction, but if you have something constructive in mind, might be helpful too

amnn: I'm interpreting your question mainly as being related to loop heads, where the formal properties are that:

- For every loop in the CFG (that is reachable from the entry block), we assign at least one back edge (i.e. back edge if and only if loop)
- In reducible CFGs, the node that the back edge points to dominates all the other nodes in the loop (that's what makes it the head).

The first property's forward direction (back edge => loop) is explained by your description, for loop => back edge,  the sketch is: 

- Suppose there is a loop and it's reachable from the entry block.
- Consider the point at which we start to explore the first node in that loop, `F` as part of the traversal (that node is now in-progress, everything else is not present in the exploration).
- Because `F` is part of a loop, it will have an edge pointing to it, from a node `L` where there is a path from `F` to `L` as well.
- `F` will not transition to `Done` until all the nodes reachable from it (including `L`) have been visited, 
- Because `F` is the first node we are seeing in the loop, this means all the other nodes in the loop, (again, including `L`) will be visited while `F` is `in-progress`.
- Therefore we will process the `L -> F` edge while `F` is in progress.
- Therefore all loops will get a back edge.

The property that all reducible CFGs correctly identify a sensible loop head (the node that dominates the others in the loop) is a bit more fiddly, and it's covered in the Bourdoncle paper, so I may just refer to that... (this property comes from the fact that we are using a DFS from the entry block and reducible CFGs are defined by this "domination" property)

tnowacki: We should shove some condensed version of my comment and your description into this file. And point to the Bourdoncle paper. Luckily though, this section is far less critical than #579, so we should focus an heavy documents to that PR/verifier pass.

tnowacki:
> * [H1 … C1_1 … H2 … C1_2 … E2 … C1_3 … E1]

I'm a bit confused by this example. I'm not sure if it is allowed or not, but it still feels "nested" to me. 
A non-nested version of this to me would be `[H1 ... C1_1 ... H2 ... C1_2 ... E1 ... E2]`

We should wait for @amnn though

amnn:
Thanks for the reviews both!

> I am also curious on whether the property that "all looks are properly nested" still hold in this algorithm?

There's two angles to this question:

1. Does this algorithm require that all loops are well-nested?
2. Does this algorithm succeeding imply that all loops are well-nested?

Before this PR, the answer to (1) was yes (or the stronger property about back edges being backward jumps) and (2) was trivially yes, because of (1).  I.e. it was only valid to create a `VMControlFlowGraph` after the control flow verifier had succeeded, and so this process succeeding obviously implies well-nestedness, because the control flow verifier checks for a stronger property.

After this PR, the answer to (1) is no, and so (2) is also no, but it was never the purpose of `VMControlFlowGraph::new` to guarantee this property.  It used to rely on it, and it doesn't anymore, but the control flow verifier is still around to make sure these properties hold.

The question is more interesting in the context of #579, which relaxes the properties that the control flow verifier checks for while guaranteeing well-nestedness (or so I claim) -- maybe this is what you meant in the first place, so I'll go into that bit here as well.

For the examples you've shared, to make sure we're on the same page, I've drawn out the CFGs:

<details><summary><strong>Examples</strong></summary>

</details>

So all three of these examples are considered "well-nested".  The easiest check (for me) that a loop is reducible is if (and only if) for all back-edges $x \to y$, $y$ dominates $x$ in every depth-first spanning tree of the CFG.  In all these examples the DFS tree is also very simple (it's the "spine" of downward arrows running down their centres).  Because the only branches off the spine are back edges, it's very easy to see that the domination property holds (all the back edges travel back up the spine) -- this is the condition that the previous `VMControlFlowGraph` generation logic relied on, with additional constraints on forward edges and that the blocks are laid out "in spine order".

I've drawn the simplest example of a non-well-nested CFG that I can think of, in two ways, below:



I've annotated it in grey with the edges in the DFS tree (dotted), the loop (commutes arrow), the offending back-edge to a non-dominating node (crossed).  The version on the left doesn't suggest a particular order to lay the blocks out in, while the version on the right does, but they are the same CFG.  The issue with this example is that from the start block, you can jump to the start or the middle of its inner loop.

P.S. In case it was relevant, I was also initially confused by the examples, in particular what the difference was between a $C_{x,y}$ node and an $E_z$ node.  I've interpreted them both as conditional jumps because that's the only way to generate the sets of loops that you've listed @meng-xu-cs.

P.P.S. The fact that this PR opens up the possibility of generating a non-well-nested `VMControlFlowGraph` is relied upon in #579 which builds one and then checks whether it is well-nested.

tnowacki (approved):
Approving as I think this PR is mostly just an improvement (in that we are now relying on less).
The problem @meng-xu-cs is interested in is covered in #579

meng-xu-cs (approved):

Issue #496, opened 2022-09-23
[Bug]  INVALID_LOOP_BREAK occur when verifying a correct code


lerencao:
# 🐛 Bug

## To reproduce

**Code snippet to reproduce**
```move
module bug_report::test_case {
    public fun test_loop(a: u64): u64 {
        if (a == 2) {
            1
        } else if (a == 3) {
            let i = 0;
            let ret = 0;
            while (i < a) {
                ret = ret + i;
                i = i+1
            };
            ret
        }
        else {
            abort 10000
        }
    }
}
```

**Stack trace/error message**

`move  sandbox publish` output:

``` 
Publishing failed with unexpected error INVALID_LOOP_BREAK
```

## Additional context

related issue https://github.com/diem/diem/issues/8220

tnowacki:
This is a good bug. Thanks for finding and reporting!

jolestar:
This bug looks so familiar, and I realized that I reported this bug a year ago as well

tnowacki:
> This bug looks so familiar, and I realized that I reported this bug a year ago as well

Do you have a link?

tnowacki:
To be clear, this is likely a bug with the compiler, not with the bytecode verifier.

amnn:
I had a look at the disassembly for the test case, and I can see why the verifier is failing (explained below).  There may be potential to improve the verifier here, as well as the compiler, depending on the intent behind this check which someone else will have to fill me in on.

## Issue

Looking at the disassembly or CFG (below) there's only one back edge: an unconditional branch from offset 30 to 17.  `check_breaks` makes sure that if you jump out of a loop, you only jump one instruction out (a `break`), but this conflicts with the unconditional branch from offset 21 to 33 which exits the loop.

## Motivation for this check

My understanding from reading the code is that `control_flow.rs` tries to check two properties (also shared as diagrams, below):
1. Loops are well nested (`check_continues`/`check_breaks`/`check_no_loop_splits`)
2. Loops don't nest too deeply (`check_loop_depth`)

These conditions are properties of the code's basic blocks but are written in terms of code offsets which imposes an additional layout assumption (i.e. that the compiler lays blocks out in a [weak-topological ordering](https://pages.cs.wisc.edu/~elder/stuff/bourdoncle.pdf)).

My thinking is that the additional layout assumption is not necessary for correctness, but that's where I need help from folks familiar with the history of the project (cc @tnowacki, @sblackshear).

[...]

PR #579, opened 2022-10-14
[2/2][verify/cfg] Relax control_flow verifier

amnn:
## Motivation

Closes https://github.com/move-language/move/issues/496

Replace the checks for invalid break, continue, loop split and depth overflow with a test for CFG reducibility based on Tarjan's Reducibility Algorithm [0].

This test preserves the properties that we care about for efficient abstract interpretation, which can be equivalently described as:

1. The CFG is reducible.
2. The CFG has a unique depth-first spanning tree.
3. The body of every loop in the CFG is dominated by its head.

But it is more lenient than the previous check in that it allows backwards jumps in bytecode that aren't continues and forward jumps (breaks) that can escape multiple loops.  As such, it also eliminates the `INVALID_BREAK` and `INVALID_CONTINUE` status codes, and only reports `INVALID_LOOP_SPLIT` and `LOOP_MAX_DEPTH_REACHED`.

[0]: Testing flow graph reducibility, R Tarjan, 1974

Yes

## Test Plan

New and old control_flow unit tests:

```
bytecode-verifier-tests$ cargo test control_flow_tests
```

New regression test for the code example described in the issue:

```
transactional-tests$ cargo test control_flow/regression_test_496
```

The regression tests and the new tests demonstrating patterns that are now okay are confirmed to have failed before this PR.


amnn:
(Don't be alarmed by the +450 lines -- these are mostly new tests, the new implementation is about the same length as the old one, if you don't include comments).

language/move-bytecode-verifier/src/control_flow.rs
```diff
@@ -7,268 +7,82 @@
...
-            Bytecode::Branch(target) | Bytecode::BrTrue(target) | Bytecode::BrFalse(target)
-                if is_back_edge(cur_instr, *target) =>
-            {
-                let (cur_loop_head, _last_continue) = loop_stack.last().unwrap();
-                if target != cur_loop_head {
-                    // Invalid back jump. Cannot back jump outside of the current loop
-                    Err(context.error(StatusCode::INVALID_LOOP_CONTINUE, cur_instr))
-                } else {
-                    Ok(())
+        let mut frontier: Vec<_> = body.iter().cloned().collect();
```
tnowacki, line 65: `iter().copied().collect()` should work?

language/move-bytecode-verifier/src/control_flow.rs
```diff
@@ -7,268 +7,82 @@
...
-                match loop_stack.last() {
-                    Some((_cur_loop_head, last_continue))
-                        if target > last_continue && *target != last_continue + 1 =>
-                    {
-                        // Invalid loop break. Must break immediately to the instruction after
-                        // the last continue
-                        Err(context.error(StatusCode::INVALID_LOOP_BREAK, cur_instr))
-                    }
-                    _ => Ok(()),
+                if pred != head && body.insert(pred) {
```
tnowacki, line 73: minor nit, I am a simpleton that can never remember the bool orderings around insert. So maybe `let was_inserted = prev != head && bool.insert(pred)` would help me

language/move-bytecode-verifier/src/control_flow.rs
```diff
@@ -7,268 +7,82 @@
...
-    context: &ControlFlowVerifier,
-    labels: &[Label],
-    mut check: F,
-) -> PartialVMResult<()> {
-    let mut loop_stack: Vec<(CodeOffset, CodeOffset)> = vec![];
-    for (cur_instr, instr, label) in context.labeled_code(labels) {
-        // Add loop to stack
-        if let Label::Loop { last_continue } = label {
-            loop_stack.push((cur_instr, *last_continue));
+    for head in summary.preorder().rev() {
```
tnowacki, line 50: Some extensive comments here would really help me. Took me about an hour to figure out what was going on lol

tnowacki: The code is very clean! But a bit dense, if you get me

tnowacki (commented):
We will probably need to discuss version gating this (i.e. running the old logic on < V6), but overall I like the change!

tnowacki (commented):
And do we know the time/space complexity of the new verify_reducibility?

language/move-bytecode-verifier/src/loop_summary.rs
```diff
@@ -0,0 +1,276 @@
...
+///   - back edges (going backward from descendant to ancestor in the spanning tree)
+///   - ... everything else.
+/// - Calculate which node is descendant of which in the DFST
+/// - Mapping nodes in the summary back to the blocks they originate from in the CFG.
+///
+/// This is used to implement Tarjan's Loop Reducibility algorithm (Tarjan 1974).
+pub struct LoopSummary {
+    /// Original block corresponding to this node, useful for recovering code offsets, e.g. for
+    /// error messages.
+    blocks: Vec<BlockId>,
```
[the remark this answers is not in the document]
amnn: I could create a type alias for `Vec`, i.e. `type NodeMap<T> = Vec<T>`.

language/move-bytecode-verifier/src/loop_summary.rs
```diff
@@ -0,0 +1,276 @@
...
+            preds,
+        }
+    }
+
+    /// Decides whether `descendant` is a descendant of `ancestor` in the depth-first spanning
+    /// tree.
+    pub fn is_descendant(&self, NodeId(ancestor): NodeId, NodeId(descendant): NodeId) -> bool {
+        // All the descendants of `ancestor` in the DFST will have the IDs immediately following
+        // it.
+        ancestor <= descendant && descendant <= ancestor + self.descs[ancestor as usize]
```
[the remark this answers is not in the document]
amnn: > If you had something like `descendant <= max_descendent(ancestor)` that would make sense to me, but it doesn't feel like that's what this check is

That is what this check is, i.e. `max_descendant(ancestor) = ancestor + self.descs[ancestor as usize]`.  We get this nice property because of what the comment above this line mentions:

```
// All the descendants of `ancestor` in the DFST will have the IDs immediately following
// it.
```

This, combined with the the fact that `self.descs[n]` is the number of transitive descendants of `n` gives us everything we need.

amnn:
> And do we know the time/space complexity of the new verify_reducibility?

This should run in `O(B lg* B)` where `B` is the number of basic blocks, and `lg*` is the [iterated logarithm](https://www.wikiwand.com/en/Iterated_logarithm), which grows extremely slowly.  For all practical purposes, it's linear in `B`.

Technically to achieve this runtime, all the uses of `BTree` should be replaced with `Hash`-based data structures, but given how dense the keys (`BlockId` and `NodeId`) are, I think this will make performance worse, even if it is asymptotically better.

tnowacki (changes requested):
I think we are in a decent spot for this! 

But we will need to gate it so the old logic is still used for <= V5 modules. Which will be a pain... so maybe just copy the old module?

amnn:
Should be all ready to go now!  Add documentation on what the reducibility check is doing, why and where it's from, and re-introduced the v5 versions of the tests (and introduced some tests of the backward compatibility as well).

tnowacki (approved):
I will want to look through the test cases in `loop_summary` a bit more closely, but I think this is good!

tnowacki:
But before we land, maybe @meng-xu-cs will want to take a look?

meng-xu-cs (approved):
Thank you Ashok for the detailed explanations and both PRs. I believe this is good to go!

PR #585, opened 2022-10-16
[test] Additions to txn testing framework to handle strings

awelc:
## Motivation

This PR introduced support for UTF8 strings to the command line options parser, thus enabling its use in the transactional testing framework.

Yes


language/move-command-line-common/src/values.rs
```diff
@@ -215,6 +217,29 @@ impl Token for ValueToken {
                 }
                 (ValueToken::ByteString, len)
             }
+            's' if matches!(chars.peek(), Some('"')) => {
```
tnowacki, line 220: I'm fine with just no prefix here. Or does that break things?

awelc: Done (does not seem to breaking anything!)

tnowacki (approved):

PR #587, opened 2022-10-17
[cherry-pick] Fuzzing setup & other smaller things

wrwg:
Cherry-picks a few PRs from the aptos branch, specifically the one which adds the fuzzing logic. Also aligns the regression tests for the bytecode verifier.


sblackshear (approved):
Thanks!

PR #593, opened 2022-10-18
[model] allow conversion of model reference types into normalized types

sblackshear:
A Move model `Type` can be converted into a `normalized::Type`, but the logic for doing this previously refused to normalize a type behind a reference.

### DIFF
--- a/language/move-model/src/ty.rs
+++ b/language/move-model/src/ty.rs
@@ -413,10 +413,16 @@ impl Type {
                 env.get_struct_type(mid, sid, &ts),
             Vector(et) => Some(MType::Vector(
                 Box::new(et.into_normalized_type(env)
-                    .expect("Invariant violation: vector type argument contains incomplete, tuple, reference, or spec type"))
+                    .expect("Invariant violation: vector type argument contains incomplete, tuple, or spec type"))
             )),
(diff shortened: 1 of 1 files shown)


language/move-model/src/ty.rs
```diff
@@ -413,10 +413,16 @@ impl Type {
                 env.get_struct_type(mid, sid, &ts),
             Vector(et) => Some(MType::Vector(
                 Box::new(et.into_normalized_type(env)
-                    .expect("Invariant violation: vector type argument contains incomplete, tuple, reference, or spec type"))
+                    .expect("Invariant violation: vector type argument contains incomplete, tuple, or spec type"))
             )),
+            Reference(r, t) =>
+                if r {
```
tnowacki, line 419: Nit, I'd probably do something like 
```
let inner = Box::new(t.into_normalized_type(env).expect("Invariant violation: reference type contains incomplete, tuple, or spec type"));
Some(if r { MType::MutableReference(inner) } else { MType::Reference(inner) })
```
but this is fine as is

tnowacki (approved):

oxade:
@sblackshear out of curiosity, why was this not allowed before?

PR #598, opened 2022-10-19
[RFC][move-vm] Add a paranoid mode in the VM

runtian-zhou:
## Motivation

This is the refined version of #493 and #509 and hopefully we can embrace the idea of runtime checks and merge it into main.

The main idea is the following: provide a runtime-configurable mode in MoveVM that checks the resource safety properties of Move. Specifically we are trying to prove that with this paranoid mode on, each `move_vm_types::Value` will have the following properties guaranteed:
1. Type Safety:
    - All primitive values are well typed. e.g: u8, u64, u128, bool
    - Structs are well typed iff all its field elements are well typed and matches the field declaration of the struct
    - Vectors are well typed iff all elements in the vector have the same type that matches with the vector declaration.

2. Ability Safety: Consumption of all `Value` needs to abide to the ability declaration of the value.
3. Encapsulation Safety: 
    - `Struct` can only be created, destructed by the module that defines the struct. Same constraint applies for global operations and borrow_field.
    - Modules can only be invoked by other modules via its public interfaces. Moreover, arguments of the public function invocation needs to match with the type declaration of arguments.

The paranoid checks aims to achieve the three properties by tagging struct with its type and ability information at runtime. We still need to verify if all the checks here is a complete in order to prove whether the claim here is correct or not.

To review this PR, it might makes sense to get started by looking under the changes in the `move-vm-types`, particularly the additional runtime checker functions in `value.rs`. Then look for changes in `move-vm-runtime` where those checks are actually invoked. Once you have a better understanding of what we are checking, look for test cases under `paranoid-test` crate to see the behaviors there.

Yes

## Test Plan

Added a full testsuite called `paranoid-test`. In this test, I used `failpoint` injection to bypass the bytecode verifier pass in the VM and published modules that intensionally violates the three properties mentioned above and make sure the newly introduced paranoid mode can indeed capture those issues.

### DIFF
--- a/language/move-binary-format/src/file_format.rs
+++ b/language/move-binary-format/src/file_format.rs
@@ -651,6 +651,10 @@ impl AbilitySet {
             | (Ability::Key as u8),
     );
 
+    pub fn singleton(ability: Ability) -> Self {
+        Self(ability as u8)
+    }
+
     pub fn has_ability(self, ability: Ability) -> bool {
         let a = ability as u8;
         (a & self.0) == a
--- a/language/move-core/types/src/value.rs
+++ b/language/move-core/types/src/value.rs
@@ -74,6 +74,12 @@ impl MoveFieldLayout {
 pub enum MoveStructLayout {
     /// The representation used by the MoveVM
     Runtime(Vec<MoveTypeLayout>),
+    /// The representation used by the MoveVM with extra unique identifier for distinct type.
+    CheckedRuntime {
+        fields: Vec<MoveTypeLayout>,
+        tag: u64,
+        ability: u8,
+    },
     /// A decorated representation with human-readable field names that can be used by clients
     WithFields(Vec<MoveFieldLayout>),
     /// An even more decorated representation with both types and human-readable field names
@@ -258,6 +264,11 @@ impl MoveStructLayout {
     pub fn fields(&self) -> &[MoveTypeLayout] {
         match self {
             Self::Runtime(vals) => vals,
+            Self::CheckedRuntime {
+                fields,
+                tag: _,
+                ability: _,
+            } => fields,
             Self::WithFields(_) | Self::WithTypes { .. } => {
                 // It's not possible to implement this without changing the return type, and some
                 // performance-critical VM serialization code uses the Runtime case of this.
@@ -270,9 +281,36 @@ impl MoveStructLayout {
     pub fn into_fields(self) -> Vec<MoveTypeLayout> {
         match self {
             Self::Runtime(vals) => vals,
+            Self::CheckedRuntime {
+                fields,
+                tag: _,
+                ability: _,
+            } => fields,
             Self::WithFields(fields) | Self::WithTypes { fields, .. } => {
                 fields.into_iter().map(|f| f.layout).collect()
             }
+        }
+    }
+
+    pub fn tag(&self) -> Option<u64> {
+        match self {
+            Self::CheckedRuntime {
+                fields: _,
+                tag,
+                ability: _,
+            } => Some(*tag),
+            Self::Runtime(_) | Self::WithFields(_) | Self::WithTypes { .. } => None,
+        }
+    }
+
+    pub fn ability(&self) -> Option<u8> {
+        match self {
+            Self::CheckedRuntime {
+                fields: _,
+                tag: _,
+                ability,
+            } => Some(*ability),
+            Self::Runtime(_) | Self::WithFields(_) | Self::WithTypes { .. } => None,
         }
     }
 }
@@ -393,7 +431,12 @@ impl<'d> serde::de::DeserializeSeed<'d> for &MoveStructLayout {
         deserializer: D,
     ) -> Result<Self::Value, D::Error> {
         match self {
-            MoveStructLayout::Runtime(layout) => {
+            MoveStructLayout::Runtime(layout)
+            | MoveStructLayout::CheckedRuntime {
+                fields: layout,
+                tag: _,
+                ability: _,
+            } => {
                 let fields =
                     deserializer.deserialize_tuple(layout.len(), StructFieldVisitor(layout))?;
                 Ok(MoveStruct::Runtime(fields))
@@ -510,7 +553,12 @@ impl fmt::Display for MoveStructLayout {
     fn fmt(&self, f: &mut fmt::Formatter) -> std::fmt::Result {
         write!(f, "{{ ")?;
         match self {
-            Self::Runtime(layouts) => {
+            Self::Runtime(layouts)
+            | Self::CheckedRuntime {
+                fields: layouts,
+                tag: _,
+                ability: _,
+            } => {
                 for (i, l) in layouts.iter().enumerate() {
                     write!(f, "{}: {}, ", i, l)?
                 }
@@ -561,7 +609,13 @@ impl TryInto<StructTag> for &MoveStructLayout {
     fn try_into(self) -> Result<StructTag, Self::Error> {
         use MoveStructLayout::*;
         match self {
-            Runtime(..) | WithFields(..) => bail!(
+            Runtime(..)
+            | CheckedRuntime {
+                fields: _,
+                tag: _,
+                ability: _,
+            }
+            | WithFields(..) => bail!(
                 "Invalid MoveTypeLayout -> StructTag conversion--needed MoveLayoutType::WithTypes"
             ),
             WithTypes { type_, .. } => Ok(type_.clone()),
--- a/language/move-vm/paranoid-tests/tests/encapsulation_safety/borrow_field.mvir
+++ b/language/move-vm/paranoid-tests/tests/encapsulation_safety/borrow_field.mvir
@@ -1,0 +1,37 @@
+//# publish
+module 0x2.A {
+  struct C { x: u64 }
+
+  public make(): Self.C {
+  label b0:
+    return C { x: 0};
+  }
+}
+
+//# publish
+module 0x2.B {
+    import 0x2.A;
+    struct C { x: u64 }
+
+    public mut_field() {
+        let v: A.C;
+        let v_ref: &A.C;
+        let x: &mut u64;
+
+    label b0:
+        v = A.make();
+        v_ref = &v;
+        x = &mut move(v_ref).C::x;
+        *move(x) = 10;
+
+        return;
+    }
+}
+
+//# run --signers 0x1 --check-runtime-types
+import 0x2.B;
+main(account: signer) {
+label b0:
+    B.mut_field();
+    return;
+}
--- a/language/move-vm/paranoid-tests/tests/type_safety/call_mismatch.mvir
+++ b/language/move-vm/paranoid-tests/tests/type_safety/call_mismatch.mvir
@@ -1,0 +1,61 @@
+//# publish
+module 0x2.A {
+  struct A has key { b: Self.B }
+  struct B has store { x: u64 }
+  struct C has store { x: u64 }
+
+  public make(): Self.C {
+  label b0:
+    return C { x: 0};
+  }
+
+  public store(x: Self.B, s: &signer) {
+  label b0:
+    move_to<A>(move(s), A { b: move(x)});
+    return;
+  }
+}
+
+//# run --signers 0x1 --check-runtime-types
+import 0x2.A;
+main(account: signer) {
+    let v: A.C;
+
+label b0:
+    v = A.make();
+    A.store(move(v), &account);
+    return;
+}
+
+//# run --signers 0x1
+import 0x2.A;
+main(account: signer) {
+    let v: A.C;
+
+label b0:
+    v = A.make();
+    A.store(move(v), &account);
+    return;
+}
+
+//# run --signers 0x2 --check-runtime-types
+import 0x2.A;
+main(account: signer) {
+    let v: u64;
+
+label b0:
+    v = 0;
+    A.store(move(v), &account);
+    return;
+}
+
+//# run --signers 0x2
+import 0x2.A;
+main(account: signer) {
+    let v: u64;
+
+label b0:
+    v = 0;
+    A.store(move(v), &account);
+    return;
+}
--- a/language/move-vm/runtime/src/interpreter.rs
+++ b/language/move-vm/runtime/src/interpreter.rs
@@ -4,13 +4,14 @@
 
 use crate::{
     loader::{Function, Loader, Resolver},
+    move_vm::RuntimeConfig,
     native_functions::NativeContext,
     trace,
 };
 use fail::fail_point;
 use move_binary_format::{
     errors::*,
-    file_format::{Bytecode, FunctionHandleIndex, FunctionInstantiationIndex},
+    file_format::{Ability, AbilitySet, Bytecode, FunctionHandleIndex, FunctionInstantiationIndex},
 };
 use move_core_types::{
     account_address::AccountAddress,
@@ -67,6 +68,8 @@ pub(crate) struct Interpreter {
     operand_stack: Stack,
     /// The stack of active functions.
     call_stack: CallStack,
+    /// Whether to perform safety checks at runtime.
+    runtime_config: RuntimeConfig,
 }
 
 struct TypeWithLoader<'a, 'b> {
@@ -95,6 +98,7 @@ impl Interpreter {
         Interpreter {
             operand_stack: Stack::new(),
             call_stack: CallStack::new(),
+            runtime_config: loader.runtime_config(),
         }
         .execute_main(
             loader, data_store, gas_meter, extensions, function, ty_args, args,
@@ -124,7 +128,7 @@ impl Interpreter {
                 .map_err(|e| self.set_location(e))?;
         }
 
-        let mut current_frame = Frame::new(function, ty_args, locals);
+        let mut current_frame = Frame::new(function, ty_args, locals, self.runtime_config);
         loop {
             let resolver = current_frame.resolver(loader);
             let exit_code =
@@ -175,7 +179,18 @@ impl Interpreter {
                         )
                         .map_err(|e| set_err_info!(current_frame, e))?;
 
+                    if self.runtime_config.paranoid_type_checks {
+                        self.check_friend_or_private_call(&current_frame.function, &func)?;
+                        self.check_function_call_type_safety(
+                            &[],
+                            func.clone(),
+                            &func.get_resolver(loader),
+                        )
+                        .map_err(|err| self.set_location(err))?;
+                    }
+
                     if func.is_native() {
+                        let resolver = func.get_resolver(loader);
                         self.call_native(
                             &resolver,
                             data_store,
@@ -225,7 +240,18 @@ impl Interpreter {
                         )
                         .map_err(|e| set_err_info!(current_frame, e))?;
 
+                    if self.runtime_config.paranoid_type_checks {
+                        self.check_friend_or_private_call(&current_frame.function, &func)?;
+                        self.check_function_call_type_safety(
+                            &ty_args,
+                            func.clone(),
+                            &func.get_resolver(loader),
+                        )
+                        .map_err(|err| self.set_location(err))?;
+                    }
+
                     if func.is_native() {
+                        let resolver = func.get_resolver(loader);
                         self.call_native(
                             &resolver, data_store, gas_meter, extensions, func, ty_args,
                         )?;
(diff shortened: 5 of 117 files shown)


language/move-core/types/src/value.rs
```diff
@@ -69,6 +69,12 @@ impl MoveFieldLayout {
 pub enum MoveStructLayout {
     /// The representation used by the MoveVM
     Runtime(Vec<MoveTypeLayout>),
+    /// The representation used by the MoveVM with extra unique identifier for distinct type.
+    CheckedRuntime {
```
runtian-zhou, line 73: This is rather ugly.... IMHO the `Runtime` variant should never be here in the first place as why would we need anything related to runtime in the `move-core-types` crate? I understand this is because we had multiple representations of the same stuff but this was quite ugly if we want to add more runtime specific thing here due to the dependency issue. cc: @vgao1996

runtian-zhou (commented):
Also the `paranoid-test` suite right now focus mostly on the negative cases and we will need some positive cases as well to demonstrate it will execute code correctly when we have a well typed module

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -1019,11 +1238,21 @@ impl Frame {
...
+                            value.has_ability(Ability::Copy)?;
+                        }
+
                         interpreter.operand_stack.push(value)?;
                     }
                     Bytecode::WriteRef => {
                         let reference = interpreter.operand_stack.pop_as::<Reference>()?;
                         let value = interpreter.operand_stack.pop()?;
+
+                        if self.runtime_config.paranoid_hot_potato_checks {
```
sblackshear, line 1252:
```suggestion
                        if self.runtime_config.paranoid_type_checks {
```

sblackshear: This bug probably also means we're missing a test that exercises this case when `paranoid_hot_potato_checks` is off, but `paranoid_type_checks` is on

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -320,8 +394,43 @@ impl Interpreter {
         // Put return values on the top of the operand stack, where the caller will find them.
         // This is one of only two times the operand stack is shared across call stack frames; the other is in handling
         // the Return instruction for normal calls
-        for value in return_values {
-            self.operand_stack.push(value)?;
+
+        if self.runtime_config.paranoid_type_checks {
+            if return_values.len() != function.return_types().len() {
```
sblackshear, line 399: This is already checked at line 386

sblackshear (commented):
A question for `paranoid_type_checks`: did you consider  using a typed operand stack rather than embedding types in the value? The operand stack is the only thing in Move that is polymorphic--every local variable and instruction is strongly typed. If you have types for the operand stack, you can check the (dynamic) types of each operand against the static type of the instruction, as you execute it.

A few other properties of this:
- It is possible (and easy) to do dynamic type checks for *everything* (as @wrwg expressed a desire for in the last community meeting), not just structs. One discomfort I have with the current scheme is that there's no rigorous argument that the set of instructions that contain checks is sufficient to achieve the properties (1) - (3). And indeed, the PR description says "We still need to verify if all the checks here is a complete in order to prove whether the claim here is correct or not." I have lots of questions along these lines--e.g., why is it ok to perform paranoid type checks on `BorrowMut` and `BorrowImmut`, but not `VecBorrowMut` and `VecBorrowImmut`?. I think it is much easier to make an rigorous argument that a typed operand stack + checking types at every instruction is sufficient to achieve the properties we care about
- The in-memory overhead is reduced to O(values on the operand stack)) from O(# struct values * # fields/struct)
- No changes to the value representation are needed
- We would need to measure, but I think this is likely to be faster both because you never need to do "nested" checking on containers or structs--checking the top-level type of each operand against the instruction is enough
- This is fairly small change to the VM--add an extra field `typed_operand_stack: Vec<(Value, TypeId)>` to the VM. When the paranoid checks are off, this field is empty and `operand_stack` is used. When the paranoid checks are on, this field gets used and `operand_stack` is empty.

vgao1996:
@sblackshear just curious, if we switch to a type stack as you mentioned, then wouldn't it mean we'll need to modify every single clause of the giant match in the main interpreter loop so that the correct type transition is made for each single instruction?

sblackshear:
> @sblackshear just curious, if we switch to a type stack as you mentioned, then wouldn't it mean we'll need to modify every single clause of the giant match in the main interpreter loop so that the correct type transition is made for each single instruction?

Correct. This is what full type-safety means--every instruction is strongly typed, and works with operands of the expected type.

wrwg:
> > @sblackshear just curious, if we switch to a type stack as you mentioned, then wouldn't it mean we'll need to modify every single clause of the giant match in the main interpreter loop so that the correct type transition is made for each single instruction?
> 
> Correct. This is what full type-safety means--every instruction is strongly typed, and consumes operands of the expected type.
> 
> We could choose to skip type checks on some instructions if there's an argument made for some checks being "unimportant" (e.g., we could make the typed operand stack use `Option<Type>` and only insert types + perform checks for struct types and `signer`), or an argument that performing partial checks (as this PR is introducing) is equivalent to performing checks everywhere. But I think the simplest/safest approach is to check everything.

Doing an independent type stack, and relying on the types of values in locations are correct, dilutes IMO the impact of the paranoid checks. It basically introduces a similar indirection as the bytecode verifier already does, with the chance of making errors in this operand stack and/or wrong types for values in locations.  

What is the problem with storing types with values? And if we are concerned about performance, why can't we make that extra type info guarded by a feature flag? Storing the type with a value seems to me the most direct, and from the viewpoint of a paranoid mind, safest and comprehensive way to do we want here.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -1253,6 +1499,11 @@ impl Frame {
                     Bytecode::MoveToGeneric(si_idx) => {
                         let resource = interpreter.operand_stack.pop()?;
                         let signer_reference = interpreter.operand_stack.pop_as::<StructRef>()?;
+
+                        if self.runtime_config.paranoid_type_checks {
+                            signer_reference.check_inner_type(&Type::Signer)?;
+                        }
+
```
sblackshear, line 1506: I believe this also wants to check that the value being moved is not a hot potato.

sblackshear:
> What is the problem with storing types with values

I have no problem with this as approach to building a dynamic type checker. My primary issue with the current approach is the following (pasted from above);

> One discomfort I have with the current scheme is that there's no rigorous argument that the set of instructions that contain checks is sufficient to achieve the properties (1) - (3). And indeed, the PR description says "We still need to verify if all the checks here is a complete in order to prove whether the claim here is correct or not." I have lots of questions along these lines--e.g., why is it ok to perform paranoid type checks on `BorrowMut` and `BorrowImmut`, but not `VecBorrowMut` and `VecBorrowImmut`?. I think it is much easier to make a rigorous argument that a typed operand stack + checking types at every instruction is sufficient to achieve the properties we care about

The argument for the correctness of the typed operand stack approach goes something like

- At the beginning of execution, the operand stack is empty, and the only values that exist are in the formals of an `entry` function. The types of these values is checked by the adapter.
- A value can only appear on the operand stack by executing a typed instruction that places a value on the operand stack
- The input types to each instruction are checked dynamically before executing an instruction, and the type(s) of the instruction outputs either come from the (trusted, checked) input values, or from the static type of hte instruction.
- Types of storage locations (local variables and globals) can be trusted because they can only be populated via the typed operand stack.

I suggest the typed operand stack approach only mostly I haven't seen a rigorous argument for the correctness of the partial checks this PR currently does, and I don't know how to formulate one. I think if we're going to implement a dynamic type checker, we should have such an argument. I don't have particularly strong opinions about how the checker works under the hood, but I do feel strongly that we need to be able to argue for its correctness.

sblackshear:
> And if we are concerned about performance, why can't we make that extra type info guarded by a feature flag?

My comments about performance are about the paranoid checks themselves (which I would like to make as fast as possible, since they are clearly useful)

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -87,6 +90,10 @@ impl<'r, 'l, S: MoveResolver> TransactionDataCache<'r, 'l, S> {
 
             let mut resources = BTreeMap::new();
             for (ty, (layout, gv)) in account_data_cache.data_map {
+                if check_store {
```
sblackshear, line 93: Is this checking for `store` transitively, or only for `gv`? I.e., if I have something like:

```
struct HotPotato {}
struct X has store { h: HotPotato }
struct G has key { x: X }

fun try_move_to(account: &signer, g: G) {
  move_to(account, G)
}
```

will, this check discover the hot potato?

language/move-vm/runtime/src/loader.rs
```diff
@@ -999,6 +1011,18 @@ impl Loader {
             })
             .map_err(expect_no_verification_errors)?;
 
+        fail::fail_point!("verifier-failpoint-2", |_| { Ok(module.clone()) });
+
+        if self.runtime_config.paranoid_type_checks {
+            if &module.self_id() != id {
```
jolestar, line 1017: Maybe we should always check this, not dependent on `paranoid_type_checks`, in case the storage return the wrong module.

vgao1996:
> My primary issue with the current approach is the following (pasted from above);
>
> One discomfort I have with the current scheme is that there's no rigorous argument that the set of instructions that contain checks is sufficient to achieve the properties (1) - (3). And indeed, the PR description says "We still need to verify if all the checks here is a complete in order to prove whether the claim here is correct or not." I have lots of questions along these lines--e.g., why is it ok to perform paranoid type checks on BorrowMut and BorrowImmut, but not VecBorrowMut and VecBorrowImmut?. I think it is much easier to make a rigorous argument that a typed operand stack + checking types at every instruction is sufficient to achieve the properties we care about

I agree that it is important for us to understand the completeness of the check, but feel like this is an orthogonal issue, no? To me having a type stack or having the values carry some of the type info seem mathematically equivalent. Maybe it'll make you feel better if @runtian-zhou could give an informal proof that the existing checks implemented can indeed guarantee the properties listed?

language/move-vm/runtime/src/interpreter.rs
runtian-zhou: I think I'll just combine the two flags. The original idea is we have `paranoid_type_checks ` for type safety and `paranoid_hot_potato_checks ` for ability safety. However, `paranoid_hot_potato_checks ` won't work unless we turn `paranoid_type_checks` on because we need value tagging..

language/move-vm/runtime/src/interpreter.rs
runtian-zhou: We check this when we generate the writeset, where we can do a full check on the tree we want to write: 

i.e: the root of the tree has to have key capability and all the nested structs have store. I believe it's best for us to perform the checks there because inner nodes of the tree could potentially be mutated by write ref?

language/move-vm/runtime/src/data_cache.rs
runtian-zhou: You are right. I think we need to check the ability safety when packing elements.

runtian-zhou:
Updated the logic for checking ability safety for key/store. Now instead of checking the logic when we generate global values, we check the ability of inner fields when packing values. That was a great catch and I added a new test case for the example you provided cc @sblackshear .

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -979,9 +1120,41 @@ impl Frame {
...
+                            let ability = resolver.abilities(&value_ty)?;
+
+                            if arg_types.fields.len() != args.len() {
+                                return Err(PartialVMError::new(
+                                    StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR,
+                                )
+                                .with_message("pack args length mismatch".to_string()));
+                            }
+
+                            let inner_abilities = if ability.has_key() {
```
runtian-zhou, line 1136: cc @sblackshear

runtian-zhou:
This is an orthogonal issue. I think @tnowacki 's point is that in the current implementation, type information for value are obtained **from** the value, which could limit our ability to further optimize the value layout: if we strip off the type information in the value, the type information will need to flow **into** the value.

The biggest difference is that for the `borrow_local` and `borrow_field` set of operations, our current way of implementation can obtain the type information of references directly from the container. However, if we change to the type stack approach where we put type and value in a separate location, the runtime will need to manually provide the tag information for the values.

To me, I agree to what @wrwg suggests. Separating the types and values are diluting the means of this mode: right now the safety argument can be constructed cleanly around the value impl, i.e: **all value that are created by the runtime need to be well typed and should be used according its ability rule.** With this argument, we can then verify the statement from our rust implementation standpoint and from rust implementation alone, without needing to look from the Move standpoint. This seems like a nice, orthogonal way of specifying type safety instead of re-implementing all the type checking rules from the static type checker. Thus I would be more inclined to keep the values and types together without separating them apart. cc @sblackshear  @tnowacki.

Would love to hear from your thoughts on the safety argument that I just provided and why that's different from our static typing rules!

tnowacki:
> Separating the types and values are diluting the means of this mode: right now the safety argument can be constructed cleanly around the value impl

I think we should be precise here. "diluting" doesn't feel precise so let me try to highlight where I see the difference, and I do not think the difference is that important. 

So the goal here is to have a type associated with a value.  In any implementation, the requirement is "this dynamic type is the correct type for the value". The difference between having it with the `Value` vs with a separate type stack changes how this requirement is met.
Any implementation I can think of will likely share the same logic for creating initial types (constructing them based on the instruction being executed), so the difference lies only in the transistions. If you have it tied to the `ValueImpl`, as done in the PR, you have to show that the types are correctly threaded through the `ValueImpl`.
And with a separate type stack, the type stack transitions must match the value stack transitions, and also has the extra work when borrowing fields (but this is similar to the creation, with a twist). 

In short:
- Any paranoid interpreter is incorrect if the creation of types from bytecode instructions is incorrect.
- Tying the types to the `ValueImpl` is incorrect if there are bugs in the `ValueImpl` internal functions. 
- Tying the with a separate type stack is incorrect if the stack transitions are wrong.
  - It is also subtly wrong if there was some magic way to modify a struct without going through borrow field... but this hopefully isn't a concern. 
  
I think verifying that the stack transitions are correct seems easy enough, especially if you make a macro to pull from the value stack and type stack at the same time. 

I also think the separate type stack is a better long term solution as @runtian-zhou alludes to here

> I think @tnowacki 's point is that in the current implementation, type information for value are obtained from the value, which could limit our ability to further optimize the value layout: if we strip off the type information in the value, the type information will need to flow into the value.

If you roughly think of the value stack as `stack: Vec<Value>`, I would long term want the stack to be managed through directly for us, via `stack: Vec<u8>`. In such a setup, `Value` and `ValueImpl` would be gone (except maybe something similar would remain just to interface with native functions? though I'm not sure). 
A separate type stack in this way feels like something we could maintain and have around for a very long time (as long as the VM has a stack :p). 

TL;DR
- A separate type stack has the additional requirements that
  - Stack transitions are the same for the type and value stack. 
    - This feels easily isolated with a macro/function 
  - Borrow field + write ref is the only way to modify the internals of a struct (Struct values cannot be modified inside between pack and borrowfield or between pack and unpack)
    - This I think feels like something we can ignore? Since if some magic starts happening I don't think any guarantee we have can be considered safe (e.g. this feels as bad as something like an arbitrary code execution exploit) 
- A separate type stack feels like the best long term engineering decision.

sblackshear:
> right now the safety argument can be constructed cleanly around the value impl, i.e: all value that are created by the runtime need to be well typed and should be used according its ability rule

On this type safety argument: I can understand/agree with the the first part of this: "all values that are created by the runtime need to be well-typed". For this, enforcement on value creation and on value updates is sufficient, and the checks that are performed at the appropriate value creation instructions (`Pack`, `VecPack`) and value update instructions (`WriteRef`, `VecPushBack`) look correct to me.

However, "all values that are created by the runtime need to be well-typed" is not sufficient for type safety--Move instructions are also typed, and they also operate on values of the expected types. This is where I am unsure about the correctness of the current implementation, which only checks the types for a subset of instructions, and I can't understand the rationale for where checks are performed and where they are not (e.g., why `BorrowMut` and `BorrowImmut`, but not `VecBorrowMut` and `VecBorrowImmut`, as mentioned above)?

If you perform type checks for every instruction, or at every instruction that might read a struct, struct reference, or container type like `vector` that might directly or transitively contain a struct (since this approach is only looking at struct types), then I can see the structure of the correctness argument. But I think if the approach is to do type checks at a subset of the instructions, an argument for why this is safe must be made.

language/move-binary-format/src/file_format.rs
```diff
@@ -651,6 +651,10 @@ impl AbilitySet {
             | (Ability::Key as u8),
     );
 
+    pub fn singleton(ability: Ability) -> Self {
```
tnowacki, line 654: FWIW, this is just `AbilitySet::EMPTY | ability`, which I do quite often. But this API isn't bad :)

language/move-vm/paranoid-tests/tests/encapsulation_safety/borrow_field.mvir
```diff
@@ -0,0 +1,37 @@
...
+
+    public mut_field() {
+        let v: A.C;
+        let v_ref: &A.C;
+        let x: &mut u64;
+
+    label b0:
+        v = A.make();
+        v_ref = &v;
+        x = &mut move(v_ref).C::x;
```
runtian-zhou, line 24: In this case for example, `B` module is trying to borrow field from a resource defined in `A`. That's why we need an extra check for `BorrowField`. cc @sblackshear

language/move-vm/paranoid-tests/tests/type_safety/call_mismatch.mvir
```diff
@@ -0,0 +1,61 @@
...
+  struct A has key { b: Self.B }
+  struct B has store { x: u64 }
+  struct C has store { x: u64 }
+
+  public make(): Self.C {
+  label b0:
+    return C { x: 0};
+  }
+
+  public store(x: Self.B, s: &signer) {
```
runtian-zhou, line 12: e.g: This is a case where `call` instruction would fail. cc @sblackshear

language/move-vm/runtime/src/interpreter.rs
runtian-zhou: Is this rule the correct rule of checking store and key ability? @sblackshear

chen-robert:
> However, "all values that are created by the runtime need to be well-typed" is not sufficient for type safety--Move instructions are also typed, and they also operate on values of the expected types. To frame this differently: if I suggested that you remove the type checks for Call, you would presumably (and rightly!) object. But observe that these checks have no impact on your stated safety property "all values created by the runtime need to be well-typed", so there must be some other reason why they are important. That reason is because well-typed values is not enough on its own--Move instructions are also typed, and we must ensure that they operate on values of the correct type.

@sblackshear Agree that the argument could be clarified a bit better, but from my understanding, the idea is to use this definition of "well-typed" to build safety arguments for critical operations such as `Call`. Perhaps we could also include a similar definition for "type-safety" which asserts something like

> All bytecode instructions which operate on values will abort if 1. the value is not well-typed or 2. the type doesn't match what is "expected"

The reasoning behind checking `Call` is that type safety across the function call boundary is critical, while type safety internal to a module _may_ not be as critical? Because you would only be able to mess with / clobber internal values.

runtian-zhou:
After assessing this PR and #642, I feel like the separate type stack approach is a better solution:
1. It became harder to address the type confusion between reference and mutable reference in this approach: adding mutablility information to values soon became much harder in the value impl and start to contaminate the downstream value apis due to the extra information needed.

2. Handling return values for reference is also not trivial: current approach requires walking through the whole value tree, which is sth both expensive and hard to test here in general.

3. It is hard to prove the exhaustiveness of the tests, as @sblackshear points out previously.

The downside of the separate type and value stack approach are following:
1. We don't keep the information of the nested structs. Thus when inner nodes of the structs are being accessed, types of the inner nodes will come from the resolver rather than from the information when it was created. This could potentially dilute the value of this check as @wrwg suggests.
2. Ability information cannot be stored alongside the value so it will need to be resolved rather than stored.

For (2) we may address the issue by adding a type to ability cache in the paranoid mode. For (1) I would argue that this is not a bigger concern to me, as inner fields of struct can only be mutated via write_ref instruction. So as long as we type check the `write_ref` instruction the solution should be equivalent to the approach where types are stored alongside the value. cc @chen-robert 

Anyway I updated the alternative approach PR and the testsuites there and have a full test coverage of all instructions other than primitive instructions, which are already type checked by the value implementation. Looking forwards to your feedbacks!

sblackshear:
Think we can close this now that the alternative has landed.

PR #599, opened 2022-10-19
[prover] Adding a test to validate whether axiomatized natives work

wrwg:
This demonstrates that some native functions can be defined purely in Move as good as they would be defined in Boogie.

We'd need to re-implement generic axioms in presence of monomoprhization to make this also work for generic functions.


meng-xu-cs (approved):
Awesome! This is going to be extremely useful in the context of decoupling prover from any specific framework as well.

rahxephon89 (approved):

PR #601, opened 2022-10-19
[prover] Resurrect generic axioms for axiomatizing of native functions

wrwg:
Generic axioms have been disabled since we introduced monomorphization. This PR re-enables them:

- Type checker has been extended to recognize the new syntax `axiom<T> forall x:T: P`. Parser and AST was fortunately already supporting this.
- Monomorphization has been extended to find instantiations of generic axioms. In theory, one would have to find all the possible (generic) quantifier matches, simulating Z3's quantifier instantiation. This is indeed infeasible. The approximation is to instantiate the axioms _for all types used in the program_. If the axiom has more than one type parameter, the cartesian product of all types is used. To avoid explosion, axiom type parameters are restricted to length 2. This should work for all practical cases, but might need to be further refined down the road.


rahxephon89 (approved):

rahxephon89:
Thanks, LGTM

language/move-prover/tests/sources/functional/opaque_native.move
```diff
@@ -4,9 +4,13 @@ module 0x42::OpaqueNative {
         pragma aborts_if_is_strict = true;
 
         // Uninterpreted spec function with axiom.
-        // Current restrictions: we cannot deal with generic functions
         fun injection(v: vector<u8>): u64;
         axiom forall v1: vector<u8>, v2: vector<u8>: v1 == v2 <==> injection(v1) == injection(v2);
+
+        // Uninterpreted generic spec function with axiom.
+        fun generic_injection<T>(x: T): u64;
+        axiom<T> forall v1: T, v2: T: v1 == v2 <==> generic_injection(v1) == generic_injection(v2);
```
junkil-park, line 12: A minor point, but an injection only requires this direction `<==`. The other direction (`==>`) is related to the property of equality. I don't recall the exact name now ...

language/move-prover/tests/sources/functional/opaque_native.move
```diff
@@ -4,9 +4,13 @@ module 0x42::OpaqueNative {
         pragma aborts_if_is_strict = true;
 
         // Uninterpreted spec function with axiom.
-        // Current restrictions: we cannot deal with generic functions
         fun injection(v: vector<u8>): u64;
         axiom forall v1: vector<u8>, v2: vector<u8>: v1 == v2 <==> injection(v1) == injection(v2);
+
+        // Uninterpreted generic spec function with axiom.
+        fun generic_injection<T>(x: T): u64;
```
junkil-park, line 11: To specify the framework contracts, where should we put these unintrepreted functions and the axioms? In the "main(?)" hash module? Or, can there be a separate spec file to contain common spec functions and axioms?

junkil-park (approved):
Thanks!

language/move-prover/tests/sources/functional/opaque_native.move
wrwg: It depends. We could put some generic stuff together in one file; injection might be an example. In general just specific for spec file probably ok.

language/move-prover/tests/sources/functional/opaque_native.move
wrwg: Yeah, this is actually a bijection. It's how we have it also for `hash` in the bpl. Think we can leave for the test like this.

PR #602, opened 2022-10-19
add support for loading resources via NativeContext

alinush:
## Motivation

It is often necessary for Move native functions implemented in Rust to read resources. Currently, there is no easy way to do this in a native function because the `load_resource` method inside the `DataStore` in a `NativeContext` is not accessible.

This PR makes it accessible, together with a `load_type` method for converting a `TypeTag` to a `Type`, which is needed as an argument of `load_resource`.

Yes!

## Test Plan

Just exporting some existing APIs. No testing.

### DIFF
--- a/language/move-vm/runtime/src/loader.rs
+++ b/language/move-vm/runtime/src/loader.rs
@@ -893,7 +893,7 @@ impl Loader {
     pub(crate) fn load_type(
         &self,
         type_tag: &TypeTag,
-        data_store: &impl DataStore,
+        data_store: &dyn DataStore,
     ) -> VMResult<Type> {
         Ok(match type_tag {
             TypeTag::Bool => Type::Bool,
@@ -932,7 +932,7 @@ impl Loader {
     pub(crate) fn load_module(
         &self,
         id: &ModuleId,
-        data_store: &impl DataStore,
+        data_store: &dyn DataStore,
     ) -> VMResult<Arc<Module>> {
         self.load_module_internal(id, &BTreeMap::new(), &BTreeSet::new(), data_store)
     }
@@ -944,7 +944,7 @@ impl Loader {
         id: &ModuleId,
         bundle_verified: &BTreeMap<ModuleId, CompiledModule>,
         bundle_unverified: &BTreeSet<ModuleId>,
-        data_store: &impl DataStore,
+        data_store: &dyn DataStore,
     ) -> VMResult<Arc<Module>> {
         // if the module is already in the code cache, load the cached version
         if let Some(cached) = self.module_cache.read().module_at(id) {
@@ -975,7 +975,7 @@ impl Loader {
     fn load_and_verify_module(
         &self,
         id: &ModuleId,
-        data_store: &impl DataStore,
+        data_store: &dyn DataStore,
         allow_loading_failure: bool,
     ) -> VMResult<CompiledModule> {
         // bytes fetching, allow loading to fail if the flag is set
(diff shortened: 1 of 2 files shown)


language/move-vm/runtime/src/native_functions.rs
```diff
@@ -167,6 +170,18 @@ impl<'a, 'b> NativeContext<'a, 'b> {
         self.extensions
     }
 
+    pub fn load_type(&self, type_tag: &TypeTag) -> VMResult<Type> {
```
tnowacki, line 173: Interesting. What case do you not have a `Type`?

language/move-vm/runtime/src/loader.rs
```diff
@@ -893,7 +893,7 @@ impl Loader {
     pub(crate) fn load_type(
         &self,
         type_tag: &TypeTag,
-        data_store: &impl DataStore,
+        data_store: &dyn DataStore,
```
tnowacki, line 896: @wrwg or @vgao1996, thoughts here? 

I don't really care just wanted to flag

tnowacki (commented):
Generally looks fine, though I am curious why you would need `load_type`

Waiting to approve to give @vgao1996 or @wrwg a chance to review too

language/move-vm/runtime/src/native_functions.rs
alinush: Indeed, such an odd scenario would be the following:

Due to a circular dependency issue, I cannot `use` a module that defines the type of a resource I am interesting in reading. To deal with this, I would construct a `TypeTag` for that resource's type in the Rust native and use `load_type` on it. Then, I can read the resource using `load_resource`.

tnowacki: Do you have a code example? I'm not quite sure how you are loading a value without having to refer to it's type at some point. 

(Note: It's a native function for whatever adapter you are working on, so some core assumptions in Move could go out the window, but it is worth mentioning there is an assumption that you only load resources in the module that defines them)

alinush: Maybe this might not be such a good idea, then?

I don't have a code example but, as mentioned above, in my use case, I can easily construct a `StructTag` and hence a `TypeTag` for say a resource at `0x01::chain_id::ChainId`. I cannot get a `Type` though because I cannot `use` the `chain_id` module (due to a circular dependency issue), fetch the `ChainId` type and pass it down to a Rust native.

language/move-vm/runtime/src/loader.rs
wrwg: That is not good since all calls now go over dynamic dispatch. Why is this needed?

wrwg (commented):
Generally, I'd say we are better off putting the `chain_id` in the `NativeTransactionContext` of Aptos. This will be lesser code changes.

language/move-vm/runtime/src/loader.rs
vgao1996: Yeah this shouldn't be needed. You can pass a `&impl Trait` to `&dyn Trait`

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=59810b6cd41987904ba479800a708871

alinush: Regarding why the switch from `impl` to `dyn` for all those calls, it's because the `DataStore` in `NativeContext` is a  `&'a mut dyn DataStore` which cannot be passed as an `&impl DataStore` because its size is unknown (AFAIU).

I therefore changed all the `&impl`'s to `&dyn`'s.

I did not try changing the `&'a mut dyn DataStore` to a `&'a mut impl DataStore` though... maybe that would've worked.

@vgao1996, yes you can do that but I needed to do the opposite.

alinush:
> Generally, I'd say we are better off putting the `chain_id` in the `NativeTransactionContext` of Aptos. This will be lesser code changes.

I guess we should close this PR, then?

PR #603, opened 2022-10-19
added support to print UTF8-encoded std::string::String's via std::debug::print

alinush:
## Motivation

Currently, debugging in Move is very painful because nobody can print an `std::string::String` via `std::debug::print`.

This PR adds support for that by checking for an `std::string::String` type argument in the native Rust implementation of  `std::debug::print` and handling it appropriately.

Yes!

## Test Plan

Added a test case in `language/move-stdlib/nursery/sources/debug.move`:

```
#[test_only]
    use std::string;

    #[test]
    fun test_print_string() {
        let str_bytes = b"Hello, sane Move debugging!";

        print(&str_bytes);

        let str = string::utf8(str_bytes);
        print<string::String>(&str);
    }
```

Ran the tests via:

```
cd language/move-stdlib/nursery/sources/
cargo test -- --nocapture
```

(Part of) the output was:

```
[debug] (&) [72, 101, 108, 108, 111, 44, 32, 115, 97, 110, 101, 32, 77, 111, 118, 101, 32, 100, 101, 98, 117, 103, 103, 105, 110, 103, 33]
[debug] Hello, sane Move debugging!
```


language/move-stdlib/src/natives/debug.rs
```diff
@@ -25,34 +27,73 @@ pub struct PrintGasParameters {
...
+
+        match is_string_type(_context, &ty, move_std_addr) {
+            true => {
+                let r = pop_arg!(args, StructRef);
+
+                let field_ref = r
+                    .borrow_field(0)? // borrows the 'bytes' field of the std::string::String struct
+                    .value_as::<Reference>()?;
+
+                let bytes = field_ref.read_ref()?.value_as::<Vec<u8>>()?;
```
wrwg, line 71: I think you can make a `&[u8]` without need to make a copy, checkout `string.rs` which always gets away without copying.

wrwg (approved):
Good first step for the generic to_string ...

language/move-stdlib/src/natives/debug.rs
```diff
@@ -25,34 +27,73 @@ pub struct PrintGasParameters {
     pub base_cost: InternalGas,
 }
 
+#[allow(dead_code)]
```
davidiw, line 30: #[cfg(feature = "testing")]

davidiw (commented):
thanks!

language/move-stdlib/src/natives/debug.rs
alinush: Fixed! Thank you!

language/move-stdlib/src/natives/debug.rs
alinush: Ah, will take a look.

alinush: Fixed!

wrwg:
Great that this landed! Can you cherry pick into the `aptos` branch here? Its unfortunately a bit cumbersome right now.

alinush:
> Great that this landed! Can you cherry pick into the `aptos` branch here? Its unfortunately a bit cumbersome right now.

Sure, will try to!

PR #612, opened 2022-10-21
Modify GasMeter functions to take in stack information about where the function is being called

0xrelapse:
## Motivation

We've been using a modified version of the aptos node locally which spits out profiles of how much gas is used by each method during a transaction. Figured it would be useful for the rest of the community so wanted to publish it.

We first need to make the GasMeter trait accept more information about where the charge call is being made. This diff creates a `FuncCallStack` type (just a vector of function names) and passes that into the gas meter methods.

The idea is that the GasMeter implementation can use the passed in information to develop a profile of how gas is used.

Would love suggestions on other context that could be useful to pass to the gas meter. 

Yes

## Test Plan

Changes add a function parameter to each method of a trait. Made sure the changes compiled and existing tests passed.

### DIFF
--- a/language/move-vm/runtime/src/interpreter.rs
+++ b/language/move-vm/runtime/src/interpreter.rs
@@ -15,12 +15,13 @@ use move_binary_format::{
 use move_core_types::{
     account_address::AccountAddress,
     gas_algebra::{NumArgs, NumBytes},
+    identifier::Identifier,
     language_storage::TypeTag,
     vm_status::{StatusCode, StatusType},
 };
 use move_vm_types::{
     data_store::DataStore,
-    gas::{GasMeter, SimpleInstruction},
+    gas::{FuncCallStack, GasMeter, SimpleInstruction},
     loaded_data::runtime_types::Type,
     values::{
         self, GlobalValue, IntegerValue, Locals, Reference, Struct, StructRef, VMValueCast, Value,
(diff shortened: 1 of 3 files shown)


language/move-vm/runtime/src/interpreter.rs
```diff
@@ -124,13 +125,22 @@ impl Interpreter {
                 .map_err(|e| self.set_location(e))?;
         }
 
+        let func_call_stack = self
+            .call_stack
+            .to_func_call_stack(function.name())
+            .map_err(|e| self.set_location(e))?;
```
vgao1996, line 131: I think there are some issues with the implementation here. You only compute the call stack once at the entry point and don't recompute/modify it when a call is made, which means the stack will only contain the entry point ever.

vgao1996 (commented):
Hi @0xrelapse, thank you for the contribution! I fully agree that it is important to have the ability to generate itemized gas bills and it's the `GasMeter`'s responsibility to make that possible. We're also planning to build something similar here at Aptos.

However I'd like to suggest that we implement these special gas meters in a different way. With the current interfaces, the `GasMeter` already has all the information it needs to construct a call stack: 
1. When `charge_call` and `charge_call_generic` are called, you push the function name on to a stack
2. When `charge_ret` or `charge_call_native` are called, you pop a function off the stack

This has a few advantages:
1. We can avoid computing the call stack in the VM, unless a client really needs it. You can also have different clients implementing it in different ways.
2. We don't have to add a call stack parameter to every callback, which is quite intrusive. (I think the general rule of thumb here is that you only add a new parameter if the information wasn't already available)
3. This can be done in a modular way: you can come up with a "gas meter combinator" that stacks two gas meters, a regular one doing normal gas metering stuff and the other one constructing the call graph.

language/move-vm/runtime/src/interpreter.rs
0xrelapse: Yup you're right, this should be inside the loop - got mixed up while rebasing

0xrelapse:
@vgao1996 Yeah I agree, your approach makes a lot more sense. The concern of tracking gas per item should be entirely delegated to the GasMater itself.

Will abandon this diff and create a new one specific to Aptos with that implementation!

PR #616, opened 2022-10-22
Show Gas Used In Unit Tests & Allow Client Supplied Gas Cost Table

oxade:
## Motivation
Knowing how much gas is used is very useful in unit tests.
We currently show instruction executed when the `--statistics` flag is supplied.
However instruction count does not provide much use in real systems, so replacing this with gas used, where the client can supply the cost table for gas charging.

This approach allows the client to supply their gas table to the test suite, which then uses it for metering as it would during normal operation.
Note that we will have to increase the instruction bound as appropriate.

Note if we want the instruction count, we simply don't provide a gas table, which defaults each instr to 1 gas as before. This counts instructions.

Example outputs below use non-unity weighting in gas table (from Sui example).
```
BUILDING Basics
Running Move unit tests
[ PASS    ] 0x0::counter_test::test_counter
[ PASS    ] 0x0::lockTest::test_lock
[ PASS    ] 0x0::test_sandwich::test_make_sandwich

Test Statistics:

┌────────────────────────────────────────┬────────────┬───────────────────────────┐
│               Test Name                │    Time    │          Gas Used         │
├────────────────────────────────────────┼────────────┼───────────────────────────┤
│ 0x0::counter_test::test_counter        │   0.010    │            493            │
├────────────────────────────────────────┼────────────┼───────────────────────────┤
│ 0x0::lockTest::test_lock               │   0.010    │            501            │
├────────────────────────────────────────┼────────────┼───────────────────────────┤
│ 0x0::test_sandwich::test_make_sandwich │   0.011    │            785            │

```

(Write your answer here.)
Yes

## Test Plan

N/A


vgao1996:
I love the rendering, but I'd like to push back a little bit: the unit testing framework is still using the legacy GasStatus and a hard-coded dummy gas schedule, so the actual gas numbers here don't really make sense or match the actual ones downstream clients use. Displaying these numbers can confuse users a lot. 

I think correct long term solution is introduce a way for the client to plug in their own gas meter implementations somehow. I was aware of this problem when I added the `GasMeter` trait, but didn't have the spare bandwidth to deal with it that time.

oxade:
> I love the rendering, but I'd like to push back a little bit: the unit testing framework is still using the legacy GasStatus and a hard-coded dummy gas schedule, so the actual gas numbers here don't really make sense or match the actual ones downstream clients use. Displaying these numbers can confuse users a lot.
> 
> I think correct long term solution is introduce a way for the client to plug in their own gas meter implementations somehow. I was aware of this problem when I added the `GasMeter` trait, but didn't have the spare bandwidth to deal with it that time.

@vgao1996 this does allow you plug in your own gas meter.
The outputs here were from Sui gas meter. We do not use the legacy gas anywhere.
If the client doesn't plug theirs in, then it defaults to 1 for each instr, which was the previous behavior.

vgao1996 (approved):
> this does allow you plug in your own gas meter

Not really. That still requires one to use the legacy `CostTable`. We have a totally custom gas meter implementation that directly interfaces with the `GasMeter` trait and uses custom structure to store its gas parameters. The unit test framework needs to be refactored so that it is generic over the gas meter, just like the VM.

With that being said, I now realize that this is more of a legacy problem and we don't have to redesign it in this PR. (I knew it when I added the `GasMeter` trait, but didn't have the bandwidth to deal with this at that time) I also checked our own use cases and we currently do not enable statistics so this change shouldn't affect us.

vgao1996:
Oh, I should also mention that even though the previous implementation claims it's counting the number of instructions, it's not really doing it correctly. By setting all gas parameters to `1`, it charges `1 + 1 * input_length_or_size` for certain instructions

oxade:
> > this does allow you plug in your own gas meter
> 
> Not really. That still requires one to use the legacy `CostTable`. We have a totally custom gas meter implementation that directly interfaces with the `GasMeter` trait and uses custom structure to store its gas parameters. The unit test framework needs to be refactored so that it is generic over the gas meter, just like the VM.
> 
> With that being said, I now realize that this is more of a legacy problem and we don't have to redesign it in this PR. (I knew it when I added the `GasMeter` trait, but didn't have the bandwidth to deal with this at that time) I also checked our own use cases and we currently do not enable statistics so this change shouldn't affect us.

True. I misread CostTable for GasMeter. Yeah we need a heavy refactor. We can scope this for a follow up if you think its high pri

oxade:
> Oh, I should also mention that even though the previous implementation claims it's counting the number of instructions, it's not really doing it correctly. By setting all gas parameters to `1`, it charges `1 + 1 * input_length_or_size` for certain instructions

Haha thats def not good. We should def scrap it.

vgao1996:
> We can scope this for a follow up if you think its high pri

Probably not that high pri right now, but I'd imagine us running into this again when we start to polish some of the user-facing tools.

PR #622, opened 2022-10-25
[build] switch to lld on linux and optimize ci build

vgao1996:
`ld`, the default linker on linux, consumes a lot of memory and does not run in parallel. This PR switches us over to `lld`, which should greatly reduce memory consumption and improve linking speed on machines with many cores. Here's a comparison on a 64-core VM:
- `ld`: 17GB Ram, 48.21s
- `lld`: 4.5GB Ram, 34.40s

The reduction in memory consumption also allows us to switch back to debug build in CI, **cutting the running time from 33m down to 15m (with cache)**.

Note: this does require `lld` to be installed on linux platforms, which is now included in the dev setup script. However existing developers may need to re-run the script or install `lld` manually.

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -11,3 +11,6 @@ nextest = "run --package x --bin x -- nextest"
 
 [target.x86_64-pc-windows-msvc]
 rustflags = ["-C", "link-arg=/STACK:8000000"]
+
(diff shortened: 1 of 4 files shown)


wrwg (approved):
Great to see this also in this repo!

PR #625, opened 2022-10-26
[Prover]  Extends the intrinsic table implementation by adding more pragmas

junkil-park:
- Added two pragmas for the intrinsic table implementation
  - map_del_return_key: a delete function that returns not only the value but also the key. `simple_map` needs this because keys are not drop-able there.
  - map_add_override_if_exists: an add function that override the value rather than abort when the key already exists. This models the `upsert` functions in the table modules

- Modified the meaning of `map_spec_set` so that the length only increases when a new entry is added.

## Motivation

To add necessary pragmas for table

yes

## Test Plan

cargo test

### DIFF
--- a/language/extensions/move-table-extension/sources/Table.spec.move
+++ b/language/extensions/move-table-extension/sources/Table.spec.move
@@ -15,7 +15,7 @@ spec extensions::table {
             map_borrow = borrow,
             map_borrow_mut = borrow_mut,
             map_spec_get = spec_get,
-            map_spec_set = spec_add,
+            map_spec_set = spec_set,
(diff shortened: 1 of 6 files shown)


meng-xu-cs:
Only a minor suggestion:

I would prefer `fun_add_override` to be named as `fun_add_override_if_exists`, just to be more explicit on the behavior.

junkil-park:
> Only a minor suggestion:
> 
> I would prefer `fun_add_override` to be named as `fun_add_override_if_exists`, just to be more explicit on the behavior.

Done. Thanks for the suggestion!

meng-xu-cs (approved):

PR #629, opened 2022-10-27
preserve previous enum order

oxade:
## Motivation

The PR preserves the order of existing enums which was altered by the new integer additions.
Changing the order affected encodings and breaks backward compatibility.

Yes
## Test Plan

Standard unit tests


tnowacki (commented):
It's a good catch! But I think should only affect TypeTag, FatType, and TransactionArgument. I don't think the other spots should be affected?

oxade:
> It's a good catch! But I think should only affect TypeTag, FatType, and TransactionArgument. I don't think the other spots should be affected?

I agree but I kinda didn't wanna take any chances.

language/move-binary-format/src/file_format.rs
```diff
@@ -864,6 +858,12 @@ pub enum SignatureToken {
     MutableReference(Box<SignatureToken>),
     /// Type parameter.
     TypeParameter(TypeParameterIndex),
+    /// Unsigned integers, 16 bits length.
```
sblackshear, line 861: For these out-of-order variants, I think it's useful to have a comment like `// added in bytecode version v6, do not reorder!` so future folks understand why it works this way.

tnowacki:
> I generally agree but I kinda didn't wanna take any chances lol.

I think it is fine for anything that doesn't implement `Serialize` or `Deserialize`.

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -142,6 +144,7 @@ pub enum IntegerValue {
     U8(u8),
     U64(u64),
     U128(u128),
+    // NOTE: Added in bytecode version v6, do not reorder!
```
tnowacki, line 147: I think this is more confusing than helpful. this is not needed here

language/move-core/types/Cargo.toml
```diff
@@ -32,6 +32,7 @@ proptest = "1.0.0"
 proptest-derive = "0.3.0"
 regex = "1.5.5"
 serde_json = "1.0.64"
+arbitrary = "1.1.7"
```
tnowacki, line 35: why the change?

language/move-compiler/src/hlir/ast.rs
```diff
@@ -278,6 +278,8 @@ pub enum Value_ {
     Bool(bool),
     // vector<type> [ <value>,* ]
     Vector(Box<BaseType>, Vec<Value>),
+
+    // NOTE: Added in bytecode version v6, do not reorder!
```
tnowacki, line 282: and here

language/move-compiler/src/expansion/ast.rs
```diff
@@ -385,6 +385,8 @@ pub enum Value_ {
     // false
     Bool(bool),
     Bytearray(Vec<u8>),
+
+    // NOTE: Added in bytecode version v6, do not reorder!
```
tnowacki, line 389: and here

language/move-command-line-common/src/values.rs
```diff
@@ -55,6 +55,7 @@ pub enum ParsedValue<Extra: ParsableValue = ()> {
         BTreeMap<String, ParsedValue<Extra>>,
     ),
     Custom(Extra),
+    // NOTE: Added in bytecode version v6, do not reorder!
```
tnowacki, line 58: and here

language/move-command-line-common/src/types.rs
```diff
@@ -41,6 +41,7 @@ pub enum ParsedType {
     Signer,
     Vector(Box<ParsedType>),
     Struct(ParsedStructType),
+    // NOTE: Added in bytecode version v6, do not reorder!
```
tnowacki, line 44: and here

language/move-binary-format/src/file_format.rs
```diff
@@ -858,6 +858,8 @@ pub enum SignatureToken {
     MutableReference(Box<SignatureToken>),
     /// Type parameter.
     TypeParameter(TypeParameterIndex),
+
+    // NOTE: Added in bytecode version v6, do not reorder!
```
tnowacki, line 862: And here

tnowacki (changes requested):
I think these comments are more confusing than helpful. The order for many of these enums does not matter. And implying it does might lead to confusion (but maybe I'm missing something)

oxade:
@tnowacki 
Yes I agree most are not useful. I wanted to be defensive and flag as many as I can and then we can exclude what we don't want; rather than the other way around.

I will review the ones you flagged

language/move-core/types/Cargo.toml
oxade: I was getting weird failures on local test where the package was not found. This seemed to fix it. I pushed it to see if CI will complain

tnowacki:
> @tnowacki Yes I agree most are not useful. I wanted to be defensive and flag as many as I can and then we can mass exclude what we decide we don't want; rather than the other way around.
> 
> I will review the ones you flagged

I understand your trepidation here and making sure noting accidentally goes wrong, but if the enum doesn't have Serialize/Deserialize, the change likely isn't observable. Though we can look case-by-case for ones with Hash.
Readability and understandability are important too. As one might expect these sorts of similar operations to be grouped together. I just worry that this ruins the readability of things like the `Bytecode` or `SignatureToken` or various `Type` enums in the compiler, in the name of being cautious.

oxade:
> > @tnowacki Yes I agree most are not useful. I wanted to be defensive and flag as many as I can and then we can mass exclude what we decide we don't want; rather than the other way around.
> > I will review the ones you flagged
> 
> I understand your trepidation here and making sure noting accidentally goes wrong, but if the enum doesn't have Serialize/Deserialize, the change likely isn't observable. Though we can look case-by-case for ones with Hash. Readability and understandability are important too. As one might expect these sorts of similar operations to be grouped together. I just worry that this ruins the readability of things like the `Bytecode` or `SignatureToken` or various `Type` enums in the compiler, in the name of being cautious.

I have reverted to the following:
* Sorted order for those which do not implement Hash or Ser/Des
* Append order for those which implement Hash or Ser/Des

Ready for review now

tnowacki (approved):
Probably isn't needed for `SignatureToken` and `Bytecode` but not worth blocking over. Thanks!

PR #630, opened 2022-10-27
[native functions] adds pretty-printing to `std::debug::print`

alinush:
This **draft** PR implements pretty-printing of Move values via `std::debug::print`, building upon #603 which added support for printing `std::string::String`.

An example of the new `std::debug::print` output:

```
[debug] 0x1::debug::TestStruct {
  addr: @0x1,
  number: 255,
  bytes: 0xc0ffee,
  name: "He\"llo",
  vec: [
    0x1::debug::TestInner {
      val: 1,
      vec: [ 130, 131 ]
    },
    0x1::debug::TestInner {
      val: 2,
      vec: [ 132, 133 ]
    }
  ]
}
```

Note that we print `vector<u8>`'s as hexadecimal strings, since this is likely to be a more useful format for developers to inspect.

Also note that when a `vector<T>` has a complex type, such as `T = vector<S>` or a `T = struct`, we display the vector entries on separate lines, instead of the same line.

**TODOs:**

 - <strike>fix `cargo build` warnings due to missing "testing" feature flag</strike>
 - <strike>print `vector<u8>` as hex-encoded strings</strike>
 - <strike>optionally-include u8/16/.../256 types</strike>
 - <strike>update & fix tests in `language/tools/move-cli/tests/sandbox_tests` via `cargo test -- print_`</strike>
 - <strike>update pretty-print example above</strike>

## Motivation

It is very difficult for Move developers to debug their code without the ability to print a `struct`'s fields names and values.

## If you are reviewing this PR

 - Tell me if I am returning the wrong `StatusCode` in the `PartialVMError` objects
 - Tell me if I am escaping strings correctly via `let str = str.replace('\\', "\\\\").replace('"', "\\\"");`

## Approach

Our approach is to first convert the `Value` passed to `std::debug::print` to a `MoveValue`, carefully handling the `vector` and `struct` cases (Specifically, structs need to be annotated with field names and types while vectors need to be unpacked into their inner elements recursively). 

Then, we reimplement the `std::fmt::Display` logic for `MoveValue`'s to pretty-print them. 

This reimplementation effort is necessary for us to:

 - manually inspect if a struct is an `std::string::String` and pretty-print it
   + Might want to do something similar for `std::option::Option`
 - manually select whether we display canonical addresses (useful for signatures) or compressed addresses (useful for pretty-printing)
 - add indentation based on current depth inside a struct

### Alternative approach: Avoid exposing `Value::as_move_value()` as public

**Reason 1:** Without a public `Value::as_move_value()` we would have to serialize & deserialize a struct `Value` into a `MoveValue::MoveStruct`, which is less efficient than calling `Value::as_move_value()`. Although performance is not a constraint for pretty-printing, it **is** a constraint for verifying signatures on canonical serialization of structs.

**Reason 2:** Furthermore, converting every `Value` to a `MoveValue` allows us to implement (almost) all printing logic inside a single `print_move_value` function that operates on `MoveValue`'s.

First, recall that we "decorate" struct `Value`'s  with their fields' type information and obtain a `MoveValue::MoveStruct`.  So if we want to quickly print such a `MoveValue::MoveStruct`, we need to implement recursively-printing any `MoveValue`, which we do in `print_move_value`.

Our next goal then becomes not to reimplement any more printing logic, which is why it is convenient to turn any `Value` into a `MoveValue` and pass it in to `print_move_value`.

Yes.

## Test Plan

 - Test pretty-printing of basic types (see all `MoveValue` types)
 - Test pretty-printing of a simple struct
 - Test pretty-printing of nested structs
 - Test pretty-printing of generic structs
 - Test pretty-printing of vectors, and vectors of vectors

### DIFF
--- a/language/move-stdlib/src/natives/debug.rs
+++ b/language/move-stdlib/src/natives/debug.rs
@@ -165,8 +165,8 @@ mod testing {
     const STRUCT_END: &str = "}";
 
     fn fmt_error_to_partial_vm_error(e: fmt::Error) -> PartialVMError {
-        let vmerr = PartialVMError::new(StatusCode::UNKNOWN_STATUS);
-        vmerr.with_message("write! macro failed with: ".to_string() + e.to_string().as_str())
+        PartialVMError::new(StatusCode::UNKNOWN_STATUS)
+            .with_message("write! macro failed with: ".to_string() + e.to_string().as_str())
     }
 
     fn to_vec_u8_type_err<E>(_e: E) -> PartialVMError {
@@ -198,12 +198,14 @@ mod testing {
         }
     }
 
-    /// Prints an std::string::String by wrapping it in double quotes and escaping double quotes inside it.
+    /// Converts a `MoveValue::Vector` of `u8`'s to a `String` by wrapping it in double quotes and
+    /// escaping double quotes and backslashes.
     ///
     /// Examples:
-    ///  - 'Hello' prints as "Hello"
-    ///  - '"Hello?" What are you saying?' prints as "\"Hello?\" What are you saying?"
-    fn print_move_value_as_string(out: &mut String, val: MoveValue) -> PartialVMResult<()> {
+    ///  - 'Hello' returns "Hello"
+    ///  - '"Hello?" What are you saying?' returns "\"Hello?\" What are you saying?"
+    ///  - '\ and " are escaped' returns "\\ and \" are escaped"
+    fn move_value_as_escaped_string(val: MoveValue) -> PartialVMResult<String> {
         match val {
             MoveValue::Vector(bytes) => {
                 let buf = MoveValue::vec_to_vec_u8(bytes).map_err(to_vec_u8_type_err)?;
@@ -216,20 +218,11 @@ mod testing {
 
                 // We need to escape displayed double quotes " as \" and, as a result, also escape
                 // displayed \ as \\.
-                let str = str.replace('\\', "\\\\").replace('"', "\\\"");
-
-                write!(out, "\"{}\"", str).map_err(fmt_error_to_partial_vm_error)?;
+                Ok(str.replace('\\', "\\\\").replace('"', "\\\""))
             }
-            _ => {
-                return Err(
-                    PartialVMError::new(StatusCode::INTERNAL_TYPE_ERROR).with_message(
-                        "Expected std::string::String to have vector<u8> bytes field".to_string(),
-                    ),
-                )
-            }
+            _ => Err(PartialVMError::new(StatusCode::INTERNAL_TYPE_ERROR)
+                .with_message("Expected a MoveValue::Vector of u8's".to_string())),
         }
-
-        Ok(())
     }
 
     fn print_padding_at_depth(out: &mut String, depth: usize) -> PartialVMResult<()> {
@@ -253,6 +246,7 @@ mod testing {
         matches!(mv, MoveValue::Vector(_) | MoveValue::Struct(_))
     }
 
+    /// Prints any `Value` in a user-friendly manner.
     pub(crate) fn print_value(
         context: &NativeContext,
         out: &mut String,
@@ -274,11 +268,12 @@ mod testing {
                 let inner_tyl = context.type_to_type_layout(inner_ty)?.unwrap();
 
                 match inner_tyl {
-                    // We cannot simply convert this vector `Value` to a `MoveValue` because there might
-                    // be a struct in the vector that needs to be "decorated" using the logic in this
-                    // function. Instead, we recursively "unpack" the vector until we get down to either
-                    // a primitive type or a struct, which this function can then forward to
-                    // `print_move_value`.
+                    // We cannot simply convert a `Value` (of type vector) to a `MoveValue` because
+                    // there might be a struct in the vector that needs to be "decorated" using the
+                    // logic in this function. Instead, we recursively "unpack" the vector until we
+                    // get down to either (1) a primitive type, which we can forward to
+                    // `print_move_value`, or (2) a struct type, which we can decorate and forward
+                    // to `print_move_value`.
                     MoveTypeLayout::Vector(_) | MoveTypeLayout::Struct(_) => {
                         // `val` is either a `Vec<Vec<Value>>`, a `Vec<Struct>`,  or a `Vec<signer>`, so we cast `val` as a `Vec<Value>` and call ourselves recursively
                         let vec = val.value_as::<Vec<Value>>()?;
@@ -372,7 +367,8 @@ mod testing {
         Ok(())
     }
 
-    /// Prints the MoveValue, optionally-including its type if `print_type` is true.
+    /// Prints the MoveValue in `mv`, optionally-printing integer types if `include_int_type` is
+    /// true.
     fn print_move_value(
         out: &mut String,
         mv: MoveValue,
@@ -420,7 +416,9 @@ mod testing {
                 }
             }
             MoveValue::Bool(b) => {
-                // The type here can be inferred just from the true/false value, when `include_int_types` is enabled.
+                // Note that when `include_int_types` is enabled, the boolean `true` and `false`
+                // values unambiguously encode their type, since they are different than any integer
+                // type value, address value, signer value, vector value and struct value.
                 write!(out, "{}", if b { "true" } else { "false" })
                     .map_err(fmt_error_to_partial_vm_error)?;
             }
@@ -468,12 +466,25 @@ mod testing {
 
                     // Check if struct is an std::string::String
                     if !canonicalize && type_.is_std_string(move_std_addr) {
-                        assert_eq!(fields.len(), 1);
+                        if fields.len() != 1 {
+                            return Err(PartialVMError::new(StatusCode::INTERNAL_TYPE_ERROR)
+                                .with_message(
+                                    "Expected std::string::String struct to have just one field"
+                                        .to_string(),
+                                ));
+                        }
 
                         let (id, val) = fields.pop().unwrap();
-                        assert_eq!(id.into_string(), "bytes");
+                        if id.into_string() != "bytes" {
+                            return Err(PartialVMError::new(StatusCode::INTERNAL_TYPE_ERROR)
+                                .with_message(
+                                    "Expected std::string::String struct to have a `bytes` field"
+                                        .to_string(),
+                                ));
+                        }
 
-                        print_move_value_as_string(out, val)?;
+                        let str = move_value_as_escaped_string(val)?;
+                        write!(out, "\"{}\"", str).map_err(fmt_error_to_partial_vm_error)?
                     } else {
                         write!(out, "{} ", type_tag).map_err(fmt_error_to_partial_vm_error)?;
                         write!(out, "{}", STRUCT_BEGIN).map_err(fmt_error_to_partial_vm_error)?;
--- a/language/move-stdlib/src/natives/debug.rs
+++ b/language/move-stdlib/src/natives/debug.rs
@@ -3,20 +3,537 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::natives::helpers::make_module_natives;
+#[cfg(feature = "testing")]
+use move_binary_format::errors::PartialVMError;
 use move_binary_format::errors::PartialVMResult;
+use move_core_types::{account_address::AccountAddress, gas_algebra::InternalGas};
+#[cfg(feature = "testing")]
 use move_core_types::{
-    account_address::AccountAddress, gas_algebra::InternalGas, language_storage::TypeTag,
+    language_storage::{StructTag, TypeTag},
+    value::{MoveStruct, MoveStructLayout, MoveTypeLayout, MoveValue},
+    vm_status::StatusCode,
 };
 use move_vm_runtime::native_functions::{NativeContext, NativeFunction};
-use move_vm_types::values::VectorRef;
-#[allow(unused_imports)]
-use move_vm_types::values::{values_impl::debug::print_reference, Reference, StructRef};
+#[cfg(feature = "testing")]
+use move_vm_types::values::Reference;
 #[allow(unused_imports)]
 use move_vm_types::{
     loaded_data::runtime_types::Type, natives::function::NativeResult, pop_arg, values::Value,
 };
 use smallvec::smallvec;
 use std::{collections::VecDeque, sync::Arc};
+#[cfg(feature = "testing")]
+use std::{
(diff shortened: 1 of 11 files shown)


language/move-stdlib/src/natives/debug.rs
```diff
@@ -60,29 +61,59 @@ fn native_print(
...
+        
+        // if ty is a struct type:
+        // get type layout in VM format (raw BCS, no field names) 
+        let old_type_layout = context.type_to_type_layout(&ty)?.unwrap();
+
+        println!("Type: {:?}", &ty);
+        println!("Value: {:?}", &args[0]);
+        println!("MoveTypeLayout: {:?}", old_type_layout);
+
+        // For some reason this line fails when calling `debug::print` on the `TestStruct` in debug.move
```
sblackshear, line 73: I got this to work on your tests with the following tweaks:

```
  let ty = ty_args.pop().unwrap();
        // get type layout in VM format (raw BCS, no field names)
        let old_layout = context.type_to_type_layout(&ty)?.unwrap();
        match &old_layout {
            MoveTypeLayout::Struct(old_struct_layout) => {
                let arg = pop_arg!(args, StructRef);
                let arg_val = arg.read_ref()?;
                // get BCS bytes for the struct
                let blob = arg_val.simple_serialize(&old_layout).unwrap();
                let move_struct =
                    MoveStruct::simple_deserialize(&blob, &old_struct_layout).unwrap();

                // get type layout in decorated format (with field names)
                let new_type_layout = context.type_to_fully_annotated_layout(&ty)?.unwrap();
                let new_struct_layout = match &new_type_layout {
                    MoveTypeLayout::Struct(struct_layout) => struct_layout,
                    _ => panic!("ooppsie; this shouldn't happen."),
                };

                let decorated_struct = move_struct.decorate(new_struct_layout);
                println!("[debug] {}", decorated_struct);
```

sblackshear: I then get:

```
INCLUDING DEPENDENCY MoveStdlib
BUILDING MoveNursery
Running Move unit tests
[ PASS    ] 0x1::errors_tests::errors_state
[ PASS    ] 0x1::guid_tests::test_basics
[ PASS    ] 0x1::roleTests::test_assign_failure
[ PASS    ] 0x1::aclTests::test_add_failure
[ PASS    ] 0x1::roleTests::test_revoke_failure
[ PASS    ] 0x1::guid_tests::test_delegation
[ PASS    ] 0x1::capability_tests::test_delegate_failure_after_revoke
[debug] 0x1::debug::TestStruct {number: 255u8, bytes: vector[192u8, 255u8, 238u8], name: 0x1::string::String {bytes: vector[72u8, 101u8, 108u8, 108u8, 111u8, 44u8, 32u8, 115u8, 97u8, 110u8, 101u8, 32u8, 77u8, 111u8, 118u8, 101u8, 32u8, 100u8, 101u8, 98u8, 117u8, 103u8, 103u8, 105u8, 110u8, 103u8, 33u8]}}
[ PASS    ] 0x1::debug::test_print_annotated_struct
[ PASS    ] 0x1::roleTests::test_success
[ PASS    ] 0x1::guid_tests::test_id
[ PASS    ] 0x1::compareTests::equality_of_simple_types
[ PASS    ] 0x1::aclTests::test_remove_failure
[ PASS    ] 0x1::event_tests::test_event_128
[ PASS    ] 0x1::capability_tests::test_delegate_success
[debug] (&) [72, 101, 108, 108, 111, 44, 32, 115, 97, 110, 101, 32, 77, 111, 118, 101, 32, 100, 101, 98, 117, 103, 103, 105, 110, 103, 33]
[debug] 0x1::string::String {bytes: vector[72u8, 101u8, 108u8, 108u8, 111u8, 44u8, 32u8, 115u8, 97u8, 110u8, 101u8, 32u8, 77u8, 111u8, 118u8, 101u8, 32u8, 100u8, 101u8, 98u8, 117u8, 103u8, 103u8, 105u8, 110u8, 103u8, 33u8]}
```

which is ugly, but does have field names!

sblackshear: I think `struct DebugStruct(MoveStruct)` with a custom `Display` impl should do the trick for the rest.

alinush: Thank you!

The following struct:
```
        let obj = TestStruct {
            number: 255u8,
            bytes: x"c0ffee",
            name: string::utf8(b"Hello, sane Move debugging!"),
        };
```

...now prints as:
```
[annotated-debug] 0x1::debug::TestStruct {number: 255u8, bytes: vector[192u8, 255u8, 238u8], name: 0x1::string::String {bytes: vector[72u8, 101u8, 108u8, 108u8, 111u8, 44u8, 32u8, 115u8, 97u8, 110u8, 101u8, 32u8, 77u8, 111u8, 118u8, 101u8, 32u8, 100u8, 101u8, 98u8, 117u8, 103u8, 103u8, 105u8, 110u8, 103u8, 33u8]}}
```

This seems to be enough for pretty-print and, as a result, could be enough for domain-separated signatures.

I could change this into a PR that adds a new `pretty_print` function in `std::debug` if you all think it's a good idea @sblackshear, @wrwg, @vgao1996 et al.

(This would also be a good way for me to familiarize myself with these "decorated" `MoveStruct`'s.)

language/move-stdlib/src/natives/debug.rs
```diff
@@ -60,29 +61,61 @@ fn native_print(
...
+                let new_struct_layout = match &new_type_layout {
+                    MoveTypeLayout::Struct(struct_layout) => struct_layout,
+                    _ => panic!("ooppsie; this shouldn't happen.")
+                };
+
+                let decorated_struct = move_struct.decorate(new_struct_layout);
+                println!("[annotated-debug] {}", decorated_struct);
+            },
+            _ => {
+                match is_string_type(context, &ty, move_std_addr) {
```
tnowacki, line 94: First time I've been looking through this code, so sorry if its a duplicate question, but, shouldn't we just add some custom printing/display logic for this? If we don't, it means we will print strings as strings when they are top level values, but not deeply inside of structs

language/move-stdlib/src/natives/debug.rs
```diff
@@ -60,29 +61,61 @@ fn native_print(
...
+                    true => {
```
tnowacki, line 95: nit, while I do love `match`, match on a `bool` feels a bit silly as it is just an `if` `else` :)

sblackshear:
Awesome work @alinush !!!

> I could change this into a PR that adds a new pretty_print function in std::debug if you all think it's a good idea @sblackshear, @wrwg, @vgao1996 et al.

This sounds reasonable to me. But I would be just as ok with (and would in fact prefer) just changing the behavior of the old `std::debug` rather than introducing a new function. The debug code can't be used unless the `testing` feature flag is on, the effects aren't observable inside Move code, and I can't imagine a case where someone would prefer the ugly old printing to the beautiful new printing.

language/move-stdlib/src/natives/debug.rs
sblackshear: Yep, agreed with this. We will also not print fields of nested structs unless we go in this direction.

tnowacki:
> Awesome work @alinush !!!
> 
> > I could change this into a PR that adds a new pretty_print function in std::debug if you all think it's a good idea @sblackshear, @wrwg, @vgao1996 et al.
> 
> This sounds reasonable to me. But I would be just as ok with (and would in fact prefer) just changing the behavior of the old `std::debug` rather than introducing a new function. The debug code can't be used unless the `testing` feature flag is on, the effects aren't observable inside Move code, and I can't imagine a case where someone would prefer the ugly old printing to the beautiful new printing.

+1 for this just replacing the normal debug print

alinush:
> > Awesome work @alinush !!!
> > > I could change this into a PR that adds a new pretty_print function in std::debug if you all think it's a good idea @sblackshear, @wrwg, @vgao1996 et al.
> > 
> > 
> > This sounds reasonable to me. But I would be just as ok with (and would in fact prefer) just changing the behavior of the old `std::debug` rather than introducing a new function. The debug code can't be used unless the `testing` feature flag is on, the effects aren't observable inside Move code, and I can't imagine a case where someone would prefer the ugly old printing to the beautiful new printing.
> 
> +1 for this just replacing the normal debug print

Beautiful, I was also thinking that changing `std::debug::print` to pretty-print is best too: who would ever use an ugly-print :P?

language/move-stdlib/src/natives/debug.rs
alinush: For sure! I should've been more clear that this is a draft PR. So all of it will change!

language/move-stdlib/src/natives/debug.rs
alinush: Indeed! I should've been more clear that this is a draft PR. So all of it will change!

vgao1996 (approved):

language/move-stdlib/src/natives/debug.rs
```diff
@@ -3,20 +3,537 @@
...
+    struct_tag.address == *move_std_addr
+        && struct_tag.module.as_str().eq("string")
+        && struct_tag.name.as_str().eq("String")
+}
+
+#[cfg(feature = "testing")]
+const HANDLE_FMT_WRITE_ERROR: fn(Error) -> PartialVMError = |e: fmt::Error| {
+    let vmerr = PartialVMError::new(StatusCode::UNKNOWN_STATUS);
+    vmerr.with_message("write! macro failed with: ".to_string() + e.to_string().as_str())
+};
```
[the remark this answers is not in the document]
sblackshear: but closures are so :sunglasses:

language/move-stdlib/src/natives/debug.rs
```diff
@@ -3,20 +3,537 @@
...
 use smallvec::smallvec;
 use std::{collections::VecDeque, sync::Arc};
+#[cfg(feature = "testing")]
+use std::{
+    fmt,
+    fmt::{Error, Write},
+};
+
+#[cfg(feature = "testing")]
+fn is_string_struct_tag(struct_tag: &StructTag, move_std_addr: &AccountAddress) -> bool {
```
sblackshear, line 32: I think this might want be useful as a function in `impl StructTag` [here](https://github.com/move-language/move/blob/main/language/move-core/types/src/language_storage.rs), since there is a decent amount of code that wants to special-case strings.

language/move-stdlib/src/natives/debug.rs
```diff
@@ -3,20 +3,537 @@
...
+fn get_vector_inner_type(ty: &Type) -> PartialVMResult<&Type> {
+    match ty {
+        Type::Vector(ty) => Ok(ty),
+        _ => Err(PartialVMError::new(StatusCode::INTERNAL_TYPE_ERROR)
+            .with_message("Could not get the inner Type of a vector's Type".to_string())),
+    }
+}
+
+#[cfg(feature = "testing")]
+fn vector_move_value_to_vec_u8(bytes: Vec<MoveValue>) -> PartialVMResult<Vec<u8>> {
```
sblackshear, line 70: Here too, might be useful in `impl MoveValue` (w/ or w/o the testing gate)

language/move-stdlib/src/natives/debug.rs
```diff
@@ -3,20 +3,537 @@
...
+        write!(out, "  ").map_err(HANDLE_FMT_WRITE_ERROR)?;
+    }
+
+    Ok(())
+}
+
+#[cfg(feature = "testing")]
+fn is_vector_u8(vec: &Vec<MoveValue>) -> bool {
+    if vec.is_empty() {
+        false
```
sblackshear, line 137: An empty vector can't be a `vector<u8>`? Probably just a naming issue, but potentially confusing.

language/move-stdlib/src/natives/debug.rs
```diff
@@ -28,62 +545,41 @@ pub struct PrintGasParameters {
...
-
-#[allow(unused_mut)]
 #[inline]
 fn native_print(
     gas_params: &PrintGasParameters,
     _context: &mut NativeContext,
     mut ty_args: Vec<Type>,
     mut args: VecDeque<Value>,
-    move_std_addr: AccountAddress,
+    _move_std_addr: AccountAddress,
```
sblackshear, line 554:
```suggestion
    move_std_addr: AccountAddress,
```

Nit: the Rust compiler will avoid emitting unused variable warnings if the name has a `_` prefix, so it is safest to only give a variable name this prefix if you are not using it.

language/move-stdlib/src/natives/debug.rs
alinush: Will move!

language/move-stdlib/src/natives/debug.rs
alinush: Will move.

language/move-stdlib/src/natives/debug.rs
alinush: Hm, should rename it indeed. The problem is I can't tell if it's a `vector<u8>` because there's no `MoveValue` I can pop to `match` for a `MoveValue::U8` on... Fortunately, I don't need to be able to identify empty `vector<u8>`'s so I just return false.

But this function deserves a clearer name like `is_non_emtpy_vector<u8>`.

language/move-stdlib/src/natives/debug.rs
alinush: Yes, indeed! But the problem with leaving it as `move_std_addr` is it emits the warning when the `testing` feature is off... because nobody uses it in that case.

language/move-stdlib/src/natives/debug.rs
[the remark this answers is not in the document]
alinush: I remember what happened here: I initially declared this inside a function and just cut & pasted it out when I realized I need it somewhere else 🤦 😆

language/move-stdlib/src/natives/debug.rs
```diff
@@ -3,20 +3,537 @@
...
 };
 use smallvec::smallvec;
 use std::{collections::VecDeque, sync::Arc};
+#[cfg(feature = "testing")]
+use std::{
+    fmt,
+    fmt::{Error, Write},
+};
+
+#[cfg(feature = "testing")]
```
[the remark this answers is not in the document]
alinush: I've wrapped everything in inside a `mod testing {}` in `debug.rs`, protected with a single `#[cfg(feature = "testing")]` line.

alinush:
I've addressed all concerns and marked them as "resolved." I will merge this tomorrow unless there are other suggestions!

PR #633, opened 2022-10-28
[unit-tests] Fix potential issue in unit tests

tnowacki:
- Unit tests did not check abort code locations, this made it difficult to test if you were checking for the right abort
- Added locations with a warning if missing, to not completely break old tests
- Added the ability to use constants for abort codes, e.g. `abort_code=0x1::m::EMyError` instead of `abort_code=0` 
- Added the ability to catch other errors
 - Arithmetic errors, vector errors, out of gas errors, or explicit major/minor codes 
 
## Followup 

Attributes aren't widely used at the moment, so it is probably fine for now, but with this change the grammar is a bit ambiguous. `location` is the first attribute that wants a module identifier instead of an expression. But checking for this means it is ambiguous if we want to resolve `a::m` as `a` being a module alias or an address alias. Similarly for `n`, do we check this as a module alias or module member alias?

I will try to clean this up in a future PR

## Test Plan

- new tests


emmazzz:
@wrwg @junkil-park This PR is failing in CI due to a timeout in prover non-linear arithmetics tests. Any idea what's causing the timeout?

language/documentation/examples/diem-framework/move-packages/DPN/tests/AccountFreezingTests.move
```diff
@@ -5,7 +5,7 @@ module DiemFramework::AccountFreezingTests {
     use std::signer;
 
     #[test(tc = @TreasuryCompliance, dr = @DiemRoot)]
-    #[expected_failure(abort_code = 1)]
+    #[expected_failure(abort_code = 1, location = DiemFramework::DiemTimestamp)]
```
wrwg, line 8: I assume the downwards compat behavior is an abort code at any location matches?

language/documentation/examples/diem-framework/move-packages/DPN/tests/ValidatorConfigTests.move
```diff
@@ -121,7 +121,7 @@ module DiemFramework::ValidatorConfigTests {
     }
 
     #[test(tc = @TreasuryCompliance, dr = @DiemRoot)]
```
wrwg, line 123: I think it would generally better we have an extensible syntax like `abort_code = "VC:5"`.

wrwg (commented):
I wonder whether this PR adds any breaking changes to existing tests outside of this repo, and specifically what existing users of Move unit tests maybe doing?

language/documentation/examples/diem-framework/move-packages/DPN/tests/ValidatorConfigTests.move
tnowacki: You can do that too (see example in the summary) :) There just aren't any constants for all of these diem cases, and it didn't seem necessary to add them

language/documentation/examples/diem-framework/move-packages/DPN/tests/AccountFreezingTests.move
tnowacki: A warning is given if it is missing

tnowacki:
> I wonder whether this PR adds any breaking changes to existing tests outside of this repo, and specifically what existing users of Move unit tests maybe doing?

It will throw a warning for _all_ existing tests, but I really think it should. The tests might be wrong, and you probably want to update them to the constant syntax.

junkil-park:
> @wrwg @junkil-park This PR is failing in CI due to a timeout in prover non-linear arithmetics tests. Any idea what's causing the timeout?

Hmm... Prover doesn't timeout on HEAD in my local machine. Does it timeout on your local machine?

If it only timeouts with the nonlinear arithmetic tests, I think you could put "verify=false" on it, and file a github issue for it. It needs a further investigation separately.

tnowacki:
> > @wrwg @junkil-park This PR is failing in CI due to a timeout in prover non-linear arithmetics tests. Any idea what's causing the timeout?
> 
> Hmm... Prover doesn't timeout on HEAD in my local machine. Does it timeout on your local machine?
> 
> If Prover timeouts only with the nonlinear arithmetic tests, I think you could put "verify=false" on it, and file a github issue for it. It needs a further investigation separately.

It passes locally. Only failing on CI for me

tnowacki:
> > @wrwg @junkil-park This PR is failing in CI due to a timeout in prover non-linear arithmetics tests. Any idea what's causing the timeout?
> 
> Hmm... Prover doesn't timeout on HEAD in my local machine. Does it timeout on your local machine?
> 
> If Prover timeouts only with the nonlinear arithmetic tests, I think you could put "verify=false" on it, and file a github issue for it. It needs a further investigation separately.

I've gone ahead and disabled it

language/tools/move-unit-test/src/test_reporter.rs
```diff
@@ -92,20 +92,28 @@ impl TestRunInfo {
...
+    pub fn no_error() -> Self {
+        FailureReason::NoError("Test did not fail as expected".to_string())
+    }
+
+    pub fn wrong_error(expected: MoveError, actual: MoveError) -> Self {
+        FailureReason::WrongError(
+            "Test did not fail with the expected error".to_string(),
+            expected,
+            actual,
+        )
```
amnn, line 104: nit: I found the wording in the test cases a little confusing because of the overload of the term "fail" here (i.e. the test failed because it "did not fail as expected") ...maybe I just spent too long thinking about this.  My suggestion is to reserve the "fail" and "pass" as terms to refer to what tests do, and wording around "errors" and producing them  for the code under test, e.g.

- "Test did not error as expected"
- "Test did not produce expected error"

amnn: ...I also remember that we use `expected_failure` as the attribute name, so, that will already make things less clear cut, so I'm less sure about this comment.  Make of it what you will.

language/tools/move-unit-test/tests/test_sources/cross_module_aborts.evm.exp
```diff
@@ -7,7 +7,7 @@ Test failures:
 Failures in 0x1::B:
 
 ┌── failing_test ──────
-│ Test was not expected to abort but it aborted with 0 here
+│ Test was not expected to fail, but it aborted with code 0 originating in an unknown location. Rooted here
```
amnn, line 10: The "Rooted here" at the end seems to dangle a bit strangely, can we make it part of the same sentence?
```suggestion
│ Test was not expected to fail, but it aborted with code 0 originating in an unknown location rooted here
```

amnn (commented):
I'm also very supportive of this change.  I've reviewed it and it looks good (left some minor comments here and there), but I'm mindful there were some questions about backwards compatibility from @wrwg.  Here's my understanding:

- If you have `expected_failure` with just a u64 literal abort code, these will now start producing a warning -- they need to be replaced with a constant from the module the abort originates from, or have a `location` field added to silence the warning.
- Even if you choose not to do this, the tests will containue to pass/fail as before, but they will just be noisier (with the addition of the warning).
- ~~If you have `expected_failure` and it already references a constant, nothing changes, unless of course, your test was wrong (the abort code was right, but it originated from a different module).~~  You couldn't reference a constant before, so this can't happen.

I don't think there's a less disruptive way to land this improvement, because if the test is wrong or at risk of being wrong, the intent is to let people know about it.

language/tools/move-unit-test/tests/test_sources/non_exsistent_native.exp
```diff
@@ -8,30 +8,7 @@ Test failures:
 Failures in 0x1::M:
 
 ┌── non_existent_native ──────
-│ ITE: An unknown error was reported. Location:
```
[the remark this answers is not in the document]
tnowacki: Huh, that is a strange one... let me take a look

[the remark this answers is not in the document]
tnowacki: Thanks for catching this!

language/move-compiler/tests/move_check/unit_test/expected_failure_out_of_range_value.unit_test.exp
```diff
@@ -1,9 +1,25 @@
+warning[W10007]: potential issue with attribute value
+  ┌─ tests/move_check/unit_test/expected_failure_out_of_range_value.move:6:24
+  │
+6 │     #[expected_failure(abort_code=18446744073709551614)]
+  │                        ^^^^^^^^^^ -------------------- Replace this value with a constant or add the 'location=...' attribute
+  │                        │           
+  │                        WARNING POTENTIALLY INACCURATE TEST. The value should point to a constant defined in the module that will abort to signify which module will abort. Or the location attribute must be specified.
```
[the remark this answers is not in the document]
tnowacki: Do you have a suggestion for shortening it? I don't think we automatically column fit currently

language/tools/move-unit-test/src/test_reporter.rs
tnowacki: I like `"Test did not produce expected error"`, thanks!

language/tools/move-unit-test/src/test_reporter.rs
```diff
@@ -198,21 +225,6 @@ impl TestFailure {
                 )
             }
             FailureReason::Property(message) => message.clone(),
-            FailureReason::Unknown(message) => {
```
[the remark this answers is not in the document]
tnowacki: I don't think it can be encountered. We should always get a `VMError` and then we will end up in `UnexpectedError` or `WrongError`

language/tools/move-unit-test/src/test_runner.rs
```diff
@@ -488,98 +490,91 @@ impl SharedTestingConfig {
...
+                                TestFailure::new(
+                                    FailureReason::timeout(),
+                                    test_run_info,
+                                    Some(err),
+                                    save_session_state(),
+                                ),
+                                test_plan,
+                            )
+                        }
+                        Some(ExpectedFailure::Expected) => {
```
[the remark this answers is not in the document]
tnowacki: > IIUC, the ordering of these cases means that the blanket expected_failure will not capture an out-of-gas error, even though it is now one of the specific failure cases, which seems potentially confusing.

That is correct, but this was the behavior before. And I was trying to preserve that existing behavior. Though I can change it if you want. 

> If anything it seems like the the ExpectedWithCodeDEPRECATED case should be moved below the OUT_OF_GAS case, and the blanket Expected case should be moved above it, thoughts?

ExpectedWithCodeDEPRECATED and OUT_OF_GAS are disjoint cases, as ExpectedWithCodeDEPRECATED only catches aborts

tnowacki:
> * If you have `expected_failure` and it already references a constant, nothing changes, unless of course, your test was wrong (the abort code was right, but it originated from a different module).

The referencing of constants is new, so this case can't happen currently.

language/tools/move-unit-test/src/test_runner.rs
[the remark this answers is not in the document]
tnowacki: I did go ahead and switch things around, I don't think anyone was (probably) relying on the old behavior for OUT_OF_GAS

language/tools/move-unit-test/tests/test_sources/timeout.exp
```diff
@@ -3,7 +3,7 @@ Running Move unit tests
 [ FAIL    ] 0x1::M::no_timeout_fail
 [ PASS    ] 0x1::M::no_timeout_while_loop
 [ TIMEOUT ] 0x1::M::timeout_fail
-[ TIMEOUT ] 0x1::M::timeout_fail_with_expected_failure
+[ PASS    ] 0x1::M::timeout_fail_with_expected_failure
```
amnn, line 6: This doesn't seem right -- is it expected?

language/move-compiler/tests/move_check/unit_test/expected_failure_out_of_range_value.unit_test.exp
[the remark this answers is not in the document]
amnn: Hmm, well seeing as we get the tip about replacing it with a comment or adding the `location` attribute, how about just warning about the problem, something like:

```
WARNING: passes for abort from any module.
```

It may need the tip to be reworded slightly to make it clear why using a constant helps, something like:

```
Replace value with constant from expected module or add `location=...` attribute.
```

language/tools/move-unit-test/src/test_reporter.rs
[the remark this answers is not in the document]
amnn: I thought this might be related to the internal error that got lost, but it seems like that was fixed by other means, so resolving!

amnn (approved):
One more test output that seems to have changed that maybe shouldn't?  Otherwise LGTM!  Accepting to unblock (pending that test), but I know this is a particularly quiet week in the US so I'll leave it with you whether to land it now or later.

Regarding backwards compatibility, is it worth considering adding a flag to silence the warning (off by default), in case people want to bump the compiler but tackle migrating their tests at their own pace?

language/tools/move-unit-test/tests/test_sources/timeout.exp
tnowacki: This is the from the switch you suggested, about having expected failure catching everything, including out of gas

Issue #631, opened 2022-10-28
[Bug][Docgen] `entry` is not displayed in function headers


junkil-park:
See the following example. The actual function signature has `entry` in it, but it's not displayed in the function header:

<img width="1149" alt="image" src="https://user-images.githubusercontent.com/7842757/198522307-1a6e53b5-a41c-4582-b98d-13cd74b53218.png">

wrwg:
Definitely a bug

PR #638, opened 2022-10-30
[docgen] displays "entry" in the function headers of entry functions

junkil-park:
- "entry" keyword was missing in the function headers
- this PR displays the keyword "entry" properly

This PR resolves https://github.com/move-language/move/issues/631

## Motivation

To properly display the "entry" keyword in the function headers

yes

## Test Plan

cargo test

### DIFF
--- a/language/move-prover/move-docgen/src/docgen.rs
+++ b/language/move-prover/move-docgen/src/docgen.rs
@@ -1077,9 +1077,15 @@ impl<'env> Docgen<'env> {
                 return_types.iter().map(|ty| ty.display(tctx)).join(", ")
             ),
         };
+        let entry_str = if func_env.is_entry() && !func_env.module_env.is_script_module() {
+            "entry ".to_owned()
+        } else {
+            "".to_owned()
+        };
         format!(
-            "{}fun {}{}({}){}",
+            "{}{}fun {}{}({}){}",
             func_env.visibility_str(),
+            entry_str,
             name,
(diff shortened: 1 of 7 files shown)


language/move-prover/move-docgen/src/docgen.rs
```diff
@@ -1077,9 +1077,15 @@ impl<'env> Docgen<'env> {
                 return_types.iter().map(|ty| ty.display(tctx)).join(", ")
             ),
         };
+        let script_str = if func_env.is_entry() && !func_env.module_env.is_script_module() {
```
meng-xu-cs, line 1080: Maybe call the variable `entry_str`?

junkil-park: Done. Good call!

junkil-park:
By the way, it seems that both of the following functions satisfy `is_script()` and `is_entry()`:
(1) a entry function a regular module
(2) a function in a script module
Is it intended?

Shouldn't it be like only (1) satisfies `is_entry()`, and only (2) satisfies `is_script()`?

rahxephon89 (approved):

meng-xu-cs:
> By the way, it seems that both of the following functions satisfy both is_script() and is_entry():
> (1) a entry function in a regular module
> (2) a function in a script module
> Is it intended?

My understanding is that `script` function is deprecated and replaced with `entry` function, so I am okay with `is_script` and `is_entry` being semantically equivalent. (and maybe deprecate `is_script` in the future?)

PR #642, opened 2022-11-02
Alternative paranoid mode

runtian-zhou:
## Motivation

This is an alternative impl of #598. In this impl the type information are split out from the value.

(Write your answer here.)

## Test Plan

### DIFF
--- a/language/move-vm/runtime/src/interpreter.rs
+++ b/language/move-vm/runtime/src/interpreter.rs
@@ -10,7 +10,7 @@ use crate::{
 use fail::fail_point;
 use move_binary_format::{
     errors::*,
-    file_format::{Bytecode, FunctionHandleIndex, FunctionInstantiationIndex},
+    file_format::{Ability, AbilitySet, Bytecode, FunctionHandleIndex, FunctionInstantiationIndex},
 };
 use move_core_types::{
     account_address::AccountAddress,
@@ -67,6 +67,8 @@ pub(crate) struct Interpreter {
     operand_stack: Stack,
     /// The stack of active functions.
     call_stack: CallStack,
+    /// Whether to perform a paranoid type safety checks at runtime.
+    paranoid_type_checks: bool,
 }
 
 struct TypeWithLoader<'a, 'b> {
@@ -95,6 +97,7 @@ impl Interpreter {
         Interpreter {
             operand_stack: Stack::new(),
             call_stack: CallStack::new(),
+            paranoid_type_checks: loader.vm_config().paranoid_type_checks,
         }
         .execute_main(
             loader, data_store, gas_meter, extensions, function, ty_args, args,
@@ -124,7 +127,9 @@ impl Interpreter {
                 .map_err(|e| self.set_location(e))?;
         }
 
-        let mut current_frame = Frame::new(function, ty_args, locals);
+        let mut current_frame = self
+            .make_new_frame(function, ty_args, locals)
+            .map_err(|err| self.set_location(err))?;
         loop {
             let resolver = current_frame.resolver(loader);
             let exit_code =
@@ -149,11 +154,15 @@ impl Interpreter {
                         current_frame.pc += 1; // advance past the Call instruction in the caller
                     } else {
                         // end of execution. `self` should no longer be used afterward
-                        return Ok(self.operand_stack.0);
+                        return Ok(self.operand_stack.value);
                     }
                 }
                 ExitCode::Call(fh_idx) => {
                     let func = resolver.function_from_handle(fh_idx);
+
+                    if self.paranoid_type_checks {
+                        self.check_friend_or_private_call(&current_frame.function, &func)?;
+                    }
 
                     // Charge gas
                     let module_id = func
@@ -188,6 +197,7 @@ impl Interpreter {
                     }
                     let frame = self
                         .make_call_frame(func, vec![])
+                        .map_err(|e| self.set_location(e))
                         .map_err(|err| self.maybe_core_dump(err, &current_frame))?;
                     self.call_stack.push(current_frame).map_err(|frame| {
                         let err = PartialVMError::new(StatusCode::CALL_STACK_OVERFLOW);
@@ -203,6 +213,10 @@ impl Interpreter {
                         .instantiate_generic_function(idx, current_frame.ty_args())
                         .map_err(|e| set_err_info!(current_frame, e))?;
                     let func = resolver.function_from_instantiation(idx);
+
+                    if self.paranoid_type_checks {
+                        self.check_friend_or_private_call(&current_frame.function, &func)?;
+                    }
 
                     // Charge gas
                     let module_id = func
@@ -233,6 +247,7 @@ impl Interpreter {
                     }
                     let frame = self
                         .make_call_frame(func, ty_args)
+                        .map_err(|e| self.set_location(e))
                         .map_err(|err| self.maybe_core_dump(err, &current_frame))?;
                     self.call_stack.push(current_frame).map_err(|frame| {
                         let err = PartialVMError::new(StatusCode::CALL_STACK_OVERFLOW);
@@ -250,18 +265,59 @@ impl Interpreter {
     ///
     /// Native functions do not push a frame at the moment and as such errors from a native
     /// function are incorrectly attributed to the caller.
-    fn make_call_frame(&mut self, func: Arc<Function>, ty_args: Vec<Type>) -> VMResult<Frame> {
+    fn make_call_frame(
+        &mut self,
+        func: Arc<Function>,
+        ty_args: Vec<Type>,
+    ) -> PartialVMResult<Frame> {
         let mut locals = Locals::new(func.local_count());
         let arg_count = func.arg_count();
+        let is_generic = !ty_args.is_empty();
         for i in 0..arg_count {
-            locals
-                .store_loc(
-                    arg_count - i - 1,
-                    self.operand_stack.pop().map_err(|e| self.set_location(e))?,
-                )
-                .map_err(|e| self.set_location(e))?;
+            locals.store_loc(arg_count - i - 1, self.operand_stack.pop()?)?;
+
+            if self.paranoid_type_checks {
+                let ty = self.operand_stack.pop_ty()?;
+                if is_generic {
+                    ty.check_eq(&func.local_types()[arg_count - i - 1].subst(&ty_args)?)?;
+                } else {
+                    // Directly check against the expected type to save a clone here.
+                    ty.check_eq(&func.local_types()[arg_count - i - 1])?;
+                }
+            }
         }
-        Ok(Frame::new(func, ty_args, locals))
+        self.make_new_frame(func, ty_args, locals)
+    }
+
+    /// Create a new `Frame` given a `Function` and the function `Locals`.
+    ///
+    /// The locals must be loaded before calling this.
+    fn make_new_frame(
+        &self,
+        function: Arc<Function>,
+        ty_args: Vec<Type>,
+        locals: Locals,
+    ) -> PartialVMResult<Frame> {
+        let local_tys = if self.paranoid_type_checks {
+            if ty_args.is_empty() {
+                function.local_types().to_vec()
+            } else {
+                function
+                    .local_types()
+                    .iter()
+                    .map(|ty| ty.subst(&ty_args))
+                    .collect::<PartialVMResult<Vec<_>>>()?
+            }
+        } else {
+            vec![]
+        };
+        Ok(Frame {
+            pc: 0,
+            locals,
+            function,
+            ty_args,
+            local_tys,
+        })
     }
 
     /// Call a native functions.
@@ -310,6 +366,16 @@ impl Interpreter {
         for _ in 0..expected_args {
             args.push_front(self.operand_stack.pop()?);
         }
+
+        if self.paranoid_type_checks {
+            for i in 0..expected_args {
+                let expected_ty =
+                    function.parameter_types()[expected_args - i - 1].subst(&ty_args)?;
+                let ty = self.operand_stack.pop_ty()?;
+                ty.check_eq(&expected_ty)?;
+            }
+        }
+
         let mut native_context = NativeContext::new(self, data_store, resolver, extensions);
         let native_function = function.get_native()?;
 
@@ -321,7 +387,7 @@ impl Interpreter {
             args.iter(),
         )?;
 
-        let result = native_function(&mut native_context, ty_args, args)?;
+        let result = native_function(&mut native_context, ty_args.clone(), args)?;
 
         // Note(Gas): The order by which gas is charged / error gets returned MUST NOT be modified
         //            here or otherwise it becomes an incompatible change!!!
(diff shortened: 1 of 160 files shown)


language/move-vm/types/src/loaded_data/annotated_types.rs
```diff
@@ -0,0 +1,174 @@
...
+#[derive(Debug, Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct ParanoidStructTypes {
+    pub fields: Vec<ParanoidType>,
+    pub abilities: AbilitySet,
+    pub idx: CachedStructIndex,
+    pub instantiations: Vec<ParanoidType>,
+}
+
+#[derive(Debug, Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub enum ParanoidType {
```
runtian-zhou, line 23: This is the new separated type tag for values

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -732,19 +844,25 @@ const OPERAND_STACK_SIZE_LIMIT: usize = 1024;
 const CALL_STACK_SIZE_LIMIT: usize = 1024;
 
 /// The operand stack.
-struct Stack(Vec<Value>);
+struct Stack {
+    value: Vec<Value>,
+    types: Vec<ParanoidType>,
```
runtian-zhou, line 849: Stacks now have a vector of types along side the value.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -824,8 +982,10 @@ impl CallStack {
 struct Frame {
     pc: u16,
     locals: Locals,
+    local_tys: Vec<ParanoidType>,
```
runtian-zhou, line 985: Locals now carry a type as well.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -864,6 +1032,188 @@ impl Frame {
...
+            .ok_or_else(|| PartialVMError::new(StatusCode::VERIFIER_INVARIANT_VIOLATION))
+    }
+
+    fn get_mut_local_ty(&self, idx: u8) -> PartialVMResult<&mut ParanoidType> {
+        self.local_tys
+            .get_mut(idx as usize)
+            .ok_or_else(|| PartialVMError::new(StatusCode::VERIFIER_INVARIANT_VIOLATION))
+    }
+
+    fn type_stack_transition(
```
runtian-zhou, line 1047: This is the main entrypoint for the paranoid checks.

language/move-vm/runtime/src/loader.rs
```diff
@@ -2032,6 +2188,7 @@ impl Function {
             type_parameters,
             native,
             def_is_native,
+            def_is_friend_or_private,
```
tnowacki, line 2191: I haven't looked where this is used, but wouldn't you want to know the exact visibility?

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -246,6 +273,55 @@ impl Interpreter {
...
+    /// Make sure only private/friend function can only be invoked by modules under the same address.
+    fn check_friend_or_private_call(
+        &self,
+        caller: &Arc<Function>,
+        callee: &Arc<Function>,
+    ) -> VMResult<()> {
+        if callee.is_friend_or_private() {
+            match (caller.module_id(), callee.module_id()) {
+                (Some(caller_id), Some(callee_id)) => {
+                    if caller_id.address() == callee_id.address() {
```
tnowacki, line 285: This isn't inherent to friends. Like sure we have this check in friends right now, but it might not always be the case. I would probably check if the module is in the friend list?

It also doesn't catch invalid private access from other addresses

language/move-vm/runtime/src/interpreter.rs
tnowacki, line 1047: Yeah, this wasn't quite what I was expecting, though this may be easier? I'm not sure. 

I was imagining replacing all interactions with the stacks with macros 

`pop` and `push` and `popn`/`pushn`/`lastn`. 

This way you could do something like `let (ty, value) = pop!(interpreter)` And even swap out the type for `ty` with a `cfg` feature flag if we want to disable the paranoid mode

language/move-vm/runtime/src/interpreter.rs
sblackshear: Nice, this seems less disruptive than having a stack of (Value, Type) pairs

language/move-vm/runtime/src/interpreter.rs
sblackshear: I think these can probably be read directly from `function.locals`

runtian-zhou:
As mentioned, we would prefer this implementation over the one in #598. Thus I'll solicit the review here. cc @sblackshear @tnowacki @vgao1996

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -251,18 +271,28 @@ impl Interpreter {
     ///
     /// Native functions do not push a frame at the moment and as such errors from a native
     /// function are incorrectly attributed to the caller.
-    fn make_call_frame(&mut self, func: Arc<Function>, ty_args: Vec<Type>) -> VMResult<Frame> {
+    fn make_call_frame(
+        &mut self,
+        func: Arc<Function>,
+        ty_args: Vec<Type>,
+        paranoid: bool,
```
vgao1996, line 278: I think you can directly access `self.runtime_config.paranoid_type_checks` without passing the flag in, no?

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -311,6 +341,20 @@ impl Interpreter {
         for _ in 0..expected_args {
             args.push_front(self.operand_stack.pop()?);
         }
+
+        if self.runtime_config.paranoid_type_checks {
+            let is_generic = !ty_args.is_empty();
```
vgao1996, line 346: Nit: you don't have to branch based on whether the function is generic or not. Just call `function.parameter_types()[expected_args - i - 1].subst(&ty_args)?` and it'll work for both generic and non-generic functions -- if `ty_args` is empty, then the substitution is effectively equivalent to a clone.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -356,9 +400,52 @@ impl Interpreter {
         for value in return_values {
             self.operand_stack.push(value)?;
         }
+
+        if self.runtime_config.paranoid_type_checks {
+            let is_generic = !ty_args.is_empty();
```
vgao1996, line 405: Similar here. No need to differentiate the two cases.

vgao1996 (approved):
Overall looks good to me. I like this implementation! Glad it turns out much simpler than the original approach.

Let's still leave @sblackshear @tnowacki and maybe @jolestar some time to take a look

jolestar (commented):
I like this solution. 

I have an idea, maybe we can check the type in `pop_as<T>`, and make sure the type in the stack matches the T.

Then we can eliminate `simple_instruction_type_stack_transition` and handle the instruction type and value together.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -869,6 +1033,504 @@ impl Frame {
...
+                let field_count = resolver.field_count(*idx);
+                let args_ty = resolver.get_struct_fields(*idx)?;
+                let output_ty = resolver.get_struct_type(*idx);
+                let ability = resolver.loader().abilities(&output_ty)?;
+                let field_abilities = if ability.has_key() {
+                    ability
+                        .remove(Ability::Key)
+                        .union(AbilitySet::singleton(Ability::Store))
+                } else {
+                    ability
```
sblackshear, line 1179: It took me a minute to convince myself that this is correct in the case that the struct being packed has `store`, but not `key`? e.g.,
```
struct A {}
struct B has store {
  a: A
}

let x = B { a: A{} } // should get flagged by runtime checks
```

A comment on why this is safe might be helpful.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -869,6 +1033,504 @@ impl Frame {
...
+            }
+            Bytecode::MutBorrowGlobal(idx) => {
+                interpreter
+                    .operand_stack
+                    .pop_ty()?
+                    .check_eq(&Type::Address)?;
+                let ty = resolver.get_struct_type(*idx);
+                interpreter
+                    .operand_stack
+                    .push_ty(Type::MutableReference(Box::new(ty)))?;
```
sblackshear, line 1353: I think we want to check for the `key` ability in all of the global storage operators. The bytecode verifier does this, anway.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -869,6 +1033,504 @@ impl Frame {
...
+                let ty = resolver.instantiate_generic_type(*idx, ty_args)?;
+                interpreter.operand_stack.push_ty(ty)?;
+            }
+            Bytecode::FreezeRef => {
+                match interpreter.operand_stack.pop_ty()? {
+                    Type::MutableReference(ty) => {
+                        interpreter.operand_stack.push_ty(Type::Reference(ty))?
+                    }
+                    Type::Reference(ty) => {
+                        interpreter.operand_stack.push_ty(Type::Reference(ty))?
```
sblackshear, line 1432: I think we want this to be an error. The verifier will not allow you to apply `FreezeRef` to an immutable reference.

sblackshear (commented):
Like what you've done with this approach Runtian! Flagged a few bugs (I think, `StoreLoc` being the most high-consequence one) and a few cosmetic items.

language/move-vm/runtime/src/interpreter.rs
runtian-zhou: There's a very interesting observation here: it seems that for vector instructions we allow implicit reference casting (mut reference can be used as immut reference). e.g: VecLen

language/move-vm/runtime/src/interpreter.rs
vgao1996: > efficiency of subst

No need to worry about this. `subst`'s implementation is near identical to clone.

runtian-zhou:
> I have an idea, maybe we can check the type in pop_as<T>, and make sure the type in the stack matches the T.

I tried that previously. The api would look more like

`fn pop_with_checks<F>(&self, check_fn: F) where F: FnOnce(Type) -> PartialVMResult<()>`

It worked but it was just quite ugly as the type transition was tangled with value pops and the code didn't seem as clean as this version.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -869,6 +1033,504 @@ impl Frame {
...
+            | Bytecode::Mod
+            | Bytecode::Div
+            | Bytecode::BitOr
+            | Bytecode::BitAnd
+            | Bytecode::Xor
+            | Bytecode::Or
+            | Bytecode::And => {
+                let lhs = interpreter.operand_stack.pop_ty()?;
+                let rhs = interpreter.operand_stack.pop_ty()?;
+                lhs.check_eq(&rhs)?;
```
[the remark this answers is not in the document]
runtian-zhou: So we don't check the types for primitive value operations, including integer operations, boolean operations and comparision operations. The reason why they were safe is because our value system have separate representation between primitive values and struct values, so had there been a type confusion they would be caught at the runtime. I can either add a test to make sure it doesn't regress or I can add the checks here. What would be your preference?

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -150,12 +160,16 @@ impl Interpreter {
...
                     } else {
                         // end of execution. `self` should no longer be used afterward
-                        return Ok(self.operand_stack.0);
+                        return Ok(self.operand_stack.value);
                     }
                 }
                 ExitCode::Call(fh_idx) => {
                     let func = resolver.function_from_handle(fh_idx);
 
+                    if self.runtime_config.paranoid_type_checks {
```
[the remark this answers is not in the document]
runtian-zhou: Same. We don't have `config` for `Frame`.. Sry it was a bit confusing.

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -869,6 +1033,504 @@ impl Frame {
...
+                    if !locals.is_invalid(idx)? {
+                        check_ability(resolver.loader().abilities(&ty)?.has_drop())?;
+                    }
+                }
+            }
+            Bytecode::Abort => {
+                interpreter.operand_stack.pop_ty()?;
+            }
+            // StLoc needs to check before execution as we need to check the drop ability of values.
+            Bytecode::StLoc(idx) => {
```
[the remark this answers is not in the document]
sblackshear: I see, I did not realize there were two sets of checks. What would go wrong if we did all the checks either before execution or after (minus the gas issue mentioned on "before", which seems fairly minor)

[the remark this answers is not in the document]
runtian-zhou: For some of the operations, the cost of resolving the types seems non trivial to me. That's the main reason that makes me feel that we should only do the checks after we've charged proper gas here. Would be great if @vgao1996 can weigh in your gas experience here. To answer your question:

- Calling everything before will work, but just that we might run the checks before we charge gas, and some of those checks could be more expensive. 
- Calling everything after won't work for all control flow operators and `StLoc`. The reason is that for control flow oeprators the after path won't be invoked at all. The `StLoc` is a very interesting case here. The reason why it won't work is because we need to check if the local we can potentially overwrite has the drop ability, and the only way to check that is before we execute the instruction.

sblackshear:
> There's a very interesting observation here: it seems that for vector instructions we allow implicit reference casting (mut reference can be used as immut reference). e.g: VecLen

Yep, and same for (e.g.) BorrowFieldImmut. Only function calls require strict reference type equality.

language/move-vm/runtime/src/interpreter.rs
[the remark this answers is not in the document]
sblackshear: Either sounds great to me--as long as something will break if we switch to a different value representation in the future. Thanks!

language/move-vm/runtime/src/interpreter.rs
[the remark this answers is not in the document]
sblackshear: Thanks for the explanation. My bias would be for clarity/readability--these checks should only be run without charging gas in the (hopefully rare/impossible) event that that there's a bug in the bytecode verifier, and the cost of any individual instruction is (hopefully) very small. So I think the scenario that splitting the checks is protecting against is both unlikely and low consequence. Whereas a bug in the checks that happens due to the before/after split would be unlikely and high consequence.

This is subjective and I'm ok with whatever you decide on this.

sblackshear (approved):

runtian-zhou:
Had a chat with @vgao1996 and we would land the PR in the current form, i.e: run the checks at two locations: pre instruction execution and post instruction execution. The reason is that we want to run as less code as possible before we charge proper gas. I updated the comment to address the gas concern here. To address @sblackshear's concern on making sure the two functions constitutes the full type safety checks I updated the code to use exhaustive match to make sure we cannot miss checks in the future if we decide to add new instructions.

Thanks for everyone for looking into this! Added a few comments and I'll merge this PR shortly.

PR #643, opened 2022-11-03
[build] Added additional diagnostics for fetching deps

awelc:
## Motivation

When building a package whose dependencies are not yet cached and have to be fetched from Git may result in a really long wait before anything gets printed to the console to the point that it appears that the build is hung. This PR adds an extra bit of diagnostics that's printed when fetching (only) remote dependencies.

Yes

### DIFF
--- a/language/tools/move-cli/src/base/build.rs
+++ b/language/tools/move-cli/src/base/build.rs
@@ -19,7 +19,7 @@ impl Build {
             if config.test_mode {
                 config.dev_mode = true;
             }
-            config.download_deps_for_package(&rerooted_path)?;
+            config.download_deps_for_package(&rerooted_path, &mut std::io::stdout())?;
             return Ok(());
         }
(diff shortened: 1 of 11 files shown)


language/tools/move-cli/src/base/build.rs
```diff
@@ -19,7 +19,7 @@ impl Build {
             if config.test_mode {
                 config.dev_mode = true;
             }
-            config.download_deps_for_package(&rerooted_path)?;
+            config.download_deps_for_package(&rerooted_path, &mut std::io::stdout())?;
             return Ok(());
         }
         let architecture = config.architecture.unwrap_or(Architecture::Move);
```
tnowacki, line 25: Can't comment that far down, but could you switch the other one below to std::io::out?

tnowacki (approved):
Looks great!

PR #646, opened 2022-11-05
[move-prover] Allow shift operation to abort in the move prover

rahxephon89:
## Motivation

As the first step to fix the behavior of shift operation in the prover, this PR changes prelude to allow shift to abort. 

Yes

## Test Plan

cargo test


junkil-park:
Great! By the way, is there anything we can do about code and spec having different meanings about the shift operations? (https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/move-stdlib/sources/error.move#L69)

rahxephon89:
> Great! By the way, is there anything we can do about code and spec having different meanings about the shift operations? (https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/move-stdlib/sources/error.move#L69)

Yes, I will try to work on that once this one is landed.

junkil-park (approved):
LGTM

meng-xu-cs (approved):
Nice!

rahxephon89:
@junkil-park , @meng-xu-cs, @wrwg   Note that when running `move prove` against move-stdlib in aptos-core, timeout will happen: 
```
error: verification out of resources/timeout (global timeout set to 40s)
   ┌─ ./sources/fixed_point32.move:63:5
   │  
63 │ ╭     public fun divide_u64(val: u64, divisor: FixedPoint32): u64 {
64 │ │         // Check for division by zero.
65 │ │         assert!(divisor.value != 0, EDIVISION_BY_ZERO);
66 │ │         // First convert to 128 bits and then shift left to
   · │
74 │ │         (quotient as u64)
75 │ │     }
   │ ╰─────^
```
If the time out is set to more than 50s, the timeout error will disappear. I am wondering whether it is OK to merge then.

PR #652, opened 2022-11-07
[byecode utils] make struct layout generation work with serdegen

sblackshear:
One of the promises of generating layout YAML's for Move structs is that we can leverage `serdegen` to autogenerate wrapper types for Move structs in all the languages `serdegen` handles. This PR moves us closer to that.
    
- Add `separator` option that allows replacing the Move source syntax separators ("::", "<", ">", ",") in type names with a user
-defined string. This is helpful because a name like `0x1::M::T<u64>` is very unlikely to be a valid identifier in any of the langua
ges `serdegen` handles. Setting (e.g.) `separator = __` allows a fully qualified Move type name to be a valid identifier.
- Add an `omit_addresses` option that omits the addresses from a fully qualified Move type name. This is useful in the (common) 
case where you are generating bindings for a package that does not have any name conflicts and don't want to bother with the addresses. The generator looks for name conflicts and panics if it finds one.
- Add an `ignore_phantom_types` option that does not include phantom type parameters in struct names, since they do not influence layout.
    
After these changes, the following PoC for serdegen works:    
    ```
    move sandbox generate struct-layouts --module storage/0x00000000000000000000000000000001/modules/M1.mv --struct G --omit-addresses --separator "__" > test.yaml
    serdegen --language python3 test.yaml > test.py
    python test.py
    ```
    
, where `test.py` is
    
```python
from dataclasses import dataclass
import typing
import serde_types as st
    
@dataclass(frozen=True)
class M1__G:
    x: st.uint64
    s: "M1__S__bool__"
    
@dataclass(frozen=True)
class M1__S__bool__:
    t: bool
```

### DIFF
--- a/language/tools/move-bytecode-utils/src/layout.rs
+++ b/language/tools/move-bytecode-utils/src/layout.rs
@@ -17,7 +17,12 @@ use move_core_types::{
     value::{MoveFieldLayout, MoveStructLayout, MoveTypeLayout},
 };
 use serde_reflection::{ContainerFormat, Format, Named, Registry};
-use std::{borrow::Borrow, collections::BTreeMap, convert::TryInto, fmt::Debug};
+use std::{
+    borrow::Borrow,
+    collections::BTreeMap,
+    convert::TryInto,
+    fmt::{Debug, Write},
+};
 
 /// Name of the Move `address` type in the serde registry
 const ADDRESS: &str = "AccountAddress";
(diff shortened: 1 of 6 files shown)


tnowacki:
For `separator`, if it is not a valid Move Identifier character, e.g. `%`, is it possible for there to be collisions? I feel like the answer is no, but I have trouble convincing myself it is true. (I think it is only true because _all_ separators get replaced this way).

Do we care about the case where there are collisions if you use something like `___` (though I think it would require a bad actor even in that case)? 

Or maybe you catch all of these collisions in the same code path that does it for `omit_addresses`? (still reading, so maybe this will become obvious later)

language/tools/move-bytecode-utils/src/layout.rs
```diff
@@ -158,55 +174,115 @@ impl<'a, T: GetModule> SerdeLayoutBuilder<'a, T> {
...
     ) -> Result<Format, T::Error> {
         // build a human-readable name for the struct type. this should do the same thing as
         // StructTag::display(), but it's not easy to use that code here
-        let struct_key = if type_arguments.is_empty() {
-            format!("{}::{}", module_id, name)
+
+        let declaring_module = self
+            .module_resolver
+            .get_module_by_id(module_id)?
+            .expect("Failed to resolve module");
```
tnowacki, line 184: I see the function returns a `Result`. Should these panics and asserts be errors?

language/tools/move-bytecode-utils/src/layout.rs
```diff
@@ -158,55 +174,115 @@ impl<'a, T: GetModule> SerdeLayoutBuilder<'a, T> {
...
+        assert_eq!(
+            normalized_struct.type_parameters.len(),
+            type_arguments.len(),
+            "Wrong number of type arguments for struct"
+        );
+
+        let generics: Vec<String> = type_arguments
+            .iter()
+            .zip(normalized_struct.type_parameters.iter())
+            .filter_map(|(type_arg, type_param)| {
```
tnowacki, line 205: minor nit, I like `filter_map`, but I feel like this would be more readable with distinct `filter` and `map` calls, since neither the filter function nor the mapping function inherently return an `Option`

language/tools/move-bytecode-utils/src/layout.rs
```diff
@@ -158,55 +174,115 @@ impl<'a, T: GetModule> SerdeLayoutBuilder<'a, T> {
...
+                "{}{}{}",
+                struct_key_module,
+                self.config.separator.as_deref().unwrap_or("::"),
+                name
+            )
         } else {
-            let generics: Vec<String> = type_arguments.iter().map(print_format_type).collect();
-            format!("{}::{}<{}>", module_id, name, generics.join(","))
+            format!(
+                "{}{}{}{}{}{}",
```
tnowacki, line 233: Maybe this whole pattern section would be easier to read with `write!`?

e.g. 
```
let mut struct_key = String::new();
if !self.config.omit_addresses {
  write!(f, "{}{}", module_id.address(), self.config.separator.as_deref().unwrap_or("::"));
}
write!(f, "{}{}{}", module_id.name(), self.config.separator.as_deref().unwrap_or("::"), name);
if !generics.is_empty() {
   ... 
}
```
This way you would get a lot more code sharing

language/tools/move-bytecode-utils/src/layout.rs
```diff
@@ -158,55 +174,115 @@ impl<'a, T: GetModule> SerdeLayoutBuilder<'a, T> {
...
-                .find_struct_def_by_name(name)
-                .unwrap_or_else(|| {
+        if self.config.shallow {
+            return Ok(Format::TypeName(struct_key));
+        }
+
+        if let Some(old_struct) = self.registry.get(&struct_key) {
+            if self.config.omit_addresses {
+                // check for conflicts (e.g., 0x1::M::T and 0x2::M::T that both get stripped to M::T because
+                // omit_addresses is on)
```
tnowacki, line 249: Ah, I think this is what I was talking about in my top level comment. Should we report conflicts regardless?

language/tools/move-cli/tests/sandbox_tests/generate_struct_layout/args.exp
```diff
@@ -5,16 +5,6 @@ Command `sandbox generate struct-layouts --module storage/0x00000000000000000000
   STRUCT:
     - t: U64
     - b: BOOL
-AccountAddress:
```
tnowacki, line 8: Could you explain the test cases changes? I don't think I understand the code enough to understand the change

language/tools/move-bytecode-utils/src/layout.rs
sblackshear: Good suggestion, that is much more concise

tnowacki: I don't think string can be used with `write` unfortunately. I think `+=` might work though?

sblackshear: It works with `string`.

language/tools/move-bytecode-utils/src/layout.rs
sblackshear: I found the conditional logic harder to write/understand by using the filter/map, so left it as is.

sblackshear:
> For `separator`, if it is not a valid Move Identifier character, e.g. `%`, is it possible for there to be collisions? I feel like the answer is no, but I have trouble convincing myself it is true. (I think it is only true because _all_ separators get replaced this way).
> 
> Do we care about the case where there are collisions if you use something like `___` (though I think it would require a bad actor even in that case)?
> 
> Or maybe you catch all of these collisions in the same code path that does it for `omit_addresses`? (still reading, so maybe this will become obvious later)

Good call--added collision checks whenever separator is not `None`

tnowacki (approved):

PR #654, opened 2022-11-07
[vm][binary-format] implements proper versioning for the code deserializer inside the VM

vgao1996:
This adds a new VM config option that would allow one to specify the maximum binary format version acceptable dynamically, making it possible for downstream clients to roll out binary format changes (e.g. https://github.com/move-language/move/pull/547) in a backward compatible way.

**Note: this PR contains multiple small commits and it might be easier to review them one at a time.**

### DIFF
--- a/language/move-binary-format/src/deserializer.rs
+++ b/language/move-binary-format/src/deserializer.rs
@@ -12,7 +12,15 @@ use std::{collections::HashSet, convert::TryInto, io::Read};
 impl CompiledScript {
     /// Deserializes a &[u8] slice into a `CompiledScript` instance.
     pub fn deserialize(binary: &[u8]) -> BinaryLoaderResult<Self> {
-        let script = CompiledScript::deserialize_no_check_bounds(binary)?;
+        Self::deserialize_with_max_version(binary, VERSION_MAX)
+    }
+
+    /// Deserializes a &[u8] slice into a `CompiledScript` instance.
+    pub fn deserialize_with_max_version(
+        binary: &[u8],
+        max_binary_format_version: u32,
+    ) -> BinaryLoaderResult<Self> {
+        let script = deserialize_compiled_script(binary, max_binary_format_version)?;
         BoundsChecker::verify_script(&script)?;
         Ok(script)
     }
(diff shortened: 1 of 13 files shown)


language/move-binary-format/src/deserializer.rs
```diff
@@ -12,29 +12,46 @@ use std::{collections::HashSet, convert::TryInto, io::Read};
...
-        let script = CompiledScript::deserialize_no_check_bounds(binary)?;
+        Self::deserialize_with_max_version(binary, VERSION_MAX)
+    }
+
+    /// Deserializes a &[u8] slice into a `CompiledScript` instance.
+    pub fn deserialize_with_max_version(
+        binary: &[u8],
+        max_binary_format_version: u32,
+    ) -> BinaryLoaderResult<Self> {
+        let script = deserialize_compiled_script(binary, max_binary_format_version)?;
```
jolestar, line 23: Maybe we can provide a function to auto-convert the new binary format to the old format.

If a new binary module does not contain the new Bytecodes, it will be converted to the old format, like  https://github.com/starcoinorg/starcoin/blob/master/vm/compiler/src/bytecode_transpose.rs

This can let the upgrade progress more gracefully.

vgao1996: This is an interesting idea I think we should definitely explore! Conceptually I see this as adding an additional compatibility mode or changing `max_binary_format_version` into an enum:
```
enum BinaryFormatCompatibilityOption {
    Upto(u32),
    CompatibleWIth(u32),
}
```

However I'd suggest we try not to squeeze this into this PR since 
- This may require some additional discussions and design work 
- The relaxed requirement does NOT actually address the VM's immediate backward compatibility concern. Let me explain:
  - Right now, existing validators run the old software, which outright rejects binary format v6
  - You want your new software to have the ability to
    - Maintain full backward compatibility with the old one, meaning that they need to have a mode that would allow them to fully reject format v6. Otherwise validators running old and new software can go out of sync during an upgrade.
    - Accept format v6 with some configuration changes
  - This means, the current behavior I'm implementing in this PR (the Upto variant in the enum above) needs to be an option even with the addition of the more relaxed mode

tnowacki (approved):
Thanks!

runtian-zhou (approved):
LGTM! The only complaint I may have is on `VMConfig`. It has `VerifierConfig` as well as `max_binary_format_version`, where the former is a group of config and the second one is a config for runtime. To match the naming I thought having a separate `RuntimeConfig` struct might make more sense but feel free to proceed as is.

vgao1996:
@runtian-zhou yeah I actually considered the alternate approach you mentioned seriously and eventually felt like it would be easier to have just one config object to pass-around rather than two. We can always change this in the future if we find the other way better

PR #655, opened 2022-11-08
[move-prover] Add mod to $shl functions

rahxephon89:
## Motivation

Current implementation of shl operation in the prelude.bpl does not match with the semantics of `<<` in the Move language. For instance, in the following spec, `category << 16` must be rewritten as `(category * 65536) % 18446744073709551616`. 

```
public fun canonical(category: u64, reason: u64): u64 {
    (category << 16) + reason
  }
  spec canonical {
    pragma opaque = true;
    // TODO: `<<` has different meanings in code and spec in case of overvlow.
    let shl_res = (category * 65536) % 18446744073709551616; // (category << 16)
    ensures [concrete] result == shl_res + reason;
    aborts_if [abstract] false;
    ensures [abstract] result == category;
  }
```

To solve this issue, this PR:

1) adds modulo operation to the shl operation;
2) allows explicit type casting in the spec language, which is useful when specifying properties related to shift operators. An example is given below:

```
    public fun create_from_rational(numerator: u64, denominator: u64): FixedPoint32 {
        let scaled_numerator = (numerator as u128) << 64;
        let scaled_denominator = (denominator as u128) << 32;
        assert!(scaled_denominator != 0, EDENOMINATOR);
        let quotient = scaled_numerator / scaled_denominator;
        assert!(quotient != 0 || numerator == 0, ERATIO_OUT_OF_RANGE);
        assert!(quotient <= MAX_U64, ERATIO_OUT_OF_RANGE);
        FixedPoint32 { value: (quotient as u64) }
    }
    spec create_from_rational {
        pragma opaque;
        include CreateFromRationalAbortsIf;
        ensures result == spec_create_from_rational(numerator, denominator);
    }
    spec schema CreateFromRationalAbortsIf {
        numerator: u64;
        denominator: u64;
        let scaled_numerator = (numerator as u128) << 64;
        let scaled_denominator = (denominator as u128) << 32;
        let quotient = scaled_numerator / scaled_denominator;
        aborts_if scaled_denominator == 0 with EDENOMINATOR;
        aborts_if quotient == 0 && scaled_numerator != 0 with ERATIO_OUT_OF_RANGE;
        aborts_if quotient > MAX_U64 with ERATIO_OUT_OF_RANGE;
    }
    spec fun spec_create_from_rational(numerator: num, denominator: num): FixedPoint32 {
        FixedPoint32{value: (numerator << 64) / (denominator << 32)}
    }
```

Yes

## Test Plan

cargo test


rahxephon89:
> Nó

The biggest issue for adding mod is correctly getting (or marking) the concrete type for the shift operator for expressions in the spec. In the expression `let scaled_value = val << 32;`, the type of val will be inferred as u64, which leads to incorrect result. Please feel free to add more concrete comments to the PR, thanks!

```
public fun create_from_u64(val: u64): FixedPoint32 {
        let value = (val as u128) << 32;
        assert!(value <= MAX_U64, ERATIO_OUT_OF_RANGE);
        FixedPoint32{value: (value as u64)}
    }
    spec create_from_u64 {
        pragma opaque;
        include CreateFromU64AbortsIf;
        ensures result == spec_create_from_u64(val);
    }
    spec schema CreateFromU64AbortsIf {
        val: num;
        let scaled_value = val << 32;
        aborts_if scaled_value > MAX_U64;
    }
```

language/move-prover/tests/sources/functional/fixed_point_arithm.move
```diff
@@ -108,7 +108,7 @@ module 0x42::FixedPointArithmetic {
     }
     spec mul_div_incorrect {
         ensures result >= x; // disproved
-        ensures result == x; // disproved
+        //ensures result == x; // disproved; commented out because of timeout
```
meng-xu-cs, line 111: - add a TODO
- and how often is the timeout happening (and if we increase the timeout, can this be proved?)

meng-xu-cs (approved):

language/move-prover/tests/sources/functional/fixed_point_arithm.move
rahxephon89: This is an interesting case. If you prove these four ensures separately, all of them will be disproved with counter examples. But the exp file only has the error info for ">=" and "<". That is why I commented out the case with "==".

PR #658, opened 2022-11-08
[Prover] Added support for new primitive types to arithmetic ops

awelc:
## Motivation

There were some missing bits in the bytecode translator and in the Boogie prelude file that this PR is attempting to add.

Yes


oxade:
Thanks for handling this @awelc 
I initially had these in my first PR but backed them out because I wasn't sure how to fully reason about the prover.

emmazzz (approved):
LGTM! 
Maybe we can add a test case like this? https://github.com/move-language/move/blob/main/language/move-prover/tests/sources/functional/cast.move

PR #659, opened 2022-11-09
[move-prover] Add support for multiple types in vectors

### LINKED ISSUE #653: [Prover] Not yet supported constant vector type: Vector(Primitive(Bool))
# 🐛 Bug

The Move Prover not yet supports constant bool vectors.

Happy to dig into it myself but tagging @rahxephon89 as he expressed interest in providing feedback.

## To reproduce

Run prover on the following module:
```
module bool_vec::bool_vec {
    fun foo(): vector<bool> {
        let v = vector[];
        v
    }
}
```

**Stack trace/error message**

```
thread 'main' panicked at 'not implemented: Not yet supported constant vector type: Vector(Primitive(Bool))', language/move-prover/bytecode/src/stackless_bytecode_generator.rs:1348:22
```

## Expected Behavior

This piece of code should verify correctly.

## System information

Main branch of the Move repo (sha: 3ac0cd7b78cc2a6d470b7fefbfd82b0c044ef576) built on a Mac.

rahxephon89:
## Motivation

Try to resolve #653 in a general way as suggested in #411. Since it is a little bit urgent, this PR is quite rough and a lot things are left unimplemented. Any comments and suggestions are welcomed. 

Yes

## Test Plan

cargo test

### DIFF
--- a/language/move-model/src/ast.rs
+++ b/language/move-model/src/ast.rs
@@ -960,7 +960,8 @@ pub enum Value {
     Number(BigInt),
     Bool(bool),
     ByteArray(Vec<u8>),
-    AddressArray(Vec<BigUint>),
+    AddressArray(Vec<BigUint>), // TODO: merge AddressArray to Vector type in the future
+    Vector(Vec<Value>),
 }
 
 impl fmt::Display for Value {
@@ -972,6 +973,7 @@ impl fmt::Display for Value {
             // TODO(tzakian): Figure out a better story for byte array displays
             Value::ByteArray(bytes) => write!(f, "{:?}", bytes),
             Value::AddressArray(array) => write!(f, "{:?}", array),
+            Value::Vector(array) => write!(f, "{:?}", array),
         }
     }
 }
--- a/language/move-prover/boogie-backend/src/boogie_helpers.rs
+++ b/language/move-prover/boogie-backend/src/boogie_helpers.rs
@@ -9,7 +9,7 @@ use num::BigUint;
 
 use move_binary_format::file_format::TypeParameterIndex;
 use move_model::{
-    ast::{MemoryLabel, TempIndex},
+    ast::{MemoryLabel, TempIndex, Value},
     model::{
         FieldEnv, FunctionEnv, GlobalEnv, ModuleEnv, QualifiedInstId, SpecFunId, StructEnv,
         StructId, SCRIPT_MODULE_NAME,
@@ -18,7 +18,7 @@ use move_model::{
     symbol::Symbol,
     ty::{PrimitiveType, Type},
 };
-use move_stackless_bytecode::function_target::FunctionTarget;
+use move_stackless_bytecode::{function_target::FunctionTarget, stackless_bytecode::Constant};
 
 use crate::options::BoogieOptions;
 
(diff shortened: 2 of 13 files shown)


language/move-prover/bytecode/src/stackless_bytecode_generator.rs
```diff
@@ -1345,7 +1345,13 @@ impl<'a> StacklessBytecodeGenerator<'a> {
                         .collect::<Vec<BigUint>>();
                     Constant::AddressArray(b)
                 }
-                _ => unimplemented!("Not yet supported constant vector type: {:?}", ty),
+                _ => {
```
awelc, line 1348: Do we want to keep `Type::Primitive(PrimitiveType::Address)` arm above or fold it into the more generic handling of constant vectors?

awelc:
@rahxephon89 - thank you for getting to this so quickly!!!

language/move-prover/bytecode/src/stackless_bytecode_generator.rs
rahxephon89: I would like to keep it for now in case my PR may break existing working code :)

language/move-model/src/ast.rs
```diff
@@ -961,6 +961,7 @@ pub enum Value {
     Bool(bool),
     ByteArray(Vec<u8>),
     AddressArray(Vec<BigUint>),
```
meng-xu-cs, line 963: Maybe leave a TODO item here to indicate that the `AddressArray` may be merged into `Vector` in the future?

rahxephon89: done

language/move-prover/boogie-backend/src/boogie_helpers.rs
```diff
@@ -329,6 +329,24 @@ pub fn boogie_address_blob(_options: &BoogieOptions, val: &[BigUint]) -> String
     }
 }
 
+pub fn boogie_constant_blob(_options: &BoogieOptions, val: &[Constant]) -> String {
+    let args = val.iter().map(|v| format!("{}", *v)).collect_vec();
```
meng-xu-cs, line 333: This doesn't seem quite right to me.., I don't think any `Constant` can be `format!()` into boogie-recognizable text.

Perhaps a better solution is to have a helper function:
```rust
pub fn boogie_constant(options: &BoogieOptions, val: &Constant) -> String { .. }
```

And then call this `boogie_constant` inside the `.map()`

language/move-prover/boogie-backend/src/boogie_helpers.rs
```diff
@@ -329,6 +329,48 @@ pub fn boogie_address_blob(_options: &BoogieOptions, val: &[BigUint]) -> String
...
+                .map(|v| boogie_constant(_options, v))
+                .collect_vec(),
+        ),
+        Constant::U16(num) => num.to_string(),
+        Constant::U32(num) => num.to_string(),
+    }
+}
+
+pub fn boogie_value_blob(_options: &BoogieOptions, val: &[Value]) -> String {
+    let args = val.iter().map(|v| format!("{}", *v)).collect_vec();
```
meng-xu-cs, line 366: I think this `ast::Value` might also needs a helper function in boogie to replace the `format!("{}", *v)`

meng-xu-cs (commented):
Just one more issue to address, all else look good to me.

PR #661, opened 2022-11-09
[move-prover] Fix the bug on handling recursive functions

### LINKED ISSUE #648: [Bug] Move Prover panicks during borrow analysis
# 🐛 Bug
The example Move project in [merge_sort.zip](https://github.com/move-language/move/files/9945968/merge_sort.zip) implements merge sort on `vector<u64>`. It compiles without issues (using move commit  3174338) and a simple test succeeds. The code does not contain any specifications apart from a `pragma opaque` which is required for a recursive function.
Still, the Move Prover panicks during model construction:
image

rahxephon89:
## Motivation

This PR tries to close #648 by avoiding panics in borrow_analysis and composition_analysis when handling recursive functions.

Yes

## Test Plan

cargo test

### DIFF
--- a/language/move-prover/bytecode/src/compositional_analysis.rs
+++ b/language/move-prover/bytecode/src/compositional_analysis.rs
@@ -38,11 +38,17 @@ impl<'a> SummaryCache<'a> {
                 if fun_env.is_native_or_intrinsic() {
                     None
                 } else {
-                    Some(
-                        fun_data.annotations.get::<Summary>().expect(
+                    // This avoids crashing when handling recursive or mutually recursive functions
+                    // TODO: add logic to support adding summary to recursive functions
+                    if let Some(summary) = fun_data.annotations.get::<Summary>() {
+                        Some(summary)
+                    } else {
+                        self.global_env.error(
+                            &fun_env.get_loc(),
                             "Failed to resolve summary; recursion or scheduler bug suspected",
-                        ),
-                    )
+                        );
+                        None
+                    }
                 }
             })
     }
--- a/language/move-prover/bytecode/src/function_target_pipeline.rs
+++ b/language/move-prover/bytecode/src/function_target_pipeline.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use core::fmt;
-use std::{collections::BTreeMap, fmt::Formatter, fs};
+use std::{cmp::Ordering, collections::BTreeMap, fmt::Formatter, fs};
 
 use itertools::Itertools;
 use log::{debug, info};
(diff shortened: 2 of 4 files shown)


language/move-prover/bytecode/src/compositional_analysis.rs
```diff
@@ -38,11 +38,7 @@ impl<'a> SummaryCache<'a> {
                 if fun_env.is_native_or_intrinsic() {
                     None
                 } else {
-                    Some(
-                        fun_data.annotations.get::<Summary>().expect(
-                            "Failed to resolve summary; recursion or scheduler bug suspected",
-                        ),
-                    )
+                    fun_data.annotations.get::<Summary>()
```
sblackshear, line 41: Would be good to add a TODO here--this will avoid the crash but (I believe) will still not allow us to create summaries for recursive or mutually recursive code.

language/move-prover/bytecode/src/function_target_pipeline.rs
```diff
@@ -313,7 +313,18 @@ impl FunctionTargetPipeline {
         let mut dep_ordered = vec![];
         while !worklist.is_empty() {
             // Put functions with 0 calls first in line, at the end of the vector
-            worklist.sort_by(|(_, callees1), (_, callees2)| callees2.len().cmp(&callees1.len()));
+            worklist.sort_by(|(_, callees1), (caller2, callees2)| {
+                if Ordering::Equal == callees2.len().cmp(&callees1.len()) {
```
sblackshear, line 317: Comment here explaining  what this is doing might be useful

sblackshear (commented):
Thanks for fixing this!

language/move-prover/bytecode/src/compositional_analysis.rs
rahxephon89: done

language/move-prover/bytecode/src/function_target_pipeline.rs
rahxephon89: done

meng-xu-cs:
More of a high-level question.

If issue #648 is about `panic` being an inappropriate behavior, we could change it into an error which exit gracefully instead of a hard panic. Personal I would prefer this approach.

As far as I know, there are several passes in the bytecode transformation pipeline that do not support recursion (borrow analysis being one for example). Each pass needs customized logic to run its analysis *in the presence of recursive calls*. Currently the way how borrow analysis handles recursive calls (i.e., by ignoring the callsite essentially) is not entirely correct in my understanding, so I would hesitate a bit in landing this part of the code.

rahxephon89:
> More of a high-level question.
> 
> If issue #648 is about `panic` being an inappropriate behavior, we could change it into an error which exits gracefully instead of a hard panic. Personally I would prefer this approach.
> 
> As far as I know, there are several passes in the bytecode transformation pipeline that do not support recursion (borrow analysis being one for example). Each pass needs customized logic to run its analysis _in the presence of recursive calls_. Currently the way how borrow analysis handles recursive calls (i.e., by ignoring the callsite essentially) is not entirely correct in my understanding, so I would hesitate a bit in landing this part of the code.

Thanks for the advice. I have added to code to raise errors.

wrwg (approved):
Thanks for adding the error message. As a rule of thumb, any restrictions of the input to the prover should result in an error messages with source context. We should neither panic nor silently accept.

PR #662, opened 2022-11-09
[move-package] Don't update tag based git dependencies

gregnazario:
## Motivation

Tag based git dependencies can't really be updated easily after the initial download, so these will not be updated.  Tags are usually used as releases, and in general shouldn't change after the fact.  This becomes tricky in general in comparison to a branch based flow.  Additionally, added an updating message and cleaned up some duplicate outputs.

Resolves: https://github.com/move-language/move/issues/608

Yes

## Test Plan

I tested this manually with a package that wasn't working correctly with this https://github.com/pontem-network/liquidswap

Now, it doesn't crash, it continues on.

Additionally, I tested with a branch based lookup without issue as well as revision based.


sblackshear (approved):
Thanks for this fix Greg!

PR #663, opened 2022-11-10
[linker][new integer types] Fix missing case in dependency verifier

tnowacki:
- Fix missing case in dependency verifier for equality on signature tokens (from different contexts)
- Fix prop test case

## Motivation

- fixing bug in dependency verifier

## Test Plan

- new test

### DIFF
--- a/language/move-bytecode-verifier/src/dependencies.rs
+++ b/language/move-bytecode-verifier/src/dependencies.rs
@@ -444,8 +444,11 @@ fn compare_types(
     match (handle_type, def_type) {
         (SignatureToken::Bool, SignatureToken::Bool)
         | (SignatureToken::U8, SignatureToken::U8)
+        | (SignatureToken::U16, SignatureToken::U16)
+        | (SignatureToken::U32, SignatureToken::U32)
         | (SignatureToken::U64, SignatureToken::U64)
         | (SignatureToken::U128, SignatureToken::U128)
+        | (SignatureToken::U256, SignatureToken::U256)
         | (SignatureToken::Address, SignatureToken::Address)
         | (SignatureToken::Signer, SignatureToken::Signer) => Ok(()),
         (SignatureToken::Vector(ty1), SignatureToken::Vector(ty2)) => {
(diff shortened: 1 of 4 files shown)


oxade:
@tnowacki thanks for catching this.
There's just so many places that these types touch and I was wondering is there a proper way to do an audit to make sure no more missing enums?

language/move-bytecode-verifier/src/dependencies.rs
```diff
@@ -472,7 +475,21 @@ fn compare_types(
                 Ok(())
             }
         }
-        _ => Err(PartialVMError::new(StatusCode::TYPE_MISMATCH)),
+        (SignatureToken::Bool, _)
+        | (SignatureToken::U8, _)
+        | (SignatureToken::U64, _)
```
oxade, line 480: Thanks for expanding these out so future changes must be explicitly handled.

oxade (approved):
Thanks!

language/move-bytecode-verifier/transactional-tests/tests/dependencies/call_integers_valid.move
```diff
@@ -0,0 +1,22 @@
+//# publish
+module 0x42::m {
+public fun plus8(a: u8): u8 { a + 1 }
+public fun plus16(a: u16): u16 { a + 1 }
+public fun plus32(a: u32): u32 { a + 1 }
+public fun plus64(a: u64): u64 { a + 1 }
+public fun plus128(a: u128): u128 { a + 1 }
+public fun plus256(a: u256): u256 { a + 1 }
```
jiangying000, line 8: formatting issue here

tnowacki: It's just a test :) This is setup this way to make it easy to see each case

While the lack of newline after `{` is a bit obtuse. I've come to like not indenting inside of the `module` block. Not really a point in doing so IMO

tnowacki:
> @tnowacki thanks for catching this. There's just so many places that these types touch and I was wondering is there a proper way to do an audit to make sure no more missing enums?

I don't think there is, sadly. Just writing tests :) This one is a bit subtle and not surprising we forgot it in the first PR. 

I did another pass, hence the fix for proptests.

sblackshear (approved):

PR #668, opened 2022-11-13
[vm] Numerous fixes

wrwg:
Contains a number of reliability fixes. Adds a configuration for the bytecode verifier to limit the size of type instantiations and basic blocks.


davidiw (approved):

language/move-bytecode-verifier/src/verifier.rs
```diff
@@ -19,6 +20,10 @@ use move_binary_format::{
 #[derive(Debug, Clone, Default)]
 pub struct VerifierConfig {
     pub max_loop_depth: Option<usize>,
+    pub treat_friend_as_private: bool,
```
jolestar, line 23: This config seam like do not use.

wrwg: Oh that's then a merge conflict. We should address this in a followup PR.

PR #672, opened 2022-11-14
[move-vm][value-impl] Fix serialization case for vecu16, u32, u256

tnowacki:
- Fixed missing serialization case
- Added tests


amnn (approved):
LGTM

oxade (approved):
Thanks

PR #676, opened 2022-11-17
[bytecode-verifier] Avoid unwraps were possible when in release mode

wrwg:
This introduces new macros `safe_unpack!` and `safe_assert!` which can be used in contexts which have a `PartialVMResult` as a return value. Those macros panic when debug assertions are enabled, but otherwise generate an internal invariant violation error. They are applied where an obvious substitution is possible.

### DIFF
--- a/language/move-binary-format/src/lib.rs
+++ b/language/move-binary-format/src/lib.rs
@@ -136,3 +136,59 @@ impl fmt::Display for SignatureTokenKind {
         f.write_str(desc)
     }
 }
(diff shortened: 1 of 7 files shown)


runtian-zhou (approved):
Looks awesome to me! I can stamp it but I'll leave for @tnowacki for another review if he wishes!

tnowacki (approved):
Looked at it briefly this morning, looks great! Love it!

PR #677, opened 2022-11-17
[core-types] cherry-pick: Make TypeTag enum use less memory

vgao1996:
Cherry-picking https://github.com/move-language/move/pull/525

> Currently a single TypeTag consumes 96 bytes because of the inlined StructTag (88 bytes), it starts to bloat up with nested TypeTag::Vector(TypeTag).
This commit reduces the size from 96 to 16 bytes, it doesn't fully solve the problem but reduces the impact by 6x.


wrwg (approved):
Wow. I thought that one was already cherry-picked.

Issue #660, opened 2022-11-09
[Prover] Error handling borrow_mut_global-like native functions


awelc:
With different underlying storage models for different Move flavors, one may have to implement their own version of operators accessing global storage. Currently, without baking it into prover itself, this is only possible via native functions, but what is supported/possible right now is not always sufficient. In particular, the borrow checker complains when trying to prove the following module:
```rust
module broken_native::broken_native {

    struct SomeStruct has key, drop {
        val: u64
    }

    native fun just_like_borrow_mut_global<T: key>(a: address): &mut T;

    fun foo(a: address) {
        let r = just_like_borrow_mut_global<SomeStruct>(a);
        r.val = 42;
    }
}
```

The error generated looks as follows:
```
thread 'main' panicked at 'inconsistent borrow information: undefined output: broken_native::just_like_borrow_mut_global
', language/move-prover/bytecode/src/borrow_analysis.rs:324:17
```

I am opening this issue to start a discussion on how this could be supported in the prover to facilitate different storage models. @emmazzz made an observation that the native function in question attempts to do something similar to what the `borrow_global_mut` operation does and perhaps the same mechanism that's used to handle the latter could also be used to handle the former (@wrwg, @stephenreborn, @junkil-park?)

meng-xu-cs:
The way I am thinking is that we can use the `pragma intrinsic` scheme to annotate that this `just_like_borrow_mut_global` funtion is essentially an intrinsic function which behaves just like `borrow_mut_global`.

Syntatically:

```
native fun just_like_borrow_mut_global<T: key>(a: address): &mut T;

spec just_like_borrow_mut_global {
    pragma intrinsic = borrow_global_mut;
}
```

We already use the similar approach to support variable implementation of the `Map` type, as shown in the `Table.spec.move` file.
```
spec Table {
    pragma intrinsic = map;
}
```

awelc:
I was erring on the side of caution and wanted to be able to uniquely identify a function in question to avoid a situation where somewhere in someone's code we'd have a function with the intrinsic pragma and the same name.

Also, I am not sure how to easily (or easier than what I implemented) to check if a given function has intrinsic pragma. We have `is_intrinsic_of` but that's for structs rather than for functions. I am sure it could be added but I wasn't sure if such addition would be warranted by this single use case.

meng-xu-cs:
> avoid a situation where somewhere in someone's code we'd have a function with the intrinsic pragma and the same name.

[...]

PR #679, opened 2022-11-18
[Prover] Generalize handling of native methods in borrow checker

awelc:
## Motivation

The goal of these changes is to support other Move language flavors implementing alternative storage models (it resolves https://github.com/move-language/move/issues/660). The main idea is that alternative storage models may need support for implementing custom mutable borrows along the same lines as "core" Move's `borrow_mut` for vectors or tables works:
```
native public fun borrow_mut<Element>(v: &mut vector<Element>, i: u64): &mut Element;

public fun borrow_mut<K: copy + drop, V>(table: &mut Table<K, V>, key: K): &mut V {
   &mut borrow_box_mut<K, V, Box<V>>(table, key).val
}
```

These functions get a special treatment from the prover to artificially create an edge from their first parameter to their return value. Additionally, Operations on the values being borrowed from have to be explicitly defined in Boogie and referred to during bytecode translation.

This PR extends this mechanism to allow native functions defined outside of "core" Move to take advantage of the same custom support. 

Yes

### DIFF
--- a/language/move-model/src/model.rs
+++ b/language/move-model/src/model.rs
@@ -1183,7 +1183,7 @@ impl GlobalEnv {
             def_idx,
             handle_idx,
             arg_names,
-            type_arg_names,
+            type_args: type_arg_names,
             spec,
             called_funs: Default::default(),
             calling_funs: Default::default(),
@@ -2932,9 +2932,9 @@ pub struct FunctionData {
     /// List of function argument names. Not in bytecode but obtained from AST.
     arg_names: Vec<Symbol>,
 
-    /// List of type argument names. Not in bytecode but obtained from AST.
+    /// List of type arguments. Not in bytecode but obtained from AST.
     #[allow(unused)]
-    type_arg_names: Vec<(Symbol, Type)>,
+    type_args: Vec<(Symbol, Type)>,
 
     /// Specification associated with this function.
     spec: Spec,
@@ -2962,7 +2962,7 @@ impl FunctionData {
             def_idx,
             handle_idx,
             arg_names: vec![],
-            type_arg_names: vec![],
+            type_args: vec![],
             spec: Spec::default(),
             called_funs: Default::default(),
             calling_funs: Default::default(),
@@ -3300,7 +3300,7 @@ impl<'env> FunctionEnv<'env> {
 
     /// Returns types of the type parameters.
     pub fn get_type_parameter_types(&self) -> Vec<Type> {
-        project_2nd(&self.data.type_arg_names)
+        project_2nd(&self.data.type_args)
     }
 
     /// Returns the type parameters associated with this function.
--- a/language/move-model/src/model.rs
+++ b/language/move-model/src/model.rs
@@ -71,7 +71,6 @@ use crate::{
         DELEGATE_INVARIANTS_TO_CALLER_PRAGMA, DISABLE_INVARIANTS_IN_BODY_PRAGMA, FRIEND_PRAGMA,
         INTRINSIC_PRAGMA, OPAQUE_PRAGMA, VERIFY_PRAGMA,
     },
-    project_2nd,
     symbol::{Symbol, SymbolPool},
     ty::{PrimitiveType, Type, TypeDisplayContext, TypeUnificationAdapter, Variance},
 };
@@ -1172,7 +1171,7 @@ impl GlobalEnv {
         loc: Loc,
         attributes: Vec<Attribute>,
         arg_names: Vec<Symbol>,
-        type_arg_names: Vec<(Symbol, Type)>,
+        type_arg_names: Vec<Symbol>,
         spec: Spec,
     ) -> FunctionData {
         let handle_idx = module.function_def_at(def_idx).function;
@@ -1183,7 +1182,7 @@ impl GlobalEnv {
             def_idx,
             handle_idx,
             arg_names,
-            type_args: type_arg_names,
+            type_arg_names,
             spec,
             called_funs: Default::default(),
             calling_funs: Default::default(),
@@ -2932,9 +2931,9 @@ pub struct FunctionData {
     /// List of function argument names. Not in bytecode but obtained from AST.
     arg_names: Vec<Symbol>,
 
-    /// List of type arguments. Not in bytecode but obtained from AST.
+    /// List of type argument names. Not in bytecode but obtained from AST.
     #[allow(unused)]
-    type_args: Vec<(Symbol, Type)>,
+    type_arg_names: Vec<Symbol>,
 
     /// Specification associated with this function.
     spec: Spec,
@@ -2962,7 +2961,7 @@ impl FunctionData {
             def_idx,
             handle_idx,
             arg_names: vec![],
-            type_args: vec![],
+            type_arg_names: vec![],
             spec: Spec::default(),
             called_funs: Default::default(),
             calling_funs: Default::default(),
@@ -3296,11 +3295,6 @@ impl<'env> FunctionEnv<'env> {
             return friend_env.get_transitive_friend();
         }
         self.clone()
-    }
-
-    /// Returns types of the type parameters.
-    pub fn get_type_parameter_types(&self) -> Vec<Type> {
-        project_2nd(&self.data.type_args)
     }
 
     /// Returns the type parameters associated with this function.
--- a/language/move-model/src/model.rs
+++ b/language/move-model/src/model.rs
@@ -71,6 +71,7 @@ use crate::{
         DELEGATE_INVARIANTS_TO_CALLER_PRAGMA, DISABLE_INVARIANTS_IN_BODY_PRAGMA, FRIEND_PRAGMA,
         INTRINSIC_PRAGMA, OPAQUE_PRAGMA, VERIFY_PRAGMA,
     },
+    project_2nd,
     symbol::{Symbol, SymbolPool},
     ty::{PrimitiveType, Type, TypeDisplayContext, TypeUnificationAdapter, Variance},
 };
@@ -1171,7 +1172,7 @@ impl GlobalEnv {
         loc: Loc,
         attributes: Vec<Attribute>,
         arg_names: Vec<Symbol>,
-        type_arg_names: Vec<Symbol>,
+        type_arg_names: Vec<(Symbol, Type)>,
         spec: Spec,
     ) -> FunctionData {
         let handle_idx = module.function_def_at(def_idx).function;
@@ -2933,7 +2934,7 @@ pub struct FunctionData {
 
     /// List of type argument names. Not in bytecode but obtained from AST.
     #[allow(unused)]
-    type_arg_names: Vec<Symbol>,
+    type_arg_names: Vec<(Symbol, Type)>,
 
     /// Specification associated with this function.
     spec: Spec,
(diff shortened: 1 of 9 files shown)


language/move-prover/bytecode/src/borrow_analysis.rs
```diff
@@ -419,15 +426,59 @@ impl FunctionTargetProcessor for BorrowAnalysisProcessor {
...
+                )
+            } else if fun_env.is_intrinsic_of(INTRINSIC_FUN_MAP_BORROW_MUT) {
+                (
+                    "GetTable".to_string(),
+                    "UpdateTable".to_string(),
+                    type_args[1].clone(),
+                )
+            } else {
+                let (r, u, i) = borrow_native_info.unwrap();
+                (r, u, type_args[i].clone())
```
emmazzz, line 479: Very nice!

language/move-model/src/model.rs
```diff
@@ -2933,7 +2934,7 @@ pub struct FunctionData {
 
     /// List of type argument names. Not in bytecode but obtained from AST.
     #[allow(unused)]
-    type_arg_names: Vec<Symbol>,
+    type_arg_names: Vec<(Symbol, Type)>,
```
emmazzz, line 2937: Maybe this field should be renamed since it's no longer just names of the type args?

language/move-prover/bytecode/src/options.rs
```diff
@@ -29,6 +29,45 @@ impl AutoTraceLevel {
...
+    /// Method's module name
+    pub mod_name: String,
+    /// Method's name
+    pub name: String,
+    /// Name of a read operation for a given type (in Boogie)
+    pub read_op: String,
+    /// Name of a Write operation for a given type (in Boogie)
+    pub write_op: String,
+    /// Method's type parameter index describing a give type
+    pub tparam_idx: usize,
```
emmazzz, line 48: This seems to assume that the borrow type is one of the type parameters? What if the function has no type parameter or if we have type params `<K, V>` but the borrow type is `Wrapper<V>`? 
I'm fine with the way it is currently too if you think allowing those scenarios would be overgeneralization.

awelc: I agree with your assessment of the limitation, but I am not sure how much it makes sense to generalize it at this point. I'd think that the current solution is already more general than the previous one which assumed that the required type is a type parameter of the argument (e.g., `Table<K,V>`) and it covers all the cases we need to cover right now.

If someone has a suggestion on how/why to generalize it further I am all ears, but perhaps we should consider it only if further need arises?

language/move-model/src/model.rs
awelc: Good call! Done!

language/move-prover/bytecode/src/options.rs
awelc: After pondering this for a bit longer I realized that we probably don't need the type info at all as it was used only to figure out which "access" functions to use in the bytecode translator.

language/move-prover/bytecode/src/stackless_bytecode.rs
```diff
@@ -298,7 +298,7 @@ pub enum BorrowEdge {
     /// Field borrow with static offset.
     Field(QualifiedInstId<StructId>, usize),
     /// Vector borrow with dynamic index.
-    Index,
+    Index((String, String)),
```
meng-xu-cs, line 301: I would recommend against bringing in things that are only know to Boogie here. Instead try to capture the borrow semantics with a higher-level of abstraction (I am proving a solution below as well).

The reason we have `Index` here as a `BorrowEdge` is to mark a conceptual linkage with the SMT array theory. If you take a closer examination of the two things that are currently being modeled as `Index`, they are `Vector` borrows and `Table` borrows. Both resembles essentially the SMT array theory of `K -> V` mapping (yes, vector is just a table with `K = int`).

So, if you are re-purposing the `Index` one with an extended and custom semantics, I am really really really curious about what are the new type of mapping you try to model here.

Based on what I have learned from our prior discussions, here is what I think you are trying to do: you are tying to model a `K -> Any` mapping to simulate heterogeneous maps that can grow dynamically. And perhaps, the `K` can be further narrowed down to `int` or `str` depending on how you mark the field identifier.

This is indeed a new functionality and justifies this new design which is beautifully implemented in this PR (but note: this is a different issue from #660). In fact, I believe that this `Heterogeneous Map` feature is a generic feature that warrants more than just a custom Boogie treatment.

The solution I am proposing to remove this `(String, String)` part in `Index(..)` is to use an enum:

```rust
pub enum ContainerKind {
    /// SMT array theory: int -> E (E is known)
    /// --> already have
    Vector,
    /// SMT array theory: K -> V (both K, V are known)
    /// --> already have
    KVMap,
    /// SMT array theory: K -> Any (K is known, Any can be any type, i.e., dynamic typing)
    /// --> I guess this is the new type of map you are envisioning
    AnyMap: 
}
```

And hence, `BorrowMap::Index(ContainerKind)`.

In `boogie-backend/src/bytecode_translator.rs`, we can further match this `ContainerKind` to roll-out the correct Boogie functions to for reference reading and updating.

However, this `Heterogeneous Map` thing is just a long shot guess and please correct me if this is wrong.

emmazzz: I believe the two `String`s in `Index((String, String))` are for the boogie function names of the custom read_aggregate and update_aggregate. It is populated at `native_annotation` in borrow_analysis.rs. So for vector index, these two strings will be "ReadVec" and "UpdateVec". For custom indices, these two strings are provided through the `BorrowNative` option. 

The first paragraph of your comment still makes sense though. Where do you think is a better place to put these custom function names? @meng-xu-cs 

Using an enum might be a good idea too if the kind of mapping we'll be doing is vastly different from what we currently have. This part hasn't really been designed yet afaik. Right now we are just trying to get Sui Move prover translation (not trying to prove any specs yet) working without errors.

emmazzz: Chatted with @meng-xu-cs earlier this week about this. IIUC the reason why we currently special case for table and vector modules is because they correspond to the two Boogie constructs we have. If the customization we want to add can be represented or modeled using the current constructs, we don't need to provide new aggregate functions. Instead, we should provide custom mapping with information specifying "this custom type should use table Boogie construct". @awelc do you think this is doable?

awelc: > we should provide custom mapping with information specifying "this custom type should use table Boogie construct". @awelc do you think this is doable?

I understand the argument better now, and I agree that this would be ideal, but I don't think it's possible (happy to be proven wrong, though!). The way I currently understand it (though admittedly my understanding is limited), table accesses are modeled by having per-table arrays storing relevant information:
```
// v is the SMT array holding the key-value assignment. e is an array which
// independently determines whether a key is valid or not. l is the length.
//
// Note that even though the program cannot reflect over existence of a key,
// we want the specification to be able to do this, so it can express
// verification conditions like "key has been inserted".
function {:constructor} Table<K, V>(v: [K]V, e: [K]bool, l: int): Table K V;
```

These are used in the aggregate functions to model the read/write operations:

```
function {:inline} GetTable<K,V>(t: Table K V, k: K): V {
    v#Table(t)[k]
}

function {:inline} UpdateTable<K,V>(t: Table K V, k: K, v: V): Table K V {
    Table(v#Table(t)[k := v], e#Table(t)[k := true], l#Table(t))
}
```

The extension proposed in this PR aims at supporting a different scenario. We need to model accessing a piece of data (more specifically a Sui object) that can have an arbitrary type. In fact, the first parameter to the native function that needs modeling is a mutable reference to the unique ID  (`UID`) of an object rather than the object itself:

```
public(friend) native fun borrow_child_object_mut<Child: key>(object: &mut UID, id: address): &mut Child;
```

The aggregate functions modeling read/write operations to an object represented by a given `UID` must tap directly into the (global) mapping representing Sui's storage model and I don't think this can be mapped to the existing array theory, and will require custom aggregate function definitions. Something like the following where we get parent object info from global storage based on its address and then gets its children from the appropriate mapping):

```
function {:inline} GetChild<V>(parent_id: $2_object_UID, child_addr: int): V {
    children#$Memory[uid_to_addr(parent_id)][child_addr]
}
```

meng-xu-cs: @awelc I sincerely ask you to read the comment again.

TLDR

```rust
pub enum ContainerKind {
    /// SMT array theory: int -> E (E is known)
    /// --> already have
    Vector,
    /// SMT array theory: K -> V (both K, V are known)
    /// --> already have
    KVMap,
    /// SMT array theory: K -> Any (K is known, Any can be any type, i.e., dynamic typing)
    /// --> I guess this is the new type of map you are envisioning
    AnyMap: 
}
```

```diff
pub enum BorrowEdge {
    /// Field borrow with static offset.
    Field(QualifiedInstId<StructId>, usize),
    /// Vector borrow with dynamic index.
-   Index,
+   Index(ContainerKind),
}
```

awelc: I did read the comment - my response was to the following comment by @emmazzz:

> If the customization we want to add can be represented or modeled using the current constructs, we don't need to provide new aggregate functions. 

Please also understand that I am doing my best here but I am also learning things as I go that may be super-obvious to everyone else...

I am super cool with introducing the new enum, but I think we still need the custom aggregate functions for `AnyMap`. Would the following (or something similar, like having a new struct as `AnyMap` variant's value instead of two strings representing names of the aggregate functions) make sense?

```
pub enum ContainerKind {
    /// SMT array theory: int -> E (E is known)
    /// --> already have
    Vector,
    /// SMT array theory: K -> V (both K, V are known)
    /// --> already have
    KVMap,
    /// SMT array theory: K -> Any (K is known, Any can be any type, i.e., dynamic typing)
    /// --> I guess this is the new type of map you are envisioning
    AnyMap(String, String)
}
```

I simply don't know how we can map what we need model to one of existing array theories (e.g., those contained in `table-array-theory.bpl ` or `vector-array-theory.bpl`). I think we need to provide our own model there and refer to it from custom aggregate functions"

meng-xu-cs: >  but I think we still need the custom aggregate functions for AnyMap.

We do. I am not fighting against this one.

I am only fighting against exposing the Boogie encoding as `(String, String)` in the bytecode. The Boogie encoding should only be known by things under `boogie-backend/src/*`.

awelc: Cool, thank you for the clarification! Now that I understand what the issue is, I have rewritten the code to avoid introducing Boogie-level aggregate names at the borrow analysis level. Instead, the `Custom` variant of the index edge encodes the name of the native function which needs special treatment and aggregate names are introduced as options to the backend rather than to the bytecode transformer. Hopefully this is acceptable

Another solution would be to create `Custom` variant without any values, and hardcode the names of aggregate functions for it somewhere in `boogie-backend/src/*` the same way that aggregate function names are hardcoded for vectors and tables?

This would satisfy our current need, but would be more limiting in that we could only have one native function that would get the custom treatment (as we would only have one set of hardcoded aggregate function names). I am hoping  for something more general, where we can have this custom treatment for multiple native functions (much like both table and vector aggregate functions are needed in core Move).

meng-xu-cs (approved):
Thanks for the changes. The new approach (of identifying a native borrow function by its name) look good to me. Accept to unblock and this is a good foundation even if we want future iterations.

*NOTE*: I noticed a stream of commits in this PR so I guess the best way to land is squashing first and then merge.

PR #693, opened 2022-11-23
[compiler] Added info about test mode to compiled functions

awelc:
## Motivation

It is sometimes useful to know if a given compiled function is a test function.

Yes

### DIFF
--- a/language/move-binary-format/src/deserializer.rs
+++ b/language/move-binary-format/src/deserializer.rs
@@ -1386,6 +1386,7 @@ fn load_function_def(cursor: &mut VersionedCursor) -> BinaryLoaderResult<Functio
         function,
         visibility,
         is_entry,
+        is_test: false,
         acquires_global_resources,
         code: code_unit,
     })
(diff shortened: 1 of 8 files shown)


language/move-binary-format/src/deserializer.rs
```diff
@@ -1386,6 +1386,7 @@ fn load_function_def(cursor: &mut VersionedCursor) -> BinaryLoaderResult<Functio
         function,
         visibility,
         is_entry,
+        is_test: false,
```
wrwg, line 1389: Drive-by comment. Wondering why do we special case this? Couldn't there be a function which asks something like `has_attribute(..., &TEST_ATTRIBUTE)`? This is how we are doing it in the move model.

awelc: It has been preliminarily deemed to much of a change as the same info can should be discoverable at the adapter during compilation. Keeping it open as a draft until I can confirm that this is the case, but it will likely never leave the draft form.

Issue #667, opened 2022-11-11
[Move-prover] The prover cannot handle mutual recursion properly


rahxephon89:
The Move prover will generate an error when handling mutually recursive functions as shown below. The issue comes from borrow_analysis where summary info cannot be obtained from callees. 
```
    module 0x42::m {

        public fun foo_1() {
            foo_2();
        }

        fun foo_2() {
            foo_3();
        }

        fun foo_3() {
            foo_1();
        }

}
```

PR #698, opened 2022-11-28
[move-prover] deterministic ordering of worklist with recursive calls

meng-xu-cs:
This commit is the first step in adding a systematic support for
recursive function calls in the Prover.

In this commit, the problem we try to solve is enforcing a deterministic
(and yet, bottom-up) ordering of functions to process.

More formally:

Imagine we construct a call graph `G` where two nodes, `A ---> B` means
`A` calls `B` directly or transitively. By this definition, if we have
both `A ---> B` and `B ---> A` then we know both `A` and `B` belong to
the same recursive call grouh, which forms a strongly-connected
component (SCC) in `G`.

When processing all functions in the bytecode transformation pipeline,
we aim to achieve the following property:

- If `A ---> B` and `B -x-> A` (i.e., `B` does not depends on `A`), we
  are sure that `B` gets processed first before `A`
- If `A -x-> B` and `B -x-> A`, we are sure that `A` and `B` gets
  processed in the order as they are defined in the program (which is
  the order they are added to the `FunctionTargetsHolder`)
- Similarly, if `A ---> B` and `B ---> A` (i.e., they are in an SCC),
  they are processed in the order as they are defined in the program.

This is a deterministic ordering which was previously well enforced with
the `worklist` algorithm, **with the assumption that there is no
recursive calls**.

With the introduction of recursive calls, the worklist algorithm will
start to fail in the following example case:

- `A` has only one callee `B`.
- `B` has two callees, `C` and `D` and `B, C, D` forms an SCC.

The correct handling is `B, C, D` and then `A`, but the `worklist`
algorithm will handle `A` first before `B` (because `A` has a smaller
number of callees).

This commit fixes this problem by augmenting the worklist algorithm
with a graph structure to
1) query reachability information around two nodes and
2) group recursive functions into correspondings SCCs

The worklist algorithm is still needed to determine the processing
order among functions (and SCCs) that do not have dependency
information. It is also used to give a deterministic ordering among
functions in the same SCC. Neither of the ordering is guaranteed by
the topological sorting algorithm in `petgraph`.

## Motivation

More systematic supporting of recursive functions

Yes

## Test Plan

CI

### DIFF
--- a/language/move-prover/bytecode/src/function_target_pipeline.rs
+++ b/language/move-prover/bytecode/src/function_target_pipeline.rs
@@ -3,10 +3,19 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use core::fmt;
-use std::{cmp::Ordering, collections::BTreeMap, fmt::Formatter, fs};
+use std::{
+    cmp::Ordering,
+    collections::{btree_map::Entry as MapEntry, BTreeMap, BTreeSet},
+    fmt::Formatter,
+    fs,
+};
 
-use itertools::Itertools;
+use itertools::{Either, Itertools};
 use log::{debug, info};
+use petgraph::{
+    algo::has_path_connecting,
+    graph::{DiGraph, NodeIndex},
+};
 
 use move_model::model::{FunId, FunctionEnv, GlobalEnv, QualifiedId};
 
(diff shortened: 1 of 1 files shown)


rahxephon89:
A side note, it would be great if the example here: https://github.com/move-language/move/issues/667 could be added as a test case

wrwg (approved):
Nice

PR #704, opened 2022-11-29
[RFC] [bytecode-verifier] Use separate thread for bytecode verifier

runtian-zhou:
## Motivation

Add an option to run the bytecode verifier in a separate thread. Bytecode verifier is one of the most critical piece of code in the Move ecosystem and yet a panic in the bytecode verifier will crash the MoveVM and has a cascading effect over the entire upstream software. This PR introduces a flag to run the bytecode verifier to run on a separate thread. Thus even when a panic is encountered, the runtime should still be able to capture such failure without crashing the node software.

Yes

## Test Plan

TBD. Ran the benchmark suite on Aptos side and observed no significant regression in performance.

### DIFF
--- a/language/move-bytecode-verifier/src/verifier.rs
+++ b/language/move-bytecode-verifier/src/verifier.rs
@@ -11,11 +11,13 @@ use crate::{
     script_signature::no_additional_script_signature_checks, signature::SignatureChecker,
     struct_defs::RecursiveStructDefChecker,
 };
+use crossbeam_utils::thread;
 use move_binary_format::{
     check_bounds::BoundsChecker,
-    errors::{Location, VMResult},
+    errors::{Location, PartialVMError, VMResult},
     file_format::{CompiledModule, CompiledScript},
 };
+use move_core_types::vm_status::StatusCode;
 
 #[derive(Debug, Clone, Default)]
 pub struct VerifierConfig {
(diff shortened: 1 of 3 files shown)


language/move-bytecode-verifier/src/verifier.rs
```diff
@@ -83,3 +92,37 @@ pub fn verify_script_with_config(config: &VerifierConfig, script: &CompiledScrip
     CodeUnitVerifier::verify_script(config, script)?;
     script_signature::verify_script(script, no_additional_script_signature_checks)
 }
+
+pub fn verify_script_with_config(config: &VerifierConfig, script: &CompiledScript) -> VMResult<()> {
+    if !config.use_separate_thread {
+        return verify_script_with_config_impl(config, script);
+    }
+    thread::scope(|s| {
```
wrwg, line 100: Even though the process doesn't abort when a thread panics, this does not make this safe.

runtian-zhou:
Sorry if it creates more confusion. This is just an idea I have in mind to avoid bytecode verifier crashing the VM and feel free to close it if you have more concerns.

language/move-bytecode-verifier/src/verifier.rs
runtian-zhou: I suppose you are concerning memory leaks?

runtian-zhou:
Closing this because thread panics might have unintended consequences due to rust's unwind mechanism.

PR #710, opened 2022-12-01
[vm] Some reliability fixes (cherry-pick #703)

wrwg:
* Add checks for max stack usage during stack balance analysis

* avoid duplicating name_def_map

* Type node bounds

This adds two measurements to limit the number of type nodes:

- in the bytecode verifier, module `limits, we restrict the number of nodes in a `SignatureToken` to 256 (configurable). Since at this point we neither know type instantiations nor struct definitions, we choose some weight factor for structs and type parameters. This measurement is mainly there to prevent and algorithms running out of time or memory which are part of the bytecode verifier.
- in the VM loader, we now not only count the depth of a type term but also the number of nodes in computation of type layouts. This includes type instantiation and struct definitions.

Co-authored-by: Zhou Runtian <runtian@aptoslabs.com>
Co-authored-by: Zekun Li <li.zekun@gmail.com>

### DIFF
--- a/language/move-bytecode-verifier/bytecode-verifier-tests/src/unit_tests/limit_tests.rs
+++ b/language/move-bytecode-verifier/bytecode-verifier-tests/src/unit_tests/limit_tests.rs
@@ -2,8 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use move_binary_format::file_format::*;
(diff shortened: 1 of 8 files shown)


jolestar (approved):

vgao1996 (approved):
LGTM but can you fix the build/tests? Thanks!

PR #713, opened 2022-12-02
[Move-prover] Add support for bitwise operation in the prover

rahxephon89:
## Motivation

This PR adds preliminary supports for bitwise operations: `&`, `|` and `^` in the Move prover. Due to the limitation of the type inference, users need to explicitly specify which function arguments or struct fields will be used in bitwise computations. Here is an simple example:

```
    public fun foo_generic<T>(i: T): T {
      i
    }

    spec foo_generic {
     // The first parameter will be of bv type if T is instantiated as a number type
      pragma bv=b"0"; 
     // The first return value will be of bv type if T is instantiated as a number type
      pragma bv_ret=b"0"; 
    }

    public fun test(i: u8): u8 {
      let x1 = foo_generic(i);
      x1 ^ x1
    }

    spec test {
      // Explicit type cast is mandatory for generating correct boogie program
      ensures result == (0 as u8); 
    }
```
The example for structs can be found in `bitwise_operator.move`. 

Currently, values in non-constant vectors can be involved in bitwise operations. However, index of the vector cannot be involved in bitwise operations directly or indirectly. As a result, for the example below:

```
    fun contains(features: &vector<u8>, feature: u64): bool {
        let byte_index = feature / 8;
        let bit_mask = 1 << ((feature % 8) as u8);
        byte_index < vector::length(features) && (*vector::borrow(features, byte_index) & bit_mask) != 0
    }
```

Since `feature` is involved in both arithmetic and bitwise operation (indirectly by `bit_mask`) and `byte_index` is used as the index to access `features`, we **cannot** directly specify the property like this:

```
    spec fun is_ith_bit_set(x: u8, i: u8): bool {
        x & (1 << i) > 0
    }

    spec contains {
        aborts_if false;
        let byte_index = feature / 8;
        ensures result == (byte_index < len(features)) &&
          is_ith_bit_set(features[byte_index], ((feature % 8) as u8));
    }
```
First, the function needs to be rewritten as below:

```
    fun contains(features: &vector<u8>, feature: u64, feature_bv: u64): bool {
        let byte_index = feature / 8;
        let bit_mask = 1 << ((feature_bv % 8) as u8);
        byte_index < vector::length(features) && (*vector::borrow(features, byte_index) & bit_mask) != 0
    }
```
Then, `spec fun` cannot contain bitwise operations. As a result,  we need to write the whole formula in the spec and all expressions that involve the bitwise operations need to be explicitly annotated with types by using the `as ..` operator: 

```
    spec contains {
        pragma bv=b"0,2";
        aborts_if false;
        ensures result == ((feature / 8) < vector::length(features) && 
(((((1 as u8) << ((feature_bv % (8 as u64)) as u64))) as u8) 
           & vector::borrow(features, feature / 8) as u8) > (0 as u8));
    }
```

The upcoming PRs will lift these limitations and allow bitwise operations in general spec functions, tables, choose function, quantified formula and other expression syntax. Smarter type inference analysis will also be added. 

Yes

## Test Plan

cargo test --release


language/move-model/src/number_operation.rs
```diff
@@ -0,0 +1,183 @@
...
+    pragmas::{BV_PARAM_PROP, BV_RET_PROP},
+    ty::Type,
+};
+use itertools::Itertools;
+use std::{collections::BTreeMap, str};
+
+static PARSING_ERROR: &str = "error happened when parsing the bv pragma";
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Default, Hash)]
+pub enum NumOperation {
```
wrwg, line 21: We should have probably discussed this better before, but I would not have expected this information to end up in the move-model. The model is not specific for the Move Prover. Adding this here is not coherent, since this is rather specific for the prover.

Instead, I would have thought this to become one of the multiple attribute extension points. It could be environment extension, or it could be an annotation.

wrwg (commented):
Before digging deeper, one general comment below about putting this into the move model. Not saying that we change this now, but how difficult would it be to change direction a bit?

language/move-model/src/builder/exp_translator.rs
```diff
@@ -831,7 +831,7 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
             }
             EA::Exp_::Cast(exp, typ) => {
                 let ty = self.translate_type(typ);
-                let exp = self.translate_exp(exp, &ty);
+                let exp = self.translate_exp(exp, expected_type);
```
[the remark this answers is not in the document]
rahxephon89: Thanks for the comment. ty and expected_type should be typed check (and probably the target type of the cast needs to be restricted) but it is not implemented yet. I will add a TODO here.

rahxephon89:
> Before digging deeper, one general comment below about putting this into the move model. Not saying that we change this now, but how difficult would it be to change direction a bit?

It is definitely possible to put this into the move model. I did this the current way because the backend code is easier to understand and change. We could discuss more on how to refactor it.

language/move-model/src/number_operation.rs
rahxephon89: I moved `GlobalNumberOperationState` from `GlobalEnv` to its environment extension. However, in the current implementation, it is hard to move the whole file out of move-model because `ExpInfo` needs to carry `NumOperation`.

language/move-model/src/model.rs
```diff
@@ -1507,11 +1536,25 @@ impl GlobalEnv {
...
+    /// Gets the number operation of the given node.
+    pub fn get_node_num_oper(&self, node_id: NodeId) -> NumOperation {
+        self.get_node_num_oper_opt(node_id)
+            .expect("node number oper defined")
+    }
+
     /// Gets the type of the given node.
     pub fn get_node_type(&self, node_id: NodeId) -> Type {
         self.get_node_type_opt(node_id).expect("node type defined")
     }
```
meng-xu-cs, line 1548: This code snippet makes me wonder why we are not fusing the `NumOperation` into `Type`.

Here is the reasoning:
- In Move Model, every expression has a `NodeId` which is associated with `Type`.
- Now in this PR you are introducing a `NumOperation` and tries to associate some `NodeId` with it
- A natural reaction will be, can we merge `NumOperation` with `Type`? Or, the other way round, whether a number is a bitvec or a math number is a *typing* information, it makes perfect sense to me to merge them. So the question becomes: what is the concern that prevent you from merging them?

In fact, I am looking at the `Type` enum in Move Model and one way I see is the following:

```diff
pub enum PrimitiveType {
    Bool,
    U8,
    U16,
    U32,
    U64,
    U128,
    U256,
    Address,
    Signer,
    // Types only appearing in specifications
    Num,
+   BV8,
+   BV16,
+   BV32,
+   BV64,
+   BV128,
+   BV256,
+   NumOrBV // not sure whether we need this bottom case though
    Range,
    EventStore,
}
```

This seems like a more elegant and compact representation of bitvec, plus you get to re-use most of the utilities we have around `Type`, e.g., type unification etc.

meng-xu-cs (commented):
Commented on whether we really need this `NumOperation` enum and some of the associated logic (see below).

Not necessarily suggesting a change here in this PR, but I think we might want to merge this info in the `Type` enum anyway since whether a Move number is really a bounded mathematical number or bitvector is indeed some form of **typing** information.

language/move-model/src/model.rs
meng-xu-cs: Perhaps I need to add some clarification on the meanings of the types:

```rust
pub enum PrimitiveType {
    Bool,
    U8,    // bounded integer allowed in arithmetic operations only
    U16,   // bounded integer allowed in arithmetic operations only
    U32,   // bounded integer allowed in arithmetic operations only
    U64,   // bounded integer allowed in arithmetic operations only
    U128,  // bounded integer allowed in arithmetic operations only
    U256,  // bounded integer allowed in arithmetic operations only
    Address,
    Signer,
    // Types only appearing in specifications
    Num,   // unbounded integer allowed in arithmetic operations only
+   BV8,   // fixed-width bitvector allowed in bitwise operations only
+   BV16,  // fixed-width bitvector allowed in bitwise operations only
+   BV32,  // fixed-width bitvector allowed in bitwise operations only
+   BV64,  // fixed-width bitvector allowed in bitwise operations only
+   BV128, // fixed-width bitvector allowed in bitwise operations only
+   BV256, // fixed-width bitvector allowed in bitwise operations only
    Range,
    EventStore,
}
```

NOTES:
- bitshift operators (`<<` and `>>`) are permitted as both arithmetic and bitwise operations. As arithmetic operations, bitshift are simply multiplications and divisions (with modulo operation)
- cast operations are permitted as both arithmetic and bitwise operations.
- an expression must be one of the following cases
  - bounded integer (i.e., `U<N>`)
  - unbounded integer (i.e., `Num`)
  - fixed-width bitvector (i.e., `BV<N>`)

rahxephon89: > This code snippet makes me wonder why we are not fusing the `NumOperation` into `Type`.
> 
> Here is the reasoning:
> 
> * In Move Model, every expression has a `NodeId` which is associated with `Type`.
> * Now in this PR you are introducing a `NumOperation` and tries to associate some `NodeId` with it
> * A natural reaction will be, can we merge `NumOperation` with `Type`? Or, the other way round, whether a number is a bitvec or a math number is a _typing_ information, it makes perfect sense to me to merge them. So the question becomes: what is the concern that prevent you from merging them?
> 
> In fact, I am looking at the `Type` enum in Move Model and one way I see is the following:
> 
> ```diff
> pub enum PrimitiveType {
>     Bool,
>     U8,
>     U16,
>     U32,
>     U64,
>     U128,
>     U256,
>     Address,
>     Signer,
>     // Types only appearing in specifications
>     Num,
> +   BV8,
> +   BV16,
> +   BV32,
> +   BV64,
> +   BV128,
> +   BV256,
> +   NumOrBV // not sure whether we need this bottom case though
>     Range,
>     EventStore,
> }
> ```
> 
> This seems like a more elegant and compact representation of bitvec, plus you get to re-use most of the utilities we have around `Type`, e.g., type unification etc.

I agree that by design it is cleaner to merge `NumOperation` into `Type`. The current implementation touches the move-model as little as possible to speed up the implementation without bringing in more bugs. We can discuss more about refactor.

language/move-model/src/builder/exp_translator.rs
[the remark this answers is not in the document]
wrwg: Maybe there is a misunderstanding, but what I was saying is that this change is wrong. You should (a) translate `exp` without a type constraint (b) then use `self.check_type` (see other examples in this function) to check that `expected_type` unifies with `ty`. Then there is a TODO needed somewhere to check what casts are actually allowed, but before that the basic type check here needs to be corrected.

language/move-prover/tests/sources/functional/bitwise_features.move
```diff
@@ -0,0 +1,52 @@
...
+        let byte_index = feature / 8;
+        let bit_mask = 1 << ((feature % 8) as u8);
+        byte_index < vector::length(features) && ((*vector::borrow(features, byte_index))) & bit_mask != 0
+    }
+
+    spec contains {
+        pragma bv=b"0";
+        aborts_if false;
+        ensures result == ((feature / 8) < vector::length(features) && ((int2bv((((1 as u8) << ((feature % (8 as u64)) as u64)) as u8)) as u8)
+            & vector::borrow(features, feature / 8) as u8) > (0 as u8));
```
junkil-park, line 27: `vector::borrow(features, feature / 8)` can be written as `features[feature/8]` in the spec, which is perhaps a bit more readable.

language/move-model/src/builder/exp_translator.rs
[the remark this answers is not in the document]
wrwg: Sorry that is still not correct. Basically casting is broken with this change (granted it was broken before as well).

If one calls `trans(exp, expected_type)` then `exp` must have the given expected type. Now what you are doing is taking the call `trans(exp as ty, expected_type)` and map it to `trans(exp, expected_type)`. So you are requiring `exp` to have the type of the whole expression _after_ the cast.

What you want here is roughly the following:

```
 let ty = self.translate_type(typ);
 // Check that ty is matching expected_type
 self.check_type(
                    &loc,
                    &ty,
                    expected_type,
                    "in cast expression",
                );
 let exp = self.translate_exp_free(exp); // Do not require any specific type during translation
 if !ty.is_number() ... // as below
```

language/move-prover/boogie-backend/src/bytecode_translator.rs
```diff
@@ -1095,13 +1245,182 @@ impl<'env> FunctionTranslator<'env> {
...
+                            let targeted = self.fun_target.module_env().is_target();
+                            let caller_mid = self.fun_target.module_env().get_id();
+                            let caller_fid = self.fun_target.get_id();
+                            let fun_verified =
+                                !self.fun_target.func_env.is_explicitly_not_verified(
+                                    &ProverOptions::get(self.fun_target.global_env()).verify_scope,
                                 );
+                            if dest_str.is_empty() {
+                                // TODO(tengzhang): a hacky way to generate calls for vector functions, need refactor
+                                if module_env
```
wrwg, line 1257: Move this test into model_env, and also check the address?

`module_env.is_std_vector()`

language/move-prover/boogie-backend/src/bytecode_translator.rs
```diff
@@ -1095,13 +1245,182 @@ impl<'env> FunctionTranslator<'env> {
...
+                                let mut bv_flag = if !srcs.is_empty() && targeted && fun_verified {
+                                    *global_state
+                                        .get_temp_index_oper(
+                                            caller_mid,
+                                            caller_fid,
+                                            srcs[0],
+                                            baseline_flag,
+                                        )
+                                        .unwrap()
+                                        == Bitwise
```
wrwg, line 1326: Can't we avoid doing the same thing here again? And can't we factor this out into an independent function? This code in fact looks rather hacky right now.

language/move-prover/boogie-backend/src/bytecode_translator.rs
```diff
@@ -1474,137 +1854,221 @@ impl<'env> FunctionTranslator<'env> {
...
+                                | Type::Reference(_, _)
+                                | Type::Fun(_, _)
+                                | Type::TypeDomain(_)
+                                | Type::ResourceDomain(_, _, _)
+                                | Type::Error
+                                | Type::Var(_) => unreachable!(),
+                            };
+                            let src_type = match &self.get_local_type(op2) {
+                                Type::Primitive(PrimitiveType::U8) => "8",
+                                Type::Primitive(PrimitiveType::U16) => "16",
```
wrwg, line 1964: Factor this out in an independent function and reuse it?

E.g. `fun get_bit_with(ty: &Type) -> String` and then you can use this in `format!("Bv{}", get_bit_width(ty))`

wrwg (commented):
Quite an herculean effort! Can we try to remove some of the code duplication in the bytecode translator and perhaps elsewhere? I'm afraid if don't do now this will stay forever. Once done, I'm happy to stamp, trusting in the right algorithmic choices specifically for the number operation inference.

language/move-model/src/builder/exp_translator.rs
```diff
@@ -831,12 +831,18 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
...
             EA::Exp_::Cast(exp, typ) => {
                 let ty = self.translate_type(typ);
-                let exp = self.translate_exp(exp, &ty);
-                ExpData::Call(
-                    self.new_node_id_with_type_loc(&ty, &loc),
-                    Operation::Cast,
-                    vec![exp.into_exp()],
-                )
+                self.check_type(&loc, &ty, expected_type, "in cast expression");
+                let (_, exp) = self.translate_exp_free(exp);
```
wrwg, line 835:
```
let (exp_ty, exp) = self.translate_exp_free(exp);
if !ty.is_number() || !exp_ty.is_number() ...
```

wrwg (approved):

PR #714, opened 2022-12-03
[compiler] Added info about function attributes as annotations

awelc:
## Motivation

This is another take at adding information about function attributes (specifically to find out if they are meant for testing) to make it available as part of compilation output. The previous take (https://github.com/move-language/move/pull/693) was aiming to add it in the file format, this one adds the required info into the annotated compiled units instead. I followed @wrwg's suggestions from the previous IR and instead of encoding just the fact if the function is involved in tests, this take encodes all function attributes.

Yes

### DIFF
--- a/language/move-compiler/src/compiled_unit.rs
+++ b/language/move-compiler/src/compiled_unit.rs
@@ -5,7 +5,7 @@
 use crate::{
     diag,
     diagnostics::Diagnostics,
-    expansion::ast::{ModuleIdent, ModuleIdent_, SpecId},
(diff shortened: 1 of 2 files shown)


tnowacki (commented):
I think things look good, but any reason this is still a draft? Just checking before I approve

awelc:
> I think things look good, but any reason this is still a draft? Just checking before I approve

No reason other than the my previous initial proposal for this (modifying file format) was also a draft to limit exposure in case this is not quite the direction to take. Since we discussed this approach elsewhere I might have taken it out of draft earlier, but it's done now!

tnowacki (approved):

PR #720, opened 2022-12-05
[move-prover] allow annotations to mark whether a fixedpoint is reached

meng-xu-cs:
This commit allows the `Annotations` struct which resides in
`FunctionData` to mark whether an annotation has reached a fixedpoint
status in the presence of recursive calls.

The core of change is this:

```diff
pub struct Annotations {
-    map: BTreeMap<TypeId, Data>,
+    map: BTreeMap<TypeId, (Data, bool)>,
}
```

Currently all annotations, when set, are marked as already reached
fixedpoint status. This is consistent with the current assumption of
no-recursive calls.

Subsequent commits will relax this unsound assumption by examining the
annotation structure produced by different passes in the transformation
pipeline.

This commit depends on PR #698. A better way to review this PR is
to just look at the second commit: 7ecd599bd95fbd9ea2fb51ae68949d10c303dd57

## Motivation

More systematic supporting of recursive functions

Yes

## Test Plan

CI

### DIFF
--- a/language/move-prover/bytecode/src/annotations.rs
+++ b/language/move-prover/bytecode/src/annotations.rs
@@ -2,7 +2,6 @@
 // Copyright (c) The Move Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-use itertools::Itertools;
 use std::{
     any::{Any, TypeId},
     collections::BTreeMap,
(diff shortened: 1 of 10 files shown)


rahxephon89 (approved):
LGTM!

wrwg (approved):

PR #721, opened 2022-12-05
[u256] minor fixes

vgao1996:
This includes two minor fixes related to u256:
1. It fixes a build error in the fuzzing code
2. It implements Default for u256 so to make it consistent with other primitive integer types. This is also needed by some other downstream tools

### DIFF
--- a/language/move-vm/types/src/values/values_impl.rs
+++ b/language/move-vm/types/src/values/values_impl.rs
@@ -3461,7 +3461,7 @@ pub mod prop {
             L::U32 => any::<u32>().prop_map(Value::u32).boxed(),
             L::U64 => any::<u64>().prop_map(Value::u64).boxed(),
             L::U128 => any::<u128>().prop_map(Value::u128).boxed(),
-            L::U256 => any::<u256>().prop_map(Value::u256).boxed(),
+            L::U256 => any::<u256::U256>().prop_map(Value::u256).boxed(),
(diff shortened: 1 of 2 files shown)


runtian-zhou (approved):

language/move-vm/types/src/values/values_impl.rs
```diff
@@ -3461,7 +3461,7 @@ pub mod prop {
             L::U32 => any::<u32>().prop_map(Value::u32).boxed(),
             L::U64 => any::<u64>().prop_map(Value::u64).boxed(),
             L::U128 => any::<u128>().prop_map(Value::u128).boxed(),
-            L::U256 => any::<u256>().prop_map(Value::u256).boxed(),
+            L::U256 => any::<u256::U256>().prop_map(Value::u256).boxed(),
```
tnowacki, line 3464: 👀 
What is this change exactly?

vgao1996: `u256` is a module so the code doesn't actually compile. `u256::U256` is the type we have in `move-core-types`.

This did not get detected by CI since the code is behind the "fuzzing" feature flag. <= This is a separate issue we'd better fix

runtian-zhou: This is a lingering change from u256. This piece of code couldn't compile previously IIUC because it was under the fuzzing feature flag, which our CI right now don't check if it is compilable.

PR #723, opened 2022-12-06
Limits for defs in a module

dariorussi:
## Motivation
This introduces a limit for all definitions in a module: structs, fields and functions.
We have some quadratic algorithm in the loader (fields of a type defined in the module) and possibly more.
This reduces the chances of possible attacks and it is a reasonable limitation to add (in fact a great one in my opinion with open publishing).
Also there is some cleanup where we added some config values that were not `Option` as the others. Making everything `Option`.
Finally there is a fix in the limits check in the verifier to run `verify_type_nodes` for scripts

Yes

## Test Plan
Added unit tests

### DIFF
--- a/language/move-bytecode-verifier/src/limits.rs
+++ b/language/move-bytecode-verifier/src/limits.rs
@@ -29,7 +29,8 @@ impl<'a> LimitsVerifier<'a> {
         };
         limit_check.verify_function_handles(config)?;
         limit_check.verify_struct_handles(config)?;
-        limit_check.verify_type_nodes(config)
+        limit_check.verify_type_nodes(config)?;
+        limit_check.verify_definitions(config)
     }
 
     pub fn verify_script(config: &VerifierConfig, module: &'a CompiledScript) -> VMResult<()> {
@@ -44,7 +45,8 @@ impl<'a> LimitsVerifier<'a> {
             resolver: BinaryIndexedView::Script(script),
         };
         limit_check.verify_function_handles(config)?;
-        limit_check.verify_struct_handles(config)
+        limit_check.verify_struct_handles(config)?;
+        limit_check.verify_type_nodes(config)
     }
 
     fn verify_struct_handles(&self, config: &VerifierConfig) -> PartialVMResult<()> {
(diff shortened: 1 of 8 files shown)


language/move-core/types/src/vm_status.rs
```diff
@@ -676,6 +676,9 @@ pub enum StatusCode {
     MEMORY_LIMIT_EXCEEDED = 4028,
     VM_MAX_TYPE_NODES_REACHED = 4029,
     VM_MAX_DEPENDENCY_DEPTH_REACHED = 4030,
+    VM_MAX_FUNTION_DEFINITIONS_REACHED = 4031,
```
sblackshear, line 679:
```suggestion
    VM_MAX_FUNCTION_DEFINITIONS_REACHED = 4031,
```

language/move-bytecode-verifier/src/limits.rs
```diff
@@ -132,4 +134,40 @@ impl<'a> LimitsVerifier<'a> {
...
+            }
+        }
+        if let Some(max_struct_definitions) = config.max_struct_definitions {
+            if let Some(defs) = self.resolver.struct_defs() {
+                if defs.len() > max_struct_definitions {
+                    return Err(PartialVMError::new(
+                        StatusCode::VM_MAX_STRUCT_DEFINITIONS_REACHED,
+                    ));
+                }
+                if let Some(max_fields_in_struct) = config.max_fields_in_struct {
```
sblackshear, line 155: If someone sets `max_struct_definitions` to `None` but `max_fields_in_struct` to `Some(_)`,  the latter will be ignored (which may be confusing).

dariorussi:
fixed issues reported by @sblackshear (thanks!) and added tests about different configurations

language/move-core/types/src/vm_status.rs
dariorussi: done

language/move-bytecode-verifier/src/limits.rs
dariorussi: done, thanks Sam!

language/move-core/types/src/vm_status.rs
```diff
@@ -676,6 +676,9 @@ pub enum StatusCode {
     MEMORY_LIMIT_EXCEEDED = 4028,
     VM_MAX_TYPE_NODES_REACHED = 4029,
     VM_MAX_DEPENDENCY_DEPTH_REACHED = 4030,
+    VM_MAX_FUNCTION_DEFINITIONS_REACHED = 4031,
+    VM_MAX_STRUCT_DEFINITIONS_REACHED = 4032,
+    VM_MAX_FIELD_DEFINITIONS_REACHED = 4033,
```
vgao1996, line 681: So far the codes starting with the VM_ prefix are all runtime errors (4xxx) detected by the VM during execution. I feel like the ones you are adding should belong to the verifier category (1xxx).

dariorussi: that's a good point and let me look into it. 
We want users to be charged when getting that error but I think you are right that a verifier error would accomplish that

dariorussi:
moved errors to the verifier group as suggested by @vgao1996 (good one).
Not clear how much of a breaking change that is, given I moved also one that was there already (though added few days ago), but it seems the right placing in the 1000-2000 group.
Let me know if you have thoughts or opinions. @wrwg take a look please

language/move-core/types/src/vm_status.rs
dariorussi: done

dariorussi:
forgot to ask, this is for everybody, just in case. 
Do we (or anybody) deeply care to have all the `VerifierConfig` new values as `None` instead of a random/reasonable value?
I suppose it does not really matter much given those can be changed any time, and are likely to be explicitly defined by each app, but figured I'd give people a chance to comment before I go in

vgao1996:
@dariorussi I think in general, we actually have to keep a None variant so that the VM offers a way to maintain full backward compatibility

dariorussi:
> @dariorussi I think in general, we actually have to keep a None variant that totally disables the check so that the VM offers a way to maintain full backward compatibility (Not saying that this requirement totally cannot be relaxed -- it's just the assumption we've been working under so far...)

@vgao1996 I agree and not questioning that. My question was more whether `None` should be the default, which it would seem more proper, rather than having a default with values that may be questionable (which is what we have with this PR)... 
thoughts?

vgao1996:
Oh I see. I think both would work as the minimal requirement is to just provide at least a way for the client to keep the old behavior, but having None as the default will probably be less surprising to the them.

wrwg (approved):

PR #726, opened 2022-12-07
[loader] Fixed a node counting problem

awelc:
## Motivation

This fixes a rather subtle bug in the loader implementation. Without adding an extra field to count annotated nodes, initial statement attempting to use the cached data in `struct_gidx_to_fully_annotated_layout` and in `struct_gidx_to_type_layout` could update the node count with the one retrieved from cache  but then NOT return the cached value (as it was cached from a different unannotated/annotated version and as such `struct_info.struct_layout`/`struct_info.annotated_struct_layout` would be, unexpectedly, 'None`). This would lead to spurious (larger) count values.

Yes


tnowacki (approved):
Approving as we cannot update our version of Move without the fix. 
Open to other solutions if someone has ideas though :)

PR #728, opened 2022-12-07
[session] Expose type_to_type_tag

tnowacki:
- Exposed loaders type_to_type_tag function in the session API

## Motivation

- I am currently manually writing this function, but it is already exposed in the NativeContext. Seems reasonable I hope to also expose it when outside of Move (i.e. where the session is being invoked)

### DIFF
--- a/.github/workflows/ci-pre-land.yml
+++ b/.github/workflows/ci-pre-land.yml
@@ -281,9 +281,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Set up Ruby 2.6
-        uses: actions/setup-ruby@v1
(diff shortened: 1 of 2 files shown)


vgao1996 (approved):

.github/workflows/ci-pre-land.yml
```diff
@@ -281,9 +281,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Set up Ruby 2.6
-        uses: actions/setup-ruby@v1
+        uses: ruby/setup-ruby@v1
```
tnowacki, line 284: @vgao1996 heads-up, I needed to change this too. Apparently the old action has been deprecated

PR #735, opened 2022-12-09
[Documentation] Correct bitwise 'or' operator

PaulFidika:
Correct me if I'm wrong, but I believe the bitwise OR operator syntax is '|' rather than '\|'

Did this get changed at some point recently?

### DIFF
--- a/language/documentation/book/src/integers.md
+++ b/language/documentation/book/src/integers.md
@@ -87,7 +87,7 @@ Bitwise operations do not abort.
 | Syntax | Operation  | Description
 |--------|------------|------------
 | `&`    | bitwise and| Performs a boolean and for each bit pairwise
-| `\|`   | bitwise or | Performs a boolean or for each bit pairwise
+| `|`   | bitwise or | Performs a boolean or for each bit pairwise
 | `^`    | bitwise xor| Performs a boolean exclusive or for each bit pairwise
 
 ### Bit Shifts


tnowacki:
I believe the escape `\` exists for rendering on GitHub

sblackshear:
This is a highly obnoxious difference between GitHub markdown and mdbook markdown...

After the PR, Git will be broken like [this](https://github.com/move-language/move/blob/524168ee8154087115845e707272794d96101be6/language/documentation/book/src/integers.md):

image

but mdbook will work. But today, the mdbook shows `\|` which is annoying. I'm sure there's some way to make it work for both, but I don't know how.

geometryolife:
> but mdbook will work. But today, the mdbook shows `\|` which is annoying. I'm sure there's some way to make it work for both, but I don't know how.

GitHub:
image

mdBook:
image

It looks good now, thanks for bringing it up @PaulFidika

sblackshear:
Weird, when I go to https://move-language.github.io/move/integers.html#bitwise I still see the `\|` 
image

geometryolife:
> Weird, when I go to https://move-language.github.io/move/integers.html#bitwise I still see the `\|`

I actually fixed this in #737 , I used HTML entities.

PaulFidika:
> > Weird, when I go to https://move-language.github.io/move/integers.html#bitwise I still see the `\|`
> 
> I actually fixed this in #737 , I used HTML entities.

Thanks for fixing the formatting! I didn't realize there was a rendering difference like that.

PR #739, opened 2022-12-11
[debug] print `std::ascii::String` as string

sblackshear:
`debug::print` special-cases `std::string::String` (utf8 strings), but not `std::ascii::String`. Use the same logic for ascii strings.

### DIFF
--- a/language/move-stdlib/src/natives/debug.rs
+++ b/language/move-stdlib/src/natives/debug.rs
@@ -485,6 +485,26 @@ mod testing {
 
                         let str = move_value_as_escaped_string(val)?;
                         write!(out, "\"{}\"", str).map_err(fmt_error_to_partial_vm_error)?
+                    } else if !canonicalize && type_.is_ascii_string(move_std_addr) {
+                        if fields.len() != 1 {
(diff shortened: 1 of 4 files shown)


language/move-stdlib/src/natives/debug.rs
```diff
@@ -483,6 +483,26 @@ mod testing {
                                 ));
                         }
 
+                        let str = move_value_as_escaped_string(val)?;
+                        write!(out, "\"{}\"", str).map_err(fmt_error_to_partial_vm_error)?
+                    } else if !canonicalize && type_.is_ascii_string(move_std_addr) {
```
tnowacki, line 488: Does it not work to combine these two cases? Something like `if !canonicalize && (type_.is_std_string(move_std_addr) || type_.is_ascii_string(move_std_addr))`

tnowacki: Code looks _really_ similar, but couldn't tell if it would work or not from staring at it

sblackshear: It would work, but it feels fragile (e.g., relies on the fact that both just happen to have a field named `bytes`).

alinush (approved):
This looks good to me! :)

PR #740, opened 2022-12-11
[cli] remove the need to pass `--filter` to `move test`

sblackshear:
`move test <str>` will now run tests that match `<str>`, whereas before you needed to write `move test --filter <str>`.


language/tools/move-cli/src/base/test.rs
```diff
@@ -40,9 +40,9 @@ pub struct Test {
     /// Bound the amount of gas used by any one test.
     #[clap(name = "gas_limit", short = 'i', long = "gas_limit")]
     pub gas_limit: Option<u64>,
-    /// A filter string to determine which unit tests to run. A unit test will be run only if it
+    /// An optional filter string to determine which unit tests to run. A unit test will be run only if it
     /// contains this string in its fully qualified (<addr>::<module_name>::<fn_name>) name.
-    #[clap(name = "filter", short = 'f', long = "filter")]
+    #[clap(name = "filter")]
```
tnowacki, line 45: We should probably improve this more long term, but this feels fine... I think?

tnowacki (approved):

PR #741, opened 2022-12-12
[1/x][move-package/lock] Preparing for Lock file

amnn:
A number of behaviour preserving changes to make it easier to integrate a lock file into the package resolution process.  See [[RFC] Move.lock](https://docs.google.com/document/d/1OV3te-SnpZv2Yxv7uxGQH6NFhE-CdqiCjB66JmYAGKs/edit) for more context.

## Refactor `parsed_manifest::Dependency`, introducing `DependencyKind`

This can be one of `Local`, `Git` or `Custom`, each containing the relevant information for resolving a dependency of that kind, replacing the previous structure where `Dependency` had optional fields for `git_info` and (custom) `node_info`.

This was motivated by two factors:

- Introducing another variant for dependencies (handled by third party package managers), which will be easier to manage when the type is described by an `enum` rather than struct of optionals.
- A new operation that needs to be implemented on dependencies before they are serialized in the lock file, re-rooting:  In the manifest, local dependencies are described relative to the manifest's directory, but the lock file may be for a transitive dependent package, in which case the dependency's root will need to be adjusted (e.g. if it were a local dependency of a git dependency, it will become a git depedency in the lock file).  This operation is also easier to describe with the enum format.

## Extracting functions to download remote dependencies

Previously member functions of `resolution_graph::ResolutionGraph`, but this operation will also be needed while discovering the set of transitive dependencies to write to the lock file, so these were moved to the `resolution` module as top-level (free) functions.

## Removing `download_to` and `local` fields for remote dependencies

These were previously set during manifest parsing, but:

  - They can always be derived from the other fields.
  - They are not read often (i.e. they are read as often as they are written).
  - Their existence complicates the previously mentioned re-rooting operation.
  - They contain absolute paths which should not be reserialized into the lock file, which should be stable across development environments.

So, they have been replaced with more top-level functions in the `resolution` module for figuring out what the local and repository paths should be given a dependency.

## General clean-ups

- Rename `writer: impl Write` to `progress_output` to indicate what gets written to it (as future commits may introduce other streams to write to).
- Use `let .. else { }` to unnest some of the logic in dependency parsing.
- Use `OsStr` when creating git commands, to avoid paths being converted to `&str` and back, which is a lossy operation on some platforms.

# Test Plan

```
```

### DIFF
--- a/language/tools/move-package/src/source_package/manifest_parser.rs
+++ b/language/tools/move-package/src/source_package/manifest_parser.rs
@@ -3,8 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use crate::{package_hooks, source_package::parsed_manifest as PM, Architecture};
-use anyhow::{bail, format_err, Context, Result};
-use move_command_line_common::env::MOVE_HOME;
+use anyhow::{anyhow, bail, format_err, Context, Result};
 use move_core_types::account_address::{AccountAddress, AccountAddressParseError};
 use move_symbol_pool::symbol::Symbol;
 use std::{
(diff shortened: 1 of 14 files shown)


language/tools/move-package/src/source_package/manifest_parser.rs
```diff
@@ -306,155 +305,118 @@ fn parse_address_literal(address_str: &str) -> Result<AccountAddress, AccountAdd
...
-                        "must provide exactly one of {} for dependency.",
-                        keys.join(" or ")
-                    )
-                }
+fn parse_dependency(dep_name: &str, mut tval: TV) -> Result<PM::Dependency> {
+    let Some(table) = tval.as_table_mut() else {
+        bail!("Malformed dependency {}", tval);
+    };
+
+    let mut known_fields = vec![
```
tnowacki, line 313: not needed now, but just wondering if this should a const that gets cloned

tnowacki (approved):

language/tools/move-package/src/source_package/manifest_parser.rs
amnn: Difficult to say whether this will make a big difference, given these are static strings, it's probably a wash.

tnowacki: For me the const was more about intention and readability than perf, though I don't think it matters much

amnn: Ah got it -- not super clear to me in this case, (I'm thinking of a scenario where I see the `const` and assume that the set of known fields is static when in fact it can be updated in the presence of package hooks).

Issue #727, opened 2022-12-07
[Feature Request] Add parse/format support to string


fzgem18:
# 🚀 Feature Request
## Motivation

`string.move` doesn't support the conversion from/to `MoveValue`, which would be useful.

## Pitch

**Describe the solution you'd like**
new functions
```
public fun fmt_utf8<T>(v: &T): String;
public fun parse<T>(s: &String): T;
```
new native functions like
```
native public fun internal_fmt_utf8<T>(v: &T): vector<u8>;
native public fun internal_parse<T>(v: &vector): T;
```

**What I want to discuss**
gas parameters of the new natives

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))
Yes

vgao1996:
@fzgem18 thanks for putting up the feature request! I see that you are trying to add two things with this proposal:
1. String formatting: this is definitely something Move currently lacks and I think implementing `public fun fmt_utf8<T>(v: &T): String;` would be a great first step. Let's work together come up with a more detailed design. cc @wrwg @tnowacki @sblackshear 
2. A generic parse function: this function would allow you to create a struct belonging to an arbitrary module, which is a fatal violation of Move's safety invariants so I don't think we can get it added in the form you suggested. Would you mind elaborating on what problem you are trying to solve with this? Maybe we could find an alternative solution to that.

sblackshear:
One thing we could consider here is exposing the functionality that powers `debug::print` (which I think is essentially what a user would want from `fmt_utf8`).

banool:
@vgao1996 Some maybe useful context is this came up after a variety of similar request came up in #dev-discussions, where people would request "how do I go from x type to a string" or "from a string to x type". Not that Move has traits, but in those terms, what people are reaching for is a Display and FromStr trait, for all types where you'd expect default implementations to exist in other languages.

Naturally that's a lot for just one PR so maybe this PR could just be scoped down to the "x to string" case, and not the other way around.

fzgem18:
Thanks to all for the comments. agree with @banool.
What I requested is only the function that parses string to the primitive types of move. and I thought this is an essential stuff in a language. I am not sure what kinda fatal violation may happen yet, as I am a learner of move. could you please explain more details?

vgao1996:
> only the function that parses string to the primitive types of move. and I thought this is an essential stuff in a language.

[...]

PR #744, opened 2022-12-12
[language] adding `string::fmt_utf8` native function

fzgem18:
WIP

## Motivation

Updated on issue #727  (fmt_utf8 part)

Yes

## Test Plan

For now, simply updated doc and added move unit test `test_fmt_utf8` in `string_tests.move` running successfully.

## Questions

- Gas parameter setting
- Where to update due to the effect with this change?

### DIFF
--- a/language/move-stdlib/src/natives/string.rs
+++ b/language/move-stdlib/src/natives/string.rs
@@ -191,6 +191,40 @@ pub fn make_native_index_of(gas_params: IndexOfGasParameters) -> NativeFunction 
 }
 
 /***************************************************************************************************
+ * native fun internal_fmt_utf8
+ *
+ *   gas cost: base_cost + unit_cost * bytes_to_return
+ *
+ **************************************************************************************************/
+#[derive(Debug, Clone)]
+pub struct FmtUtf8GasParameters {
+    pub base: InternalGas,
+    pub per_byte: InternalGasPerByte,
+}
+
+fn native_fmt_utf8(
+    gas_params: &FmtUtf8GasParameters,
+    _context: &mut NativeContext,
+    _ty_args: Vec<Type>,
+    args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(args.len() == 1);
+
+    let v = Value::vector_u8("string".as_bytes().iter().cloned());
+    let cost = gas_params.base;
+
+    NativeResult::map_partial_vm_result_one(cost, Ok(v))
+}
+
+pub fn make_native_fmt_utf8(gas_params: FmtUtf8GasParameters) -> NativeFunction {
+    Arc::new(
+        move |context, ty_args, args| -> PartialVMResult<NativeResult> {
+            native_fmt_utf8(&gas_params, context, ty_args, args)
+        },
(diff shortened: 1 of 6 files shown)


language/move-stdlib/src/natives/string.rs
```diff
@@ -190,6 +190,45 @@ pub fn make_native_index_of(gas_params: IndexOfGasParameters) -> NativeFunction
...
+    _context: &mut NativeContext,
+    _ty_args: Vec<Type>,
+    mut args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(args.len() == 1);
+
+    // pop value
+    let ref_to_val = pop_arg!(args, Reference);
+    let val = ref_to_val.read_ref()?;
+    let v_str = &val.to_string()[..];
```
wrwg, line 216: The suggestion from @sblackshear was to use existing debug_print functionality for Move. Rust to_string might not be the right thing here.

Independent of that lets first agree on the issue whether we actually want this.

sblackshear: - +1 to attempting to reuse the `debug_print` logic
- One way we could proceed is to start by defining this function, but gating it with `#[test]`. I suspect that including this as test-only functionality is fairly non-controversial, but opening it up as a production function will require a lot more scrutiny. I'd be very curious to hear @alinush's opinion on this

alinush: Hey Sam, when I did the [debug print PR](https://github.com/move-language/move/pull/630), I wrote it so that one could also call it with `canonicalize = true` in a non-testing native function (e.g., a native for canonically-serializing any Move value for signing).

I am no expert on the VM internals though, so I am merely hoping it would be safe to do this.

For example, I hope there are no safety issues around my PR having exposed a `pub fn as_move_value(&self, layout: &MoveTypeLayout) -> MoveValue` function in `Value`.

I also noticed I made a `ValueImpl` function public too, which perhaps was unnecesary. See [here](https://github.com/move-language/move/pull/630/files#diff-36a1671e9470ce7df72a77413b182e251dbe7df3ccaf7a46b5016e045a8cdac9R3535)

sblackshear: >Hey Sam, when I did the debug print PR, I wrote it so that one could also call it with canonicalize = true in a non-testing native function (e.g., a native for canonically-serializing any Move value for signing).

Got it, that is helpful.

> For example, I hope there are no safety issues around my PR having exposed a pub fn as_move_value(&self, layout: &MoveTypeLayout) -> MoveValue function in Value.

None at all!

sblackshear (changes requested):
Discussed this in the Move community meeting and what we decided is:

- We're comfortable with supporting this as test-only functionality in the `debug` module for now. We are planning to add per-type string conversions for primitive types (rather than polymorphic string conversion, as this PR introduces) in the near future.
- We'd prefer to implement this by reusing the existing string conversion functionality in the `debug` module (basically, a native function that returns `out` here: https://github.com/move-language/move/blob/main/language/move-stdlib/src/natives/debug.rs#L50)

PR #745, opened 2022-12-12
[RFC][2/x][move-package/lock] Extend handling of Dependency subdir

amnn:
Proposing a more general handling of the `subdir` field for dependencies, which boils down to:

- Explicitly disallowing the field on `local` dependencies (this will produce an error, where previously it would be silently ignored).
- Enabling its use with custom dependencies.

The second change is not likely to be used directly in `Move.toml` (although it could be), but instead is intended to be used with lock files (which are designed to serialize/deserialize dependencies in a format that can be parsed by the existing `manifest_parser::parse_dependency`), in the context of dependency re-rooting, which already occurs implicitly for `git` dependencies, when:

- `a`, depends on `b`, a git dependency, with subdir `s`, and
- `b`, depends on `c`, a local dependency, with relative path `r`.

From `a`'s perspective, `c` is a git dependency with `subdir` `s.push(r)`, which should still be a path within `b`'s git repo (if it is not, resolution will fail, but with an internal error).

When introducing a lock file, this operation becomes explicit, i.e. in the lock file for `a`, `c` will appear as a git dependency with the aforementioned subdir, which is referred to as "re-rooting" `c`, and the error case of `c`'s path falling outside of `b`'s will become an explicit error.

This raises a question for custom dependencies -- can they have local transitive dependencies?  If so, they need a `subdir` field to be written to the lock file, which this commit proposes, and this makes the `Move.toml` format slightly more expressive as well.  If not, that case can be treated as an error, but this seems overly restrictive.

## Test Plan

New unit test:

```
move-package$ cargo nextest
```

## Stack

- #741

### DIFF
--- a/language/tools/move-package/tests/test_sources/resolution/package_hooks/Move.toml
+++ b/language/tools/move-package/tests/test_sources/resolution/package_hooks/Move.toml
@@ -3,4 +3,4 @@ name = "test"
 version = "0.0.0"
 
 [dependencies]
-Pkg = { custom = "localhost:8080", address = "0x1", package = "framework" }
+Pkg = { custom = "localhost:8080", address = "0x1" }
--- a/language/tools/move-package/src/resolution/mod.rs
+++ b/language/tools/move-package/src/resolution/mod.rs
@@ -278,6 +278,7 @@ fn repository_path(kind: &DependencyKind) -> PathBuf {
             node_url,
             package_address,
             package_name,
+            subdir: _,
         }) => [
             &*MOVE_HOME,
             &format!(
@@ -296,7 +297,9 @@ fn repository_path(kind: &DependencyKind) -> PathBuf {
 fn local_path(kind: &DependencyKind) -> PathBuf {
     let mut repo_path = repository_path(kind);
 
-    if let DependencyKind::Git(GitInfo { subdir, .. }) = kind {
+    if let DependencyKind::Git(GitInfo { subdir, .. })
+    | DependencyKind::Custom(CustomDepInfo { subdir, .. }) = kind
+    {
         repo_path.push(subdir);
     }
 
--- a/language/tools/move-package/src/source_package/manifest_parser.rs
+++ b/language/tools/move-package/src/source_package/manifest_parser.rs
@@ -337,10 +337,15 @@ fn parse_dependency(dep_name: &str, mut tval: TV) -> Result<PM::Dependency> {
 
     let kind = match (
         table.remove("local"),
+        table.remove("subdir"),
         table.remove("git"),
         custom_key_opt.as_ref().and_then(|k| table.remove(k)),
     ) {
-        (Some(local), None, None) => {
+        (Some(local), subdir, None, None) => {
+            if subdir.is_some() {
+                bail!("'subdir' not supported for local dependencies");
+            }
+
             let Some(local) = local.as_str().map(PathBuf::from) else {
                 bail!("Local source path not a string")
             };
@@ -348,7 +353,7 @@ fn parse_dependency(dep_name: &str, mut tval: TV) -> Result<PM::Dependency> {
             PM::DependencyKind::Local(local)
         }
 
-        (None, Some(git_url), None) => {
+        (None, subdir, Some(git_url), None) => {
             let Some(git_rev) = table.remove("rev") else {
                 bail!("Git revision not supplied for dependency")
             };
@@ -361,7 +366,7 @@ fn parse_dependency(dep_name: &str, mut tval: TV) -> Result<PM::Dependency> {
                 bail!("Git URL not a string")
             };
 
-            let subdir = match table.remove("subdir") {
+            let subdir = match subdir {
                 None => PathBuf::new(),
                 Some(path) => path
                     .as_str()
@@ -376,7 +381,7 @@ fn parse_dependency(dep_name: &str, mut tval: TV) -> Result<PM::Dependency> {
             })
         }
 
-        (None, None, Some(custom_key)) => {
+        (None, subdir, None, Some(custom_key)) => {
             let Some(package_address) = table.remove("address") else {
                 bail!("Address not supplied for 'node' dependency");
             };
@@ -389,12 +394,21 @@ fn parse_dependency(dep_name: &str, mut tval: TV) -> Result<PM::Dependency> {
                 bail!("Git URL not a string")
             };
 
+            let subdir = match subdir {
+                None => PathBuf::new(),
+                Some(path) => path
+                    .as_str()
+                    .map(PathBuf::from)
+                    .ok_or_else(|| anyhow!("'subdir' not a string"))?,
+            };
+
             let package_name = Symbol::from(dep_name);
 
             PM::DependencyKind::Custom(PM::CustomDepInfo {
                 node_url,
                 package_address,
                 package_name,
+                subdir,
             })
         }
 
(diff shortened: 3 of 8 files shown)


language/tools/move-package/tests/test_sources/resolution/package_hooks/Move.toml
```diff
@@ -3,4 +3,4 @@ name = "test"
 version = "0.0.0"
 
 [dependencies]
-Pkg = { custom = "localhost:8080", address = "0x1", package = "framework" }
```
amnn, line 6: Was the `package = "framework"` field used before?  From what I could tell, the answer was no, and the field was triggering a warning during `Move.toml` parsing so I removed it, but I can add it back.

tnowacki: I think the idea was for renaming of dependencies if needed, so something like `PkgA = { url = ..., package = "pkg" } PkgB = { ..., package = "pkg" }`

tnowacki:
> If not, that case can be treated as an error, but this seems overly restrictive.

I think it might be best to handle these sort of things for custom deps as the needs arise. In other words, let's keep things restrictive for now, and we can relax it as time goes on if needed. I just foresee custom deps being more granular than git or local deps. 
Thoughts?

language/tools/move-package/tests/test_sources/resolution/package_hooks/Move.toml
amnn: Hmm, do you know if that is supposed to still work or is it a holdover? The manifest parser doesn't recognise the field, and it doesn't get passed to the package hook (I also looked at the [`AptosPackageHook`](https://github.com/aptos-labs/aptos-core/blob/main/crates/aptos/src/move_tool/package_hooks.rs))

amnn:
> I think it might be best to handle these sort of things for custom deps as the needs arise.

To clarify the context, the need in this case was that this seemed the most straightforward way to implement a necessary feature in lock files (more straightforward than not allowing the pattern, even).

> I just foresee custom deps being more granular than git or local deps.

I definitely agree.  I don't think people will (need to) use the `subdir` field on custom dependencies in `Move.toml`, but what I was on the fence about was a custom dependency which has a (relative) local dependency in turn (i.e. the local dependency's source is contained within the artifact downloaded for the custom dependency) -- to support this in lock files requires a `subdir` field in the equivalent definition in the lock file schema, and if it exists there, it might as well exist in `Move.toml` (it is simpler to implement that way).

Today, this will just work because packages are resolved recursively, so the custom dependency gets resolved, and then its dependencies are resolved relative to it (similar for git dependencies where I know we do this quite a bit, which is why I was more on the fence about this case), and there's no need for a `subdir` field.

But when lock files are introduced, they need to describe a package's transitive dependencies, so if we have `A` custom depending on `B` (relative) locally depending on `C`, then in `A`'s `Move.lock`, we need some way to describe `C` relative to `A`.

An example of the same thing happening for all local dependencies looks like this (taken from [this test](https://github.com/move-language/move/pull/760/files#diff-5cd659faf1fa2d8aeeaef6a32074ae6e282d45e1bb697be6719ed641dcf8967f) in #760), where these manifests:

```toml
# ./Move.toml
[package]
name = "NestedDeps"
version = "0.0.0"

[dependencies]
Nested = { local = "./deps_only/nested" }

# ./deps_only/nested/Move.toml
[package]
name = "Nested"
version = "0.0.0"

[dependencies]
More = { local = "./more" }

# ./deps_only/nested/more/Move.toml
[package]
name = "More"
version = "0.0.0"
```

Turn into this lock file:

```toml
# @generated by Move, please check-in and do not edit manually.

[move]
version = 0

[[move.dependency]]
name = "More"
source = { local = "deps_only/nested/more" }

[[move.dependency]]
name = "Nested"
source = { local = "deps_only/nested" }
dependencies = [
  "More",
]
```

The path to package `More` in the lock file is `deps_only/nested/more`, not just `./more`, becase it's relative to the root package now.  The same happens for a git dep that has a local dep: The transitive dependency becomes a git dep with its `subdir`fixed up to point to the local dep (example [here](https://github.com/move-language/move/pull/760/files#diff-150a8a1c48091e2512427e737b1f79708b92686239308a29adf809bb535e5f3f)), the logic to do this is [here](https://github.com/move-language/move/pull/760/files#diff-1ba60059eed6183ca91774b152b33684785e3d95d214d8bf9d0028d6c32afd03R104).

Currently the parser for the `source` field in the lock file is exactly the same as the one for a value in the `dependencies` table of a manifest, so when re-rooting support requires adding `subdir` for custom dependencies in the lockfile, it becomes permissible in the manifest as well as a side-effect.  The alternatives are:

- To duplicate the parsing logic for lock files and specialise it to allow a `subdir` field for custom deps, but not in the manifest.  This is a lot of work if the feature is benign (and in some sense, achievable using transitive dependencies).
- To explicitly disallow transitive local dependencies on custom dependencies, which is doable, but seems like a confusing error to navigate.

language/tools/move-package/tests/test_sources/resolution/package_hooks/Move.toml
tnowacki: no idea, sorry :)

tnowacki:
> Currently the parser for the `source` field in the lock file is exactly the same as the one for a value in the `dependencies` table of a manifest, so when re-rooting support requires adding `subdir` for custom dependencies in the lockfile, it becomes permissible in the manifest as well as a side-effect. The alternatives are:
> 
> * To duplicate the parsing logic for lock files and specialise it to allow a `subdir` field for custom deps, but not in the manifest.  This is a lot of work if the feature is benign (and in some sense, achievable using transitive dependencies).
> * To explicitly disallow transitive local dependencies on custom dependencies, which is doable, but seems like a confusing error to navigate.

Thanks for the detailed explanation! I think I have a better understanding of what's going on.

Going back to this original question 

> This raises a question for custom dependencies -- can they have local transitive dependencies? 

I suppose I feel like the answer is "sure? why not?". Sorry thats maybe not the most helpful answer. 

But another thought, could some custom dependency provider ban this behavior if they wanted to? 
My thought is perhaps this seems reasonable to us now, but is mayhem for a custom dep provider to handle. Can they ban it easily?

amnn:
> I suppose I feel like the answer is "sure? why not?". Sorry thats maybe not the most helpful answer.

No, I think that's helpful -- this was my thinking as well, so it's comforting to have it corroborated.

In terms of disabling it, the answer is also a bit unsatisfying:

- it can't be disabled in the package hook, but it couldn't be before (and in fact the `subdir` field would have been silently ignored by the `local` and custom dependency types, while after this stack it will either be explicitly handled or explicitly error).
- it can be disabled by making it impossible to store a package with sub-directories in the package repository that the package hook is pulling from.

tnowacki:
> it can be disabled by making it impossible to store a package with sub-directories in the package repository that the package hook is pulling from.

This feels reasonable to me as an answer for any package repo that doesn't want to support this. If I understand this correctly, it just means they would have some extra set of checks when uploading to the repository?

amnn:
Yep, that's right, by doing that, the `subdir` field becomes useless -- you can still set it, but all it will do is cause manifest parsing to fail (which is probably one step better than today where it is silently ignored).

tnowacki (approved):

vgao1996 (approved):

PR #747, opened 2022-12-13
[language] Add catch_unwind panic handling

chen-robert:
## Motivation

This helps harden the VM attack surface by catching panics that happen in Move deserialization and verification, similar to #704. All objects passed across the `catch_unwind` boundary are read-only and thus [UnwindSafe](https://doc.rust-lang.org/std/panic/trait.UnwindSafe.html)[^1]. 

This PR also exposes an API to set and query the state of the Move VM via a shared thread-local. This can be useful for custom panic handlers that want to abort _unless_ the code is executing inside a known panic handler (such as deserialization or verification). 

## Test Plan

Not super familiar with the workflows here, open to ideas for what tests should be added. 

[^1] A stronger argument for safety is as follows. 

`catch_unwind` can [never lead to soundness violations](https://doc.rust-lang.org/core/panic/trait.UnwindSafe.html#what-is-unwindsafe).

To break a logical invariant, we need to persist state. This implies that we need to be concerned about two types of modifications. 
1. Objects that are passed through the `catch_unwind` boundary. All objects are fully immutable (including interior mutability), meaning this isn't a concern. 
2. Global state. Rust's implementation of `std::thread` also [uses catch_unwind](https://github.com/rust-lang/rust/blob/master/library/std/src/thread/mod.rs#L549), implying that functions that are safe to run on threads also can't violate logical invariants on panic. The one exception would be thread local data, but if we're concerned can also get around that by spawning a separate thread.

### DIFF
--- a/language/move-binary-format/src/deserializer.rs
+++ b/language/move-binary-format/src/deserializer.rs
@@ -4,7 +4,7 @@
 
 use crate::{check_bounds::BoundsChecker, errors::*, file_format::*, file_format_common::*};
 use move_core_types::{
-    account_address::AccountAddress, identifier::Identifier, metadata::Metadata,
+    account_address::AccountAddress, identifier::Identifier, metadata::Metadata, state::VMState,
     vm_status::StatusCode,
 };
 use std::{collections::HashSet, convert::TryInto, io::Read};
@@ -43,9 +43,21 @@ impl CompiledModule {
         binary: &[u8],
         max_binary_format_version: u32,
     ) -> BinaryLoaderResult<Self> {
-        let module = deserialize_compiled_module(binary, max_binary_format_version)?;
-        BoundsChecker::verify_module(&module)?;
-        Ok(module)
+        let prev_state = move_core_types::state::set_state(VMState::DESERIALIZER);
+        let result = std::panic::catch_unwind(|| {
+            let module = deserialize_compiled_module(binary, max_binary_format_version)?;
+            BoundsChecker::verify_module(&module)?;
+
+            Ok(module)
+        })
+        .unwrap_or_else(|_| {
+            Err(PartialVMError::new(
+                StatusCode::VERIFIER_INVARIANT_VIOLATION,
+            ))
+        });
+        move_core_types::state::set_state(prev_state);
+
+        result
     }
 
     // exposed as a public function to enable testing the deserializer
--- a/language/move-core/types/src/state.rs
+++ b/language/move-core/types/src/state.rs
@@ -1,0 +1,21 @@
+use std::cell::RefCell;
+
+#[derive(Clone, Copy, Debug)]
+pub enum VMState {
+    DESERIALIZER,
+    VERIFIER,
+    RUNTIME,
+    OTHER,
+}
+
+thread_local! {
+    static STATE: RefCell<VMState> = RefCell::new(VMState::OTHER);
+}
+
+pub fn set_state(state: VMState) -> VMState {
+    STATE.with(|s| s.replace(state))
+}
+
+pub fn get_state() -> VMState {
+    STATE.with(|s| *s.borrow())
+}
--- a/language/move-core/types/src/state.rs
+++ b/language/move-core/types/src/state.rs
@@ -1,3 +1,7 @@
+// Copyright (c) The Diem Core Contributors
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
 use std::cell::RefCell;
 
 #[derive(Clone, Copy, Debug)]
--- a/language/move-bytecode-verifier/src/verifier.rs
+++ b/language/move-bytecode-verifier/src/verifier.rs
@@ -13,9 +13,10 @@ use crate::{
 };
 use move_binary_format::{
     check_bounds::BoundsChecker,
-    errors::{Location, VMResult},
+    errors::{Location, PartialVMError, VMResult},
     file_format::{CompiledModule, CompiledScript},
 };
+use move_core_types::{state::VMState, vm_status::StatusCode};
 
 #[derive(Debug, Clone)]
 pub struct VerifierConfig {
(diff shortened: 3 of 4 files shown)


tnowacki (commented):
Do we need to land this `state` alongside this PR? I'm not against it and have no real strong opinions, but just curious as it looks to be unused (as far as I can tell) and it's usage looks speculative at this point

chen-robert:
@tnowacki I believe it's relevant for some projects such as Aptos which register [their own panic hook](https://github.com/aptos-labs/aptos-core/blob/e32cb5ff4cbe5f64a57ee7c69ba38faa6e894414/crates/crash-handler/src/lib.rs#L48) to abort the process in case of a panic. In this case, they would probably want to abort unless the panic happened during Move bytecode verification or deserialization.

tnowacki:
Thanks for the context, makes sense to me!

language/move-core/types/src/state.rs
```diff
@@ -0,0 +1,21 @@
...
+    VERIFIER,
+    RUNTIME,
+    OTHER,
+}
+
+thread_local! {
+    static STATE: RefCell<VMState> = RefCell::new(VMState::OTHER);
+}
+
+pub fn set_state(state: VMState) -> VMState {
```
tnowacki, line 15: I wish we could make this pub(crate) if we imagine folks reading get_state

tnowacki (commented):
Happy to approve this, but I am not the most familiar with `std::panic::catch_unwind` and any related intricacies, so getting some other eyes might be worthwhile before landing. Maybe @vgao1996 or @runtian-zhou?

language/move-binary-format/src/deserializer.rs
```diff
@@ -43,9 +43,21 @@ impl CompiledModule {
         binary: &[u8],
         max_binary_format_version: u32,
     ) -> BinaryLoaderResult<Self> {
-        let module = deserialize_compiled_module(binary, max_binary_format_version)?;
-        BoundsChecker::verify_module(&module)?;
-        Ok(module)
+        let prev_state = move_core_types::state::set_state(VMState::DESERIALIZER);
```
runtian-zhou, line 46: Do you mind explain a bit why we need this global variable? It doesn't seem to me that we are using this anywhere? Is it just for external observability?

chen-robert: This would be relevant for panic handlers that might want to change behavior depending on where the panic originates. 

> This PR also exposes an API to set and query the state of the Move VM via a shared thread-local. This can be useful for custom panic handlers that want to abort unless the code is executing inside a known panic handler (such as deserialization or verification).

language/move-core/types/src/state.rs
chen-robert: Would `pub(crate)` work here? I think it would break access because `move-core/types` is its own crate.

runtian-zhou (approved):

language/move-core/types/src/state.rs
```diff
@@ -0,0 +1,25 @@
...
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use std::cell::RefCell;
+
+#[derive(Clone, Copy, Debug)]
+pub enum VMState {
+    DESERIALIZER,
+    VERIFIER,
+    RUNTIME,
```
dariorussi, line 11: this is currently unused and I assume you have it here in case we may want to have runtime catches?
Wonder fi we want it at all, question for you (@chen-robert) and everybody else

language/move-core/types/src/state.rs
dariorussi: I agree that public is inevitable here as this may be needed "way outside" the Move VM

language/move-binary-format/src/deserializer.rs
dariorussi: the point @chen-robert is making here is quite important and let me try to give it another spin as well (sorry if well understood and repeated once again).
If any system embedding the VM has a global and custom panic handler installed, that handler will be called before the `catch_unwind` here. If, say, that global handler exited the process (as for instance the custom handler installed by the blockchain process in the original Libra code) the `catch_unwind` here would never run.
The principle here (and the conversation we had) was that a global custom handler could check the thread local and see the state of it. And if inside a verification or deserialization state, do nothing - or something specific - knowing that the `catch_unwind` may be called.

language/move-core/types/src/state.rs
```diff
@@ -0,0 +1,25 @@
...
+    OTHER,
```
dariorussi, line 12: do we have a better name for this? not a big deal but this really means we are not within a `catch_unwind` right?

language/move-binary-format/src/deserializer.rs
```diff
@@ -43,9 +43,21 @@ impl CompiledModule {
...
+        let result = std::panic::catch_unwind(|| {
+            let module = deserialize_compiled_module(binary, max_binary_format_version)?;
+            BoundsChecker::verify_module(&module)?;
+
+            Ok(module)
+        })
+        .unwrap_or_else(|_| {
+            Err(PartialVMError::new(
+                StatusCode::VERIFIER_INVARIANT_VIOLATION,
```
dariorussi, line 55: I would go for a different and specific error here. `VERIFIER_INVARIANT_VIOLATION` may be handled in different ways from system embedding the VM and  this error really has strong implications.
To be honest I am not even clear `VERIFIER_INVARIANT_VIOLATION` is an "impossible" error.
Thoughts?

sblackshear:
> Not super familiar with the workflows here, open to ideas for what tests should be added.

Two ideas on testing:
- Use something like https://docs.rs/failpoints/latest/failpoints/ to inject crashes in various verifier passes
- Add an extra pass gated by `cfg(test)` that crashes in various ways when you give it different "magic" modules

runtian-zhou:
Re: testing: Here's how I test the paranoid mode:

I first set the failpoints here: https://github.com/move-language/move/blob/4fd451044144ab5345aad3eda88c0da2a37b070c/language/move-vm/runtime/src/loader.rs#L685

And then I set the fail behavior here: https://github.com/move-language/move/blob/4fd451044144ab5345aad3eda88c0da2a37b070c/language/move-vm/paranoid-tests/tests/tests.rs#L12

The tests can then inject the behavior to the runtime so that I could test the paranoid logics. I think we can do sth similar for this PR. i.e: Add a failpoint at the beginning of verifier logic and set the failpoint behavior to panic. Lemme know if you need any help there!

runtian-zhou:
Commandered the PR with a new test!

language/move-core/types/src/state.rs
chen-robert: Yeah, added it in for runtime catches, might make sense to remove for now and add it back later.

PR #749, opened 2022-12-15
Type instantiation cache

dariorussi:
## Motivation

Make a type instantiation cache to avoid stalling the system

Yes

## Test Plan
Not yet


language/move-vm/runtime/src/loader.rs
```diff
@@ -1270,6 +1270,26 @@ impl Loader {
         Ok(())
     }
 
+    // Return an instantiated type given a generic and an instantiation.
+    // Stopgap to avoid a recursion that is either taking too long or using too
+    // much memory
+    fn subst(&self, ty: &Type, ty_args: &[Type]) -> PartialVMResult<Type> {
```
wrwg, line 1276: Good to have this here!

PR #750, opened 2022-12-15
[language] Add catch_unwind panic handling

runtian-zhou:
## Motivation

Commandered the #747. Added testing via failpoint injection.

Yes
## Test Plan

With a new test.

### DIFF
--- a/language/move-binary-format/src/deserializer.rs
+++ b/language/move-binary-format/src/deserializer.rs
@@ -4,7 +4,7 @@
 
 use crate::{check_bounds::BoundsChecker, errors::*, file_format::*, file_format_common::*};
 use move_core_types::{
-    account_address::AccountAddress, identifier::Identifier, metadata::Metadata,
+    account_address::AccountAddress, identifier::Identifier, metadata::Metadata, state::VMState,
     vm_status::StatusCode,
 };
 use std::{collections::HashSet, convert::TryInto, io::Read};
@@ -43,9 +43,21 @@ impl CompiledModule {
         binary: &[u8],
         max_binary_format_version: u32,
     ) -> BinaryLoaderResult<Self> {
-        let module = deserialize_compiled_module(binary, max_binary_format_version)?;
-        BoundsChecker::verify_module(&module)?;
-        Ok(module)
+        let prev_state = move_core_types::state::set_state(VMState::DESERIALIZER);
+        let result = std::panic::catch_unwind(|| {
+            let module = deserialize_compiled_module(binary, max_binary_format_version)?;
+            BoundsChecker::verify_module(&module)?;
+
+            Ok(module)
+        })
+        .unwrap_or_else(|_| {
+            Err(PartialVMError::new(
+                StatusCode::VERIFIER_INVARIANT_VIOLATION,
+            ))
+        });
+        move_core_types::state::set_state(prev_state);
+
+        result
     }
 
     // exposed as a public function to enable testing the deserializer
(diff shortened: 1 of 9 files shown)


language/move-binary-format/src/deserializer.rs
```diff
@@ -43,9 +43,21 @@ impl CompiledModule {
...
+        let prev_state = move_core_types::state::set_state(VMState::DESERIALIZER);
+        let result = std::panic::catch_unwind(|| {
+            let module = deserialize_compiled_module(binary, max_binary_format_version)?;
+            BoundsChecker::verify_module(&module)?;
+
+            Ok(module)
+        })
+        .unwrap_or_else(|_| {
+            Err(PartialVMError::new(
+                StatusCode::VERIFIER_INVARIANT_VIOLATION,
```
dariorussi, line 55: can we make a specific error instead of a `VERIFIER_INVARIANT_VIOLATION`?
`VERIFIER_INVARIANT_VIOLATION` is raised in other context as well and its handling may vary depending on the app embedding the VM.
I think this deserves its own error that would be easy to filter out from a client which could do the proper thing.
For instance I would imagine a client would want to raise an alert in this cases.
In that respect I even wonder if this code should live on the client and we should have clear entry points (which I think we do). I guess filtering a specific error may be fine and a client may have all the info in place to log something helpful.
Thoughts on this?

language/move-bytecode-verifier/bytecode-verifier-tests/src/unit_tests/catch_unwind.rs
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) The Move Contributors
```
dariorussi, line 1: this is great thanks, though in time it would be good to have some multithread tests with injected panics. 
Also it would be lovely to have a OOM error and see what happens.
Again up to you and we can do that later

language/move-core/types/src/state.rs
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+use std::cell::RefCell;
+
+#[derive(Clone, Copy, Debug)]
+pub enum VMState {
+    DESERIALIZER,
+    VERIFIER,
+    RUNTIME,
+    OTHER,
```
dariorussi, line 12: I don't know about `OTHER` I think a more meaningful state that indicates we are not inside a catch unwind would be great. Something like `YO_NOT_IN_A_CATCH_UNWIND_YOU_YO` the point I was trying to make is the following:
a thread enters the VM, hits deserializer and verifier and there is no problem. The thread local is now left with `OTHER` later that thread panics and a global panic handler picks up and checks the thread local, if it finds `OTHER` that means there is no catch unwind and the global handler can do as it wishes to do.
So `OTHER` really means not in a catch unwind but it hardly reads that way.
BTW I love `YO_NOT_IN_A_CATCH_UNWIND_YOU_YO` and if we put it there I will accept immediately

dariorussi (commented):
two things for me that I would love us to think about:
1- should this be under a feature flag? This would work very poorly with the global handler that APTOS or the old Libra/DIEM code uses. A compilation directive sounds much more proper to me (but it's me)
2- should this code be on the client? I wonder about this code living in the VM and I am not sure that is the right thing. Particularly considering that would have to be integrated with any client that has got global panics handler (which I would assume is all of them, and if not they will be at some point). It seems this code is wrapping entire entry point which makes me wonder if it is the right thing. I am personally starting to think this should not live in the VM. Which, incidentally, would free any client from having to agree on something that is truly a client choice....

language/move-binary-format/src/deserializer.rs
wrwg: This error code looks good to me as long as we attach a message. Its used in many places for similar reason, and we should have monitoring and alerting on for it. Besides we need to keep those PRs minimal please for cherry picking.

wrwg (approved):
Lets land this and cherry pick! As @dariorussi suggested, a feature flag should be added IMO.

Issue #729, opened 2022-12-07
[Prover][Bug] another instance of data invariant issue in loop


junkil-park:
## How to reproduce
Run Prover for the following example:
```
module 0x42::havoced_pool {
    use std::vector;
    use aptos_std::simple_map::{Self, SimpleMap};

    // A simplified version of pool_u64::Pool.
    struct Pool {
        shares: SimpleMap<address, u64>,
        shareholders: vector<address>,
        dummy: bool,
    }
    spec Pool {
        // `shares` contains the key `addr` if and only if `shareholders` contains `addr.
        invariant forall addr: address:
            (simple_map::spec_contains_key(shares, addr) == vector::spec_contains(shareholders, addr));

        // `shareholders` does not contain duplicates.
        invariant forall i in 0..len(shareholders), j in 0..len(shareholders):
            shareholders[i] == shareholders[j] ==> i == j;
    }

    public fun test(): Pool {
        let i = 0;
        let p = Pool {
            shares: simple_map::create<address, u64>(),
            shareholders: vector::empty<address>(),
            dummy: false,
        };
        while(i < 10) {
            p.dummy = true;
            i = i + 1;
        };
        p
    }
}
```

Prover produces the following output:
```
error: data invariant does not hold
   ┌─ ./sources/havoced_pool.move:13:9
   │
13 │ ╭         invariant forall addr: address:
14 │ │             (simple_map::spec_contains_key(shares, addr) == vector::spec_contains(shareholders, addr));
   │ ╰───────────────────────────────────────────────────────────────────────────────────────────────────────^
   │
   =     at ./sources/havoced_pool.move:22: test
   =         i = 0
   =     at ./sources/havoced_pool.move:24: test
   =     at ./sources/havoced_pool.move:25: test
   =     at ./sources/havoced_pool.move:26: test
   =     at ./sources/havoced_pool.move:23: test
   =     at ./sources/havoced_pool.move:13
   =     at ./sources/havoced_pool.move:17
   =     at ./sources/havoced_pool.move:23: test
   =         p =
   =           havoced_pool.Pool{
   =             shares = Table{},
   =             shareholders = vector{},
   =             dummy = false}
   =     at ./sources/havoced_pool.move:29: test
   =     enter loop, variable(s) i, p havocked and reassigned
   =         i = 7
   =         p =
   =           havoced_pool.Pool{
   =             shares = Table{..},
   =             shareholders =
[...]

PR #752, opened 2022-12-16
[move-prover] explicit well-formed assumption after havoc

### LINKED ISSUE #716: [Prover][Bug] struct invariants for vector elements are not properly assumed after havocing the vector in a loop
It seems that struct invariants for vector elements are not properly assumed after havocing the vector in a loop.

## How to reproduce

Run Prover for the following module:
```
module 0x42::havoced_option {
    use std::option::{Self, Option};
    use std::vector;

    public fun test() {
        let i = 0;
        let r = vector::empty<Option<u64>>();
        while(i < 10) {
            vector::push_back(&mut r, option::none());
            i = i + 1;
        }
    }
}
```

Prover produces the following error:
```
error: data invariant does not hold
   ┌─ /Users/jpark/Work/aptos-core/aptos-move/framework/aptos-stdlib/../move-stdlib/sources/option.move:13:9
   │
13 │         invariant len(vec) <= 1;
   │         ^^^^^^^^^^^^^^^^^^^^^^^^
   │
   =     at ./sources/havoced_option.move:6: test
   =         i = 0
   =     at ./sources/havoced_option.move:7: test
   =         r = vector{}
   =     at ./sources/havoced_option.move:8: test
   =     enter loop, variable(s) i, r havocked and reassigned
   =         i = 1
   =         r =
   =           vector{
   =             (size): 18446744073709539482,
   =             6334: <? List([Literal("Vec_10341"), Literal("|T@[Int]Int!val!0|"), Literal("2")])>,
   =             default: undef}
   =     at ./sources/havoced_option.move:9: test
   =     at /Users/jpark/Work/aptos-core/aptos-move/framework/aptos-stdlib/../move-stdlib/sources/option.move:13

Error: exiting with verification errors
```

Looking at the generated Boogie file `output.bpl`, the variable `r` is havoced in the loop, but the data invariants for the elements of `r` are not properly assumed:
```
...
    // $t1 := vector::empty<option::Option<u64>>() on_abort goto L6 with $t3 at ./sources/havoced_option.move:7:17+28
    assume {:print "$at(2,148,176)"} true;
    call $t1 := $1_vector_empty'$1_option_Option'u64''();
    if ($abort_flag) {
        assume {:print "$at(2,148,176)"} true;
        $t3 := $abort_code;
        assume {:print "$track_abort(2,0):", $t3} $t3 == $t3;
        goto L6;
    }
...
    // $t1 := havoc[val]() at ./sources/havoced_option.move:8:15+1
    havoc $t1;
    assume $IsValid'vec'$1_option_Option'u64'''($t1);
...
```
Note that `$IsValid'vec'$1_option_Option'u64'''` does not include the assumption about the data invariants.

## Expected output
Prover produces no error.

meng-xu-cs:
Currently well-formed check is instrumented at Boogie translation step. This commit explicitly instrument the well-formed check right after the havoc bytecode.

This should be semantically equivalent in terms of loop handling, but greatly simplifies data invariant instrumentation as the data invariant instrumentation pass looks for `WellFormed` opcode as hook-up points to instrument the data invariants.

This commit closes #716.

## Motivation

Data invariants in loops

Yes

## Test Plan

- CI
- A new test case is added

### DIFF
--- a/language/move-prover/boogie-backend/src/bytecode_translator.rs
+++ b/language/move-prover/boogie-backend/src/bytecode_translator.rs
@@ -1298,12 +1298,6 @@ impl<'env> FunctionTranslator<'env> {
                     Havoc(HavocKind::Value) | Havoc(HavocKind::MutationAll) => {
                         let var_str = str_local(dests[0]);
                         emitln!(writer, "havoc {};", var_str);
-                        // Insert a WellFormed check
-                        let ty = &self.get_local_type(dests[0]);
-                        let check = boogie_well_formed_check(env, &var_str, ty);
-                        if !check.is_empty() {
-                            emitln!(writer, &check);
-                        }
                     }
                     Havoc(HavocKind::MutationValue) => {
                         let ty = &self.get_local_type(dests[0]);
@@ -1317,11 +1311,6 @@ impl<'env> FunctionTranslator<'env> {
                             var_str,
                             temp_str
                         );
-                        // Insert a WellFormed check
-                        let check = boogie_well_formed_check(env, &var_str, ty);
-                        if !check.is_empty() {
-                            emitln!(writer, &check);
-                        }
                     }
                     Stop => {
                         // the two statements combined terminate any execution trace that reaches it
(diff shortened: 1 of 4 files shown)


rahxephon89 (approved):
LGTM, thanks!

junkil-park (approved):
Thank you for the fix!! Could you also add the example of #729 as a test case? It can go to the same test file `data_invariant_in_loop.move` with renaming one of the function names.

meng-xu-cs:
Sorry that I missed the adding a test case part. Will follow up with another PR.

PR #753, opened 2022-12-16
[3/x][move-package/lock] Refactor move-package test-runner

amnn:
Some updates to the test runner to make it more manageable to add a fourth kind of test (for lock file snapshots).

- Change set-up so that for each test there is one snapshot file, (Not `Move.exp` and one of `Move.compile`, `Move.model` etc).
- Extract out functions for scrubbing test output.
- Introduce helper functions to make use of try (`?`) syntax.
- "Built model" files get a trailing newline to make the linter happy.

## Test Plan

```
move/language/tools/move-package$ cargo nextest --no-fail-fast
```

## Stack

- #741 
- #745


language/tools/move-package/tests/test_runner.rs
```diff
@@ -2,135 +2,164 @@
...
+        let mut sink = std::io::sink();
+        let mut resolved_package =
+            config.resolution_graph_for_package(self.toml_path, &mut sink)?;
+
+        let Some(ext) = self.expected.extension().and_then(OsStr::to_str) else {
+            bail!("Unexpected snapshot file extension: {:?}", self.expected.extension());
+        };
+
+        Ok(match ext {
+            "compiled" => {
```
tnowacki, line 119: anyway to bring back the constants for these?

language/tools/move-package/tests/test_runner.rs
```diff
@@ -2,135 +2,164 @@
...
+                "No snapshot file found for {:?}, please add a file with the same basename and one \
+                 of the following extensions: {:#?}\n\n\
+                 You probably want to re-run with `env UPDATE_BASELINE=1` after adding this file.",
+                toml_path,
+                EXTENSIONS,
+            ).into());
+        };
+
+        let mut remaining: Vec<_> = candidates.collect();
+        if !remaining.is_empty() {
```
tnowacki, line 82: Could we support just running the tool for each snapshot? Could be helpful and let us re-use toml files (which are duplicated a lot today if I recall correctly

tnowacki (approved):
3/x looks good. But I think it would be interesting to just run the command/test for each snapshot. And might allow for a lot more reuse among tests

The problem though persisted before this change so no need to block for it

language/tools/move-package/tests/test_runner.rs
amnn: I had this idea as well... Check out two PRs up in the stack.

tnowacki:
Oh, looks like my suggestion is 5/x? You are way ahead of me  :P

language/tools/move-package/tests/test_runner.rs
amnn: It's possible (to match against a `const`), but I was finding that pulling these values out into constants wasn't really helping me, because the constant names matched the contents of the string literal, so if I changed one, I would probably change the other anyway, and they are only used in two places (here and in the list of all extensions), so the extra indirection wasn't providing me with much.

That being said, I don't feel strongly about this, so let me know if you feel slightly stronger, and I can pull these out into constants.

tnowacki: No strong feelings, just a knee-jerk reaction when I see important values like this sitting all alone without a cozy `const` to keep them safe

language/tools/move-package/tests/test_runner.rs
```diff
@@ -2,135 +2,164 @@
...
+            .iter()
+            .map(|ext| toml_path.with_extension(*ext))
+            .filter(|p| p.is_file());
+
+        let Some(expected) = candidates.next() else {
+            return Err(anyhow!(
+                "No snapshot file found for {:?}, please add a file with the same basename and one \
+                 of the following extensions: {:#?}\n\n\
+                 You probably want to re-run with `env UPDATE_BASELINE=1` after adding this file.",
+                toml_path,
```
vgao1996, line 76: So is my understanding correct that, with this change, running the test with UPDATE_BASELINE=1 set will not automatically generate the snapshot files if they do not exist?

vgao1996 (approved):

language/tools/move-package/tests/test_runner.rs
amnn: That's correct, this becomes more palatable after #759, when a given test can have multiple snapshots.

PR #754, opened 2022-12-16
[4/x][move-package/lock] Fix error message from parse_package_manifest

amnn:
Fix the order in which the directory and `Move.toml` are joined together (should be directory and then `Move.toml`, not the other way around as it previously was), and don't drop the context of the underlying error.

Also restructuring the function slightly to be less deeply nested, by using `?`.

## Test Plan

```
```

## Stack

- #741 
- #745 
- #753


tnowacki (approved):
4/x looks good!

vgao1996 (approved):

PR #757, opened 2022-12-18
[tools] print imports, fully qualified function names in disassembler

sblackshear:
Previously, the disassembler did not print dependency information or use fully qualified function names. That makes it difficult to see what code a module is using/which functions it's calling, and (in the case of name conflicts) can create ambiguity.

- Add a module alias map to the disassembler state that disambiguates between modules with the same name
- Use the alias map to print dep info, using source langauge syntax. Resolve name conflicts using `as` (again, like the source langguage).
- Also print fully qualified function names. Internal function calls have no module name prefix or `::`, as in the source language.

See added test for details on what the printed output looks like + how conflicts are resolved.


wrwg (approved):

language/tools/move-cli/tests/sandbox_tests/module_publish_view/args.exp
```diff
@@ -4,8 +4,7 @@ Publishing a new module 00000000000000000000000000000042::Module (wrote 120 byte
 Wrote 120 bytes of module ID's and code
 Command `sandbox view storage/0x00000000000000000000000000000042/modules/Module.mv`:
 // Move bytecode v6
-module 42.Module {
-struct S {
+module 42.Module {struct S {
```
amnn, line 7: This seems not right.

language/tools/move-disassembler/src/disassembler.rs
```diff
@@ -61,25 +63,49 @@ pub struct Disassembler<'a> {
...
+        let mut module_names = HashMap::new();
+        let mut module_aliases = HashMap::new();
+        source_mapper
+            .bytecode
+            .self_id()
+            .map(|id| module_names.insert(id.name().as_str().to_string(), 0));
+        for h in source_mapper.bytecode.module_handles() {
+            let id = source_mapper.bytecode.module_id_for_handle(h);
+            let module_name = id.name().as_str().to_string();
+            module_names
```
amnn, line 83: nit: I believe because of `Deref` coercion, you can call `to_string` directly on `Symbol`.

```suggestion
            .self_id()
            .map(|id| module_names.insert(id.name().to_string(), 0));
        for h in source_mapper.bytecode.module_handles() {
            let id = source_mapper.bytecode.module_id_for_handle(h);
            let module_name = id.name().to_string();
            module_names
```

language/tools/move-disassembler/src/disassembler.rs
```diff
@@ -61,25 +63,49 @@ pub struct Disassembler<'a> {
...
 }
 
 impl<'a> Disassembler<'a> {
     pub fn new(source_mapper: SourceMapping<'a>, options: DisassemblerOptions) -> Self {
+        let mut module_names = HashMap::new();
+        let mut module_aliases = HashMap::new();
+        source_mapper
+            .bytecode
+            .self_id()
+            .map(|id| module_names.insert(id.name().as_str().to_string(), 0));
```
amnn, line 79: `map` doesn't seem like the right iterator function to use here, how about `for_each`, or better, `extend` on the HashMap?

```suggestion
        module_names.extend(
            source_mapper.bytecode.self_id().map(|id| (id.name().to_string(), 0)),
        );
```

amnn (approved):

language/tools/move-cli/tests/sandbox_tests/module_publish_view/args.exp
sblackshear: Good call, forgot to update the output here

PR #758, opened 2022-12-18
[verifier] check type instantiations behind a reference or inside vec…

sblackshear:
…tor generics

The verifier current doesn't check ability constraints on types that are behind a reference, or inside a vector (i.e., the `T` in `&T`, `&mut T`, or `vector<T>`). This sounds alarming, but is fortunately not possible to take advantage of because doing so always creates a bad struct instantiation that will be flagged b the ability checker.

Fix the bug in version 7--left it in version 6 and below because it is (to the best of our understanding) harmless and the fix is a breaking change in the verifier--it will reject some function declarations that were previously accepted.

### DIFF
--- a/language/move-bytecode-verifier/src/signature.rs
+++ b/language/move-bytecode-verifier/src/signature.rs
@@ -317,24 +317,40 @@ impl<'a> SignatureChecker<'a> {
 
     fn check_type_instantiation(
         &self,
-        ty: &SignatureToken,
+        s: &SignatureToken,
         type_parameters: &[AbilitySet],
     ) -> PartialVMResult<()> {
-        match ty {
-            SignatureToken::StructInstantiation(idx, type_arguments) => {
-                // Check that the instantiation satisfies the `idx` struct's constraints
-                // Cannot be checked completely if we do not know the constraints of type parameters
-                // i.e. it cannot be checked unless we are inside some module member. The only case
-                // where that happens is when checking the signature pool itself
-                let sh = self.resolver.struct_handle_at(*idx);
-                self.check_generic_instance(
-                    type_arguments,
-                    sh.type_param_constraints(),
-                    type_parameters,
-                )
+        for ty in s.preorder_traversal() {
+            match ty {
+                SignatureToken::StructInstantiation(idx, type_arguments) => {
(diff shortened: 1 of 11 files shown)


language/move-binary-format/src/file_format_common.rs
```diff
@@ -405,8 +405,12 @@ pub const VERSION_5: u32 = 5;
 ///  + u16, u32, u256 integers and corresponding Ld, Cast bytecodes
 pub const VERSION_6: u32 = 6;
 
+/// Version 7: changes compared with version 6
+/// Fixed bug in signature checking pass of verifier--did not do ability checks on the `T` in `&T`, `&mut T`, or `vector<T>`
+pub const VERSION_7: u32 = 7;
+
 // Mark which version is the latest version
-pub const VERSION_MAX: u32 = VERSION_6;
+pub const VERSION_MAX: u32 = VERSION_7;
```
sblackshear, line 413: @vgao1996 @tnowacki: I assume it's too late to get this into version 6, but let me know if not

wrwg: We would love to have this in v6 but it basically needs to land ASAP.

wrwg (approved):

language/move-binary-format/src/file_format_common.rs
wrwg: Though I'm wondering: why would this be a file format change?

We have adapted the general policy that restricting bytecode from doing crazy things is _not_ a breaking change. This hurts us in replay, as @runtian-zhou knows, but it seems rather unpractical to allow old 'malicious/broken/...' code to continue to exist. Runtian has build a tool which scans our published modules for whether they are affected, and we try to get an idea what we are breaking. Our policy is if you earlier successfully published a module and we figure it is a danger for our network, we do not continue to allow it.

wrwg (approved):

language/move-binary-format/src/file_format_common.rs
sblackshear: > We have adapted the general policy that restricting bytecode from doing crazy things is not a breaking change.

This principle makes sense to me. My preference would also be to get rid of the version gating--will fix/land.

PR #759, opened 2022-12-19
[5/x][move-package/lock] Re-use test packages for multiple test kinds

amnn:
Another refactor of the move-package datatest runner, to allow the same test fixtures (move packages) to be re-used for multiple kinds of test (e.g. rather than duplicating `basic_no_deps` for resolution, compilation, and parsing, put the snapshots for all three tests under a single `basic_no_deps` directory at the `test_sources` top-level).

Also:

- cleans up vestigial references to `Move.compile` and `Move.model` which were the previous method for indicating the kind of test being run.
- Gets rid of the separation of tests at the top-level between `compilation`, `model`, `parsing`, and `resolution` given that tests are now shared.

This change is motivated by the fact that many of the resolution tests will also double as lock file tests when lock files are introduced, and this measure avoids further duplication.

## Test Plan

```
```

##  Stack

- #741 
- #745 
- #753 
- #754


tnowacki (approved):
Hooray!

vgao1996 (approved):
Ah, this seems much better organized than 3/x.

PR #760, opened 2022-12-19
[6/x][move-package/lock] LockFile and DependencyGraph

amnn:
Introduce `DependencyGraph` as the in-memory representation of a `BuildConfig`-agnostic transitive dependency graph, and `LockFile` which is a handle for the lock file that this graph can be serialized into and deserialized out of.

The lock file is created at the start of a build as a temporary file in a directory within `build`, and can optionally be committed at a path supplied in `BuildConfig`.  If the explicit `commit` API is not called, its contents will be discarded.  This means in the case of a crash or abort, the lock file will not be left in an inconsistent state, although if this happens and `LockFile`'s `drop` handler is not called, a temporary file may leak in the `build` directory. (This is usually cleaned up by a later "clean" operation in a subsequent run of the compiler).

`lock_file::schema` comes with `serde`-compatible times for deserializing the lock file, reading its version to determine
compatibility, and extracting the serialized dependencies from within it (to be further parsed into `parsed_manifest::Dependency`).

This commit adds `DependencyGraph` and lock file generation as part of package resolution, to write snapshot tests for it, but it does not replace similar logic in `resolution_graph` (for exploring a package's set of transitive dependencies), to come in a later commit.

Because `DependencyGraph` is `BuildConfig`-agnostic, this commit introduces a behaviour change in that a package whose transitive dependency graph is inconsistent in dev-mode (contains conflicting dependencies) will not compile in normal mode.

## Test Plan

New snapshot tests for lock files, and existing CLI and move-package/resolution tests:

```
move/language/tools/move-cli$ cargo nextest
```

## Stack 

- #741 
- #745 
- #753 
- #754 
- #759


language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -0,0 +1,521 @@
...
+            let new_frontier = self.always_deps.insert(package);
+            if !new_frontier {
+                continue;
+            }
+
+            frontier.extend(
+                self.package_graph
+                    .edges(package)
+                    .filter_map(|(_, dep, mode)| {
+                        (mode == &DependencyMode::Always).then_some(dep)
```
tnowacki, line 392: I always find these sort of things easier to read with a `filter` then a `map`, make of that what you will (I don't really care one way or another)

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -0,0 +1,521 @@
...
+                    })
+            );
+        }
+    }
+}
+
+impl<'a> fmt::Display for DependencyTOML<'a> {
```
tnowacki, line 399: Can we not use `toml::ser`? Or maybe that module doesn't do what I think it does?

tnowacki (approved):
Approving with some comments. But this has been a big stack of PRs with just me reviewing, might be good to get some eyes from other stake holders 

Regardless, very excited to see this go through!

language/tools/move-package/tests/test_runner.rs
```diff
@@ -102,34 +102,47 @@ impl Test<'_> {
...
-
-        let Some(ext) = self.expected.extension().and_then(OsStr::to_str) else {
-            bail!("Unexpected snapshot file extension: {:?}", self.expected.extension());
-        };
+        let resolved_package = config.resolution_graph_for_package(self.toml_path, &mut sink);
 
         Ok(match ext {
+            "locked" => fs::read_to_string(&lock_path)?,
+
+            "not-locked" if lock_path.is_file() => {
```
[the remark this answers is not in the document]
amnn: I can't think of an OS that restricts use of `-` or `_` -- sqlite's temporary files append `-shm` and `-journal` etc to the sqlite db's filename (which usually ends in an extension -- `.db` or `.sqlite`).  The only restriction of this kind that I can think of is Windows' 8.3 file names (max 8 characters for file names and 3 for the extension) but that is a DOS holdover and not a real restriction unless you plan on runnig under DOS compatibility mode (I think).

All that being said, happy to change the extension here.

language/tools/move-package/src/resolution/dependency_graph.rs
amnn: Sadly not -- I left a comment on the definitio of `DependencyTOML` as to why:

```
/// Wrapper struct to display a dependency as an inline table in the lock file (matching the
/// convention in the source manifest).  This is necessary becase the `toml` crate does not
/// currently support serializing types as inline tables.
```

The `source` field of a dependency in the lock file matches its dependency configuration in the manifest, and it's an inline table -- the TOML crate has issues with serializing to those, and even a bug for our specific case (inline table, inside a table, inside an array, inside a table) which I need to file an issue for.

language/tools/move-package/src/resolution/lock_file/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+impl DerefMut for LockFile {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        self.file.as_file_mut()
+    }
+}
+
+/// Write the initial part of the lock file.
+fn write_lock_file_prologue(file: &mut NamedTempFile) -> io::Result<()> {
```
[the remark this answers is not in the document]
amnn: Unlike with `DependencyTOML` this might work!  I'll give it a try, the only issue is if `Schema` deriving `Serialize` causes a problem for `Schema<Dependencies>` (because `Dependencies` can't derive `Serialize`).

vgao1996 (approved):
Sorry for the late review. This looks good to me.

PR #764, opened 2022-12-20
[compiler] Fix instantiation of field types with type parameters

### LINKED ISSUE #763: [Bug] compiler fix for bug from 78a9e2b01da9c87af31910bbbcb5eeb4d51c909d
```
module 0x42::Test {
    struct A<T: copy> has copy { a: T }

    struct B<T> has copy {
        data: vector<A<T>>
    }
}
```

will be rejected by the verifier, but is allowed by the compiler

tnowacki:
- Fixed a bug where ability constraints where not properly checked for fields containing the structs type parameter

The bug was only present with type parameters, for example `struct S<T: copy> {} struct B { f: S<signer> }` would be rejected prior to this fix 

## Motivation

Fixes #763

## Test Plan

- New tests. Looked like we had one written already for this but the output was just incorrect....

### DIFF
--- a/language/move-compiler/src/typing/translate.rs
+++ b/language/move-compiler/src/typing/translate.rs
@@ -513,6 +513,16 @@ fn struct_def(context: &mut Context, s: &mut N::StructDefinition) {
         N::StructFields::Defined(m) => m,
     };
 
+    // instantiate types and check constraints
+    for (_field_loc, _field, idx_ty) in field_map.iter() {
+        let loc = idx_ty.1.loc;
+        let inst_ty = core::instantiate(context, idx_ty.1.clone());
+        context.add_base_type_constraint(loc, "Invalid field type", inst_ty.clone());
+    }
+    core::solve_constraints(context);
+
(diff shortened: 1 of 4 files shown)


language/move-compiler/src/typing/translate.rs
```diff
@@ -523,16 +533,14 @@ fn struct_def(context: &mut Context, s: &mut N::StructDefinition) {
     for (_field_loc, _field, idx_ty) in field_map.iter() {
         let loc = idx_ty.1.loc;
         let subst_ty = core::subst_tparams(tparam_subst, idx_ty.1.clone());
-        let inst_ty = core::instantiate(context, subst_ty);
```
tnowacki, line 526: The root of the bug is that `instantiate` adds the ability constraints to the context. But, for abilities we substitute in the Any type before checking ability field requirements. So the type parameters would erroneously satisfy the constraints even when they shouldn't.

The fix is then to instantiate the type for the constraints separately from the Any substitution.

sblackshear (approved):

PR #767, opened 2022-12-23
[stdlib] add `vector::insert`

sblackshear:
We have `vector::remove` (which is very similar), and it's a pain to ask users to handroll this.

### DIFF
--- a/language/move-stdlib/sources/vector.move
+++ b/language/move-stdlib/sources/vector.move
@@ -146,7 +146,27 @@ module std::vector {
         pragma intrinsic = true;
     }
 
-    /// Swap the `i`th element of the vector `v` with the last element and then pop the element.
+    /// Insert `e` at position `i` in the vector `v`.
+    /// If `i` is in bounds, this shifts the old `v[i]` and all subsequent elements to the right.
+    /// If `i == length(v)`, this adds `e` to the end of the vector.
+    /// This is O(n) and preserves ordering of elements in the vector.
+    /// Aborts if `i > length(v)`
+    public fun insert<Element>(v: &mut vector<Element>, e: Element, i: u64) {
+        let len = length(v);
+        // i too big abort
+        if (i > len) abort EINDEX_OUT_OF_BOUNDS;
+
+        push_back(v, e);
+        while (i < len) {
+            swap(v, i, len);
+            i = i + 1
+        }
+    }
+    spec insert {
+        pragma intrinsic = true;
+    }
+
+    /// Swap the `i`th element of the vector `v` with the last element and then pop the vector.
     /// This is O(1), but does not preserve ordering of elements in the vector.
     /// Aborts if `i` is out of bounds.
     public fun swap_remove<Element>(v: &mut vector<Element>, i: u64): Element {
@@ -194,5 +214,4 @@ module std::vector {
             v1[i..len(v1)] == v2[i + 1..len(v2)]
         }
     }
-
 }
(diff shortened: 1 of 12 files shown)


language/move-stdlib/sources/vector.move
```diff
@@ -146,7 +146,26 @@ module std::vector {
...
+    public fun insert<Element>(v: &mut vector<Element>, e: Element, i: u64) {
+        let len = length(v);
+        // i out of bounds; abort
+        if (i >= len) abort EINDEX_OUT_OF_BOUNDS;
+
+        push_back(v, e);
+        swap(v, len, i);
+        while (i < len) {
+            i = i + 1;
+            swap(v, i, len);
```
jasper-c-p-s, line 161: This always does an unnecessary `swap(v, len, len)` at the end. Either decrease by one, or even cleaner:

```rust
push_back(v,e);
while (i < len) {
  swap(v, i, len);
  i = i + 1;
}
```

language/move-stdlib/sources/vector.move
```diff
@@ -146,7 +146,26 @@ module std::vector {
...
     }
 
-    /// Swap the `i`th element of the vector `v` with the last element and then pop the element.
+    /// Insert `e` at position `i` in the vector `v`, shifting the old `v[i]` and all subsequent elements to the right.
+    /// This is O(n) and preserves ordering of elements in the vector.
+    /// Aborts if `i` is out of bounds.
+    public fun insert<Element>(v: &mut vector<Element>, e: Element, i: u64) {
+        let len = length(v);
+        // i out of bounds; abort
+        if (i >= len) abort EINDEX_OUT_OF_BOUNDS;
```
jasper-c-p-s, line 155: This works fine for i = len, any reason to disallow this? Might be unexpected, to be fair.

language/move-stdlib/sources/vector.move
sblackshear: Good catch!

language/move-stdlib/sources/vector.move
sblackshear: Good question--this is certainly semantically coherent, and I actually initially implemented it this way. But I then realized that all of the other vector functions abort if `i` is not in bounds, and thus decided it was simplest to follow the same convention here.

amnn: Conventionally in other languages (I checked Rust, C++, Python and Java) this works for `i = len` so my 2c is to also allow it here.

amnn (commented):
LGTM, modulo the comment about allowing insert at the last position (if it helps, I like to think of `i` as an index into the gaps of the vector, rather than the elements).

Also, I notice the docs for the nursery have disappeared -- is the nursery not long for this world?  Asking because I was using it in a test for lock files.

amnn (approved):
Forgot to stamp.

language/move-stdlib/sources/vector.move
```diff
@@ -146,7 +146,25 @@ module std::vector {
         pragma intrinsic = true;
     }
 
-    /// Swap the `i`th element of the vector `v` with the last element and then pop the element.
+    /// Insert `e` at position `i` in the vector `v`, shifting the old `v[i]` and all subsequent elements to the right.
+    /// This is O(n) and preserves ordering of elements in the vector.
+    /// Aborts if `i` is out of bounds.
+    public fun insert<Element>(v: &mut vector<Element>, e: Element, i: u64) {
```
wrwg, line 152: I would have expected this to be native. Wondering what the gas cost will be for this swap technique, and whether it is acceptable? @vgao1996

wrwg (approved):
Good with it, but isn't this such a fundamental function that it should be native?

sblackshear:
> Good with it, but isn't this such a fundamental function that it should be native?

No strong feelings on this--I used the same implementation approach as `insert` for consistency.

PR #769, opened 2022-12-24
[move-vm] expose metadata to session

davidiw:
In order to look up struct metadata during changeset evaluation for merging storage / resource groups, we need to expose this aspect of the loader to the session.

### DIFF
--- a/language/move-vm/runtime/src/session.rs
+++ b/language/move-vm/runtime/src/session.rs
@@ -16,6 +16,7 @@ use move_core_types::{
     effects::{ChangeSet, Event},
     identifier::IdentStr,
     language_storage::{ModuleId, TypeTag},
+    metadata::Metadata,
     resolver::MoveResolver,
     value::MoveTypeLayout,
 };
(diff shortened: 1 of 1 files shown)


wrwg (commented):
This is already exposed as `vm.get_module_metadata`, which fits the lifetime of the loader. Its also already used for Aptos, check `framework/src/module_metadata.rs`. Error maps and views use this today.

davidiw:
> This is already exposed as `vm.get_module_metadata`, which fits the lifetime of the loader. Its also already used for Aptos, check `framework/src/module_metadata.rs`. Error maps and views use this today.

From my investigation, the best place to put the storage group merge is in the session output to changeset, here: https://github.com/aptos-labs/aptos-core/blob/f19ed79b3c180c59a6102e5deb5a5a8c8d1cc78a/aptos-move/aptos-vm/src/move_vm_ext/session.rs#L104

Passing in a MoveVM there leads to some pretty gnarly, spaghetti code.

Doesn't seem like it hurts to expose this here. It would be nice not having to bounce around move versions, even in the case, where we don't end up using this.

wrwg (approved):
Sure we can expose this. However, that you can't access the VM from a session is pretty broken in the first place, but we can fix this another time...

davidiw:
Pausing on this... realized that we eat the session in that code, so this approach might be useless as well...

PR #773, opened 2022-12-29
Move vector<u64> easy casting to rust Vec<u64>

zjma:
## Motivation

This allows move values of type `vector<u64>` to be easily cast to `Vec<u64>` in rust using the `pop_arg` macro.

Yes.

## Test Plan

UT test case should be sufficient.


vgao1996 (approved):

wrwg:
Notice that if this should be used in Aptos, it needs to be cherry picked into the `aptos-main` branch. In general, it might be more efficient to directly submit against that branch. We will then later cherry pick sets of PRs into the regular `main` branch.

PR #774, opened 2022-12-29
[move-lang] Inlined functions (formerly macros): parsing and type checking (rebased for new aptos branch)

wrwg:
This PR realizes the 1st step towards inlined functions. Inlined functions are expanded at caller site. Moreover, they can take _lambdas_ (functions) as parameters. Example:

```
    public inline fun foreach<T>(v: &vector<T>, action: |&T|) {
        let i = 0;
        while (i < Vector::length(v)) {
            action(Vector::borrow(v, i));
            i = i + 1;
        }
    }

   fun use_foreach() {
        let v = vector[1, 2, 3];
        let sum = 0;
        foreach(&v, |e| sum = sum + *e);
        assert!(sum == 6, 1);
    }

```

This PR implements parsing and type checking of this style of functions up to the typing AST. Actual inlining is not yet implemented.

The biggest change is enabling function types, lambda expressions, and calls of function values for naming and typing. Those had been supported before for the spec language, but weren't supported in move-lang beyond expansion. In the design, we represent function types by a new `BuiltinTypeName_::Fun` which allows to minimize changes to the code, as no new type variant needs to be introduced, and existing functionality for inference et. al can be adopted with minimal changes. This consistent e.g. with the treatment of function types in Rust, where they are just type constructors with some syntactic sugar.

A new test case has been added at [lambda.move](language/move-compiler/tests/move_check/typing/lambda.move) which shows usage and intends to cover all type check failure scenarios.

### DIFF
--- a/language/move-compiler/src/expansion/ast.rs
+++ b/language/move-compiler/src/expansion/ast.rs
@@ -199,6 +199,7 @@ pub struct SpecId(usize);
 pub struct Function {
     pub attributes: Attributes,
     pub loc: Loc,
+    pub inline: bool,
     pub visibility: Visibility,
     pub entry: Option<Loc>,
     pub signature: FunctionSignature,
@@ -416,7 +417,7 @@ pub enum Exp_ {
     While(Box<Exp>, Box<Exp>),
     Loop(Box<Exp>),
     Block(Sequence),
-    Lambda(LValueList, Box<Exp>), // spec only
+    Lambda(LValueList, Box<Exp>),
     Quant(
         QuantKind,
         LValueWithRangeList,
@@ -675,6 +676,10 @@ impl AbilitySet {
 
     pub fn collection(loc: Loc) -> Self {
         Self::from_abilities_(loc, Self::COLLECTION.to_vec()).unwrap()
+    }
+
+    pub fn functions() -> Self {
+        Self::empty()
     }
 }
 
@@ -1231,6 +1236,7 @@ impl AstDebug for (FunctionName, &Function) {
             Function {
                 attributes,
                 loc: _loc,
+                inline,
                 visibility,
                 entry,
                 signature,
@@ -1247,7 +1253,11 @@ impl AstDebug for (FunctionName, &Function) {
         if let FunctionBody_::Native = &body.value {
             w.write("native ");
         }
-        w.write(&format!("fun {}", name));
+        if *inline {
+            w.write(&format!("inline fun {}", name));
+        } else {
+            w.write(&format!("fun {}", name));
+        }
         signature.ast_debug(w);
         if !acquires.is_empty() {
             w.write(" acquires ");
@@ -1330,9 +1340,9 @@ impl AstDebug for Type_ {
                 s.ast_debug(w)
             }
             Type_::Fun(args, result) => {
-                w.write("(");
+                w.write("|");
                 w.comma(args, |w, ty| ty.ast_debug(w));
-                w.write("):");
+                w.write("|");
                 result.ast_debug(w);
             }
             Type_::UnresolvedError => w.write("_|_"),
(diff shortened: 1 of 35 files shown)


language/move-model/src/builder/exp_translator.rs
```diff
@@ -563,6 +563,10 @@ impl<'env, 'translator, 'module_translator> ExpTranslator<'env, 'translator, 'mo
                         U256 => Type::new_prim(PrimitiveType::U256),
                         Vector => Type::Vector(Box::new(self.translate_hlir_base_type(&args[0]))),
                         Bool => Type::new_prim(PrimitiveType::Bool),
+                        Fun => Type::Fun(
+                            self.translate_hlir_base_types(&args[0..args.len() - 1]),
+                            Box::new(self.translate_hlir_base_type(&args[args.len() - 1])),
+                        ),
```
meng-xu-cs, line 569: Wow, smooth integration with the Spec type system!

language/move-compiler/src/parser/syntax.rs
```diff
@@ -1864,7 +1882,7 @@ fn parse_struct_type_parameters(
 
 // Parse a function declaration:
 //      FunctionDecl =
-//          "fun"
+//          [ "macro" ] "fun"
```
meng-xu-cs, line 1885:
```suggestion
//          [ "inline" ] "fun"
```

language/move-compiler/src/expansion/ast.rs
```diff
@@ -1247,7 +1253,11 @@ impl AstDebug for (FunctionName, &Function) {
         if let FunctionBody_::Native = &body.value {
             w.write("native ");
         }
-        w.write(&format!("fun {}", name));
+        if *inline {
+            w.write(&format!("macro {}", name));
```
meng-xu-cs, line 1257:
```suggestion
            w.write(&format!("inline fun {}", name));
```

Or, rather,
```rust
if *inline {
    w.write("inline ");
}
w.write(&format!("fun {}", name));
```

meng-xu-cs (approved):
Awesome PR!

There are some minor leftovers of using `inline fun` vs `macro` consistently, but these can be changed along the way.

I really like the fact that it integrates with the spec type system smoothly (at least so far). Looking forward to more development!

language/move-compiler/src/parser/syntax.rs
```diff
@@ -1718,6 +1727,15 @@ fn parse_type(context: &mut Context) -> Result<Type, Box<Diagnostic>> {
     Ok(spanned(context.tokens.file_hash(), start_loc, end_loc, t))
 }
 
+/// Checks whether the next tokens looks like the start of a type. Must be aligned
+/// with `parse_type`.
+fn is_start_of_type(context: &mut Context) -> bool {
+    matches!(
+        context.tokens.peek(),
+        Tok::LParen | Tok::Amp | Tok::AmpMut | Tok::Pipe | Tok::Identifier
```
meng-xu-cs, line 1735: This might easily go out of sync with `parse_type`, for example, `Tok::NumValue` is also possible as a start of type `0x1::MyModule::MyStruct`.

language/move-compiler/src/parser/syntax.rs
```diff
@@ -1696,7 +1696,16 @@ fn parse_type(context: &mut Context) -> Result<Type, Box<Diagnostic>> {
         }
         Tok::Pipe => {
             let args = parse_comma_list(context, Tok::Pipe, Tok::Pipe, parse_type, "a type")?;
-            let result = parse_type(context)?;
+            let result = if is_start_of_type(context) {
```
meng-xu-cs, line 1699: One possibility of removing this `is_start_of_type` is to call the `parse_type(context)` still and one of the following cases should happen:
- `parse_type` returns `Ok(t)`, then we take `t` as the return type
- `parse_type` returns `Err(_)` **AND** cursor does not advance, then we take `Unit` as a return type
- `parse_type` returns `Err(_)` **AND** cursor does advance, then we just propagate the diagnostic back

language/move-compiler/src/parser/syntax.rs
wrwg: ... which is what the comment also says. Feels we do not change the syntax that often that this would be an issue.

PR #778, opened 2022-12-29
[compiler] warn on unused `&mut` qualifier in parameters

sblackshear:
Add new warning type: unused `&mut` qualifier. A qualifier is considered unused if the corresponding variable is not:
- mutated
- used in an expression that mutably borrows a field
- passed as an argument to a call that requires a parameter of `&mut` type
- returned from a function with a `&mut` return type

Added tests for each of these cases.

- Note 1: there are 45 existing compiler tests that are failing because they now (correctly) get flagged with unused `&mut` warnings. My inclination is to to update the expected values rather than modifying the tests, but let me know if that doesn't sound right.
- Note 2: I added this warning inside translate.rs because that's where very similar unused variables warning lives, but the file is perhaps getting a bit crowded. Maybe both of these warnings want to be separate passes in the future, but I'm not sure.

### DIFF
--- a/language/move-compiler/src/diagnostics/codes.rs
+++ b/language/move-compiler/src/diagnostics/codes.rs
@@ -218,6 +218,7 @@ codes!(
         DeadCode: { msg: "dead or unreachable code", severity: Warning },
         StructTypeParam: { msg: "unused struct type parameter", severity: Warning },
         Attribute: { msg: "unused attribute", severity: Warning },
+        MutQualifier: { msg: "unused &mut qualifier", severity: Warning }
     ],
     Attributes: [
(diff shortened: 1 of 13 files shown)


wrwg (commented):
I think we need to come up with a way to quiet such warnings (via attributes like in Rust?), because changing mutability of a parameter is a breaking change IIRC, and many projects treat warnings similar as errors.

sblackshear:
> I think we need to come up with a way to quiet such warnings (via attributes like in Rust?), because changing mutability of a parameter is a breaking change IIRC, and many projects treat warnings similar as errors.

Agreed--these annotations are long overdue. Can change this to allow suppression with `_` for now (as we do for unused vars), but something more fine-grained would be nice.

PR #779, opened 2022-12-29
Test for instantiation operation

dariorussi:
## Motivation
Test performance of `Resolver` and `Interpreter` instantiation API.
Value will be tuned according to limits and other implementation details

Yes

## Test Plan
This is it

### DIFF
--- a/language/move-vm/integration-tests/src/tests/instantiation_tests.rs
+++ b/language/move-vm/integration-tests/src/tests/instantiation_tests.rs
@@ -1,0 +1,1279 @@
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![allow(dead_code)]
+#![allow(unused_must_use)]
+#![allow(unused_imports)]
+
+use move_binary_format::{
+    errors::VMResult,
+    file_format::{
+        AbilitySet, AddressIdentifierIndex, Bytecode, Bytecode::*, CodeUnit, CompiledModule,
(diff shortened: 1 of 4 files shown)


language/move-vm/integration-tests/src/tests/instantiation_tests.rs
```diff
@@ -0,0 +1,846 @@
...
+fn test_instantiation_vec_pack_gen_simple() {
+    let (res, ref_time) = run_with_module(&mut get_gas_meter(1000), load_pop);
+    assert_eq!(
+        res.err().unwrap().major_status(),
+        StatusCode::OUT_OF_GAS,
+        "Must finish OutOfGas"
+    );
+    let (res, time) = run_with_module(&mut get_gas_meter(1000), vec_pack_gen_simple);
+    println!("* vec_pack_gen_simple: {}\nStatus: {:?}", time, res);
+    assert_eq!(
```
sblackshear, line 122: Like this approach! 

Wondering if these tests could become even easier to write with helper function around `run_with_module` that:
- runs the "control" gas check with `load_pop`
- runs the "experiment" gas check with the function provided by the user (e.g., `vec_pack_gen_simple` here
- checks that both control and experiment finish with `OUT_OF_GAS` and the experiment is <10x slower than the control

and of course could parameterize things like the choice of control function, gas budget, 10x threshold.

dariorussi: done, I think. Though I am going to add few more things before I make it a regular PR (non draft)

sblackshear (approved):

PR #783, opened 2022-12-31
[move-prover] do not assume data invariants for `&mut` params

### LINKED ISSUE #780: [Prover][Bug] incorrectly assumed the data invariant of a mutable reference of struct
## How to reproduce:

Run Prover the following example to reproduce the bug:
```
module 0x42::struct_invariant_bug {
    use std::vector;

    struct S {
        v: vector<u64>,
    }
    spec S {
        invariant len(v) == 0;
    }

    public fun empty(): S {
        S { v: vector::empty<u64>() }
    }

    public fun push_1(s: &mut S) {
        spec {
            assert len(s.v) == 0;
        };
        vector::push_back(&mut s.v, 1);
    }
}
```
Prover currently produces no error for this example.

## Expected result
Prover fails on the inlined spec assert.

meng-xu-cs:
Data invariants are only checked when the `&mut` borrow ends. During the lifetime of the `&mut` reference, the invariant can be temporarily violated and hence, assuming data invariants there can lead to inconsistency issues.

## Motivation

Fixes #780 

Yes

## Test Plan

- CI
- A new regression test

### DIFF
--- a/language/move-prover/tests/sources/regression/data_invariant_for_mut_ref_arg.exp
+++ b/language/move-prover/tests/sources/regression/data_invariant_for_mut_ref_arg.exp
@@ -1,0 +1,10 @@
+Move prover returns: exiting with verification errors
+error: unknown assertion failed
+   ┌─ tests/sources/regression/data_invariant_for_mut_ref_arg.move:22:13
+   │
+22 │             assert len(s.v) == 0;
+   │             ^^^^^^^^^^^^^^^^^^^^^
+   │
+   =     at tests/sources/regression/data_invariant_for_mut_ref_arg.move:15: push_1
+   =         s = <redacted>
+   =     at tests/sources/regression/data_invariant_for_mut_ref_arg.move:22: push_1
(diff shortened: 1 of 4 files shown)


language/move-prover/tests/sources/regression/data_invariant_for_mut_ref_arg.exp
```diff
@@ -0,0 +1,10 @@
+Move prover returns: exiting with verification errors
+error: unknown assertion failed
```
rahxephon89, line 2: I am wondering whether we could generate more concrete error message, e.g. the assertion fails because `s` might not be in a consistent state?

rahxephon89 (approved):
LGTM, thanks!

language/move-prover/tests/sources/regression/data_invariant_for_mut_ref_arg.exp
meng-xu-cs: This is going to be hard... This falls in the line of explaining why an assertion fail (e.g., due to a missing assumption), which is theoretically intractable.

meng-xu-cs:
Unable to get around the current modeling of `Options.move`, dropping this PR in favor of #792 for a better solution.

PR #784, opened 2023-01-02
[8/x][move-package/lock] DependencyGraph::immediate_dependencies

amnn:
Function to return a package's immediate dependencies in the transitive dependency graph, for use by `ResolutionGraph` when resolving renamings and in-scope addresses during resolution

## Test Plan

New tests:

```
move/language/tools/move-package$ cargo nextest
```

## Stack

- #741 
- #745 
- #753 
- #754 
- #759 
- #760 
- #765

PR #785, opened 2023-01-02
[9/x][move-package/lock] ResolvingTable

amnn:
Extract the logic for unifying named addresses across address assignments, substitutions and renamings from `ResolvingGraph` into its own class -- `ResolvingTable`.

This is to be used when integrating `DependencyGraph` into `ResolutionGraph`.

## Test Plan

New unit tests:

```
```

## Stack

- #741 
- #745 
- #753 
- #754 
- #759 
- #760 
- #765 
- #784


language/tools/move-package/src/resolution/resolving_table.rs
```diff
@@ -0,0 +1,187 @@
...
+
+    /// Iterates over the bindings in this table that are within `pkg`'s scope.
+    pub fn bindings(
+        &self,
+        pkg: PackageName,
+    ) -> impl Iterator<Item = (NamedAddress, &Option<AccountAddress>)> {
+        let start = (pkg, NamedAddress::from(""));
+        self.redirection
+            .range(start..)
+            .take_while(move |((scope, _), _)| *scope == pkg)
```
tnowacki, line 50: Trying to understand this, as I've never used `BtreeMap::range` before but it looks nifty!
Is this more different than `self.redirection.filter(|((scope, _), _)| *scope == pkg)`?

language/tools/move-package/src/resolution/resolving_table.rs
```diff
@@ -0,0 +1,187 @@
...
+    assignments: Vec<Assignment>,
+
+    /// Mapping named addresses to an entry in the `assignments` table.
+    redirection: BTreeMap<QualifiedAddress, usize>,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+enum Assignment {
+    Assign(Option<AccountAddress>),
+    Linked(usize),
```
tnowacki, line 30: IIRC correctly, this was previously done with pointers (`Ref` or something similar)? Just checking to make sure I'm following 

Regardless, +1 to this sort of flattened design. Maybe it's too much time in type systems, but I always find these sort of custom index solutions a lot less error prone :)

language/tools/move-package/src/resolution/resolving_table.rs
```diff
@@ -0,0 +1,187 @@
...
+
+        let mut chain = vec![c];
+        while let Assignment::Linked(ggp) = self.assignments[gp] {
+            (c, p, gp) = (p, gp, ggp);
+            chain.push(c);
+        }
+
+        for link in chain {
+            self.assignments[link] = Assignment::Linked(gp);
+        }
```
tnowacki, line 146: I'm not sure I understand why the updates are done here? Are they necessary or just an optimization?

language/tools/move-package/src/resolution/resolving_table.rs
```diff
@@ -0,0 +1,187 @@
...
+            Ordering::Equal => None,
+
+            Ordering::Greater => {
+                let (jr, ir) = self.get_mut_pair(jx, ix)?;
+                Some((ir, jr))
+            }
+
+            Ordering::Less => {
+                let (l, r) = self.assignments.split_at_mut(ix + 1);
+                Some((&mut l[ix], &mut r[jx - ix - 1]))
```
tnowacki, line 164: wild... this is a fun trick to keep up your sleeve for accessing multiple elements of a vector at once

language/tools/move-package/src/resolution/resolving_table.rs
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /// Add the binding `a = b` to the table.  Fails if this introduces a contradiction (A path
+    /// through bindings between two account addresses that are unequal to each other), and succeeds
+    /// otherwise.
+    pub fn unify(&mut self, a: QualifiedAddress, b: QualifiedAddress) -> Result<()> {
+        let ix = self.get_or_create_assignment(a);
+        let jx = self.get_or_create_assignment(b);
+
+        let Some((ir, jr)) = self.get_mut_pair(ix, jx) else {
```
tnowacki, line 97: While it is a neat trick. I wonder if it would just be simpler to just do this all immutably, and then borrow at the point of modification? 

Like sure... you will have to index into the vector twice, but does that ever matter perf wise? 
```
if ix == jx { return Ok(()) }
let Assignment::Assign(ia) = ir else { unreachable!("Non-root assignment"); };
let Assignment::Assign(ja) = jr else { unreachable!("Non-root assignment"); };
match (ia, ja) {
    (None, Some(_)) => self.assignment[ix] = Assignment::Linked(jx),
    (Some(_), None) | (None, None) => self.assignment[ir] = Assignment::Linked(ix),
    (Some(ia), Some(ja)) => {
        if ia != ja {
            return Err(unification_error(a.1, *ia, *ja));
        }
    }
};
```

tnowacki: In other words, my worry is more about the understandability of this code long term. Took me a bit to see what was going on is all.

language/tools/move-package/src/resolution/resolving_table.rs
amnn: It's different from a performance perspective -- range will do a lookup `lg(N)` lookup to find the first element and then continue from there (so if there are `K` entries for a particular package and `N` overall, it will be `O(lg(N) + K)` vs filter's `O(lg(N))`.

language/tools/move-package/src/resolution/resolving_table.rs
amnn: Indeed, this used to be done with an `Rc<RefCell<Option<AccountAddress>>>`

language/tools/move-package/src/resolution/resolving_table.rs
amnn: Just an optimisation, it's path compression just like for union find, it is a trick I learnt from one of Oleg Kiselyov's articles about implementing Algorithm W, a long time ago: https://okmij.org/ftp/ML/generalization.html (search for "path compression").

language/tools/move-package/src/resolution/resolving_table.rs
amnn: You're right -- I think I got bit by the borrow checker and then tried to steer well clear of it and went too far.

tzakian (approved):

PR #788, opened 2023-01-02
[12/x][move-package/lock] Use DependencyGraph in download_dependency_repos

amnn:
Another occasion where the previous implementation recursively traversed the transitive dependency graph -- the new approach discovers the graph and then iterates through its nodes to download them.

The second step is currently technically unnecessary, but becomes relevant in the presence of externally resolved dependencies where third-party resolvers can introduce sub-graphs into the `DependencyGraph` without having to explore them locally.

## Test Plan

```
```


amnn:
There are a couple of potential future improvements to package fetching that I haven't planned for this stack, but are probably worth doing in the limit (I will create issues for these):

- Centralise the management of fetching/updating remote repositories to avoid multiple updates in a single build.  Currently a remote dependency might get fetched once while exploring the DependencyGraph, and then again when converting it into a ResolvedGraph.  The second fetch/update will be quick, but is unnecessary nevertheless (and it means that the progress output shows two fetches, which is annoying and potentially confusing).
- Always commit the lock file and rely on reading that to get a dependency's transitive dependencies (rather than recursively traversing manifest files).

Issue #781, opened 2022-12-29
[Prover][Bug] Asserting data invariants is not enforced when mutable references are frozen.


junkil-park:
## How to reproduce:
Run Prover for the following example to reproduce the bug:
```
module 0x42::struct_invariant_bug {
    use std::vector;

    struct S {
        v: vector<u64>,
    }
    spec S {
        invariant len(v) == 0;
    }

    public fun empty(): S {
        S { v: vector::empty<u64>() }
    }

    public fun push_1(s: &mut S) {
        vector::push_back(&mut s.v, 1);
        let t = freeze(s);
        let _ = vector::length(&t.v);
    }
}
```
Prover currently produces no errors.

## Expected output
Prover fails on proving the struct invariant because the invariant should be enforced when freezing `s` in the function `push_1`.

PR #792, opened 2023-01-03
[move-prover] data invariant assume and assertion for `&mut` params

meng-xu-cs:
For functions with `&mut` params `s: S`, the data invariant for `S` will be assumed at function entry (via `Wellformedness` - this is already done) and re-checked on any write-back to this reference --- this is newly added in this commit.

This PR is the opposite of PR #783 which does the opposite, i.e., drop the assumption at the function entry. PR #783 is proposed following the following guideline:

> Data invariants are only checked when the `&mut` borrow ends. During
> the lifetime of the `&mut` reference, the invariant can be temporarily
> violated and hence, assuming data invariants there can lead to
> inconsistency issues.

In theory, both PR's fix are better than the current practice (which is essentially assuming without assertion).

However, as the test result in PR #783 shows, the `Options.move` in `stdlib` heavily relies on the entrypoint assumption that a `&mut Option` has either 0 or 1 entry in its vector. Therefore, in this commit, we tighten the conditions where we assert data invariant.

In summary, after this PR, data invariants will be checked:

- on write-backs to local root, arised from `let r = &mut s` where `s` is a local variable
- on write-backs to global root, arised from `let r = borrow_global_mut<S>(addr)`
- on write-backs to `&mut` parameter.

```move
struct S {
    f: u8
}
spec S {
    invariant f != 0;
}

fun foo(s: &mut S) {
    // spec: assume s.f != 0;

    let r = &mut s.f;
    r = 0;

    // spec: write-back[s.f](r);
    // spec: assert s.f != 0;
}
```

## Motivation

Fixes both #780 and #781

Yes

## Test Plan

- CI
- Two new test cases adapted from the issues are added

### DIFF
--- a/language/documentation/examples/diem-framework/move-packages/experimental/sources/Vote.move
+++ b/language/documentation/examples/diem-framework/move-packages/experimental/sources/Vote.move
@@ -329,6 +329,7 @@ module ExperimentalFramework::Vote {
         let removed_ballots = vector::empty();
         while ({
             spec {
+                invariant unique_ballots(ballots);
                 invariant no_expired_ballots(ballots, DiemTimestamp::spec_now_seconds(), i);
                 invariant vector_subset(ballots, old(ballot_data).ballots);
                 invariant i <= len(ballots);
--- a/language/move-prover/bytecode/src/memory_instrumentation.rs
+++ b/language/move-prover/bytecode/src/memory_instrumentation.rs
@@ -273,28 +273,25 @@ impl<'a> Instrumenter<'a> {
 
                 // issue a chain of write-back actions
                 for action in chain {
-                    // decide if we need a pack-ref (i.e., data structure invariant checking)
-                    if matches!(
-                        action.dst,
-                        BorrowNode::LocalRoot(..) | BorrowNode::GlobalRoot(..)
-                    ) {
-                        // On write-back to a root, "pack" the reference i.e. validate all its invariants.
-                        let ty = &self
-                            .builder
-                            .get_target()
-                            .get_local_type(action.src)
-                            .to_owned();
-                        if self.is_pack_ref_ty(ty) {
-                            self.builder.emit_with(|id| {
-                                Bytecode::Call(
-                                    id,
-                                    vec![],
-                                    Operation::PackRefDeep,
-                                    vec![action.src],
-                                    None,
-                                )
-                            });
+                    // decide if we need a pre-writeback pack-ref (i.e., data structure invariant checking)
+                    let pre_writeback_check_opt = match &action.dst {
+                        BorrowNode::LocalRoot(..) | BorrowNode::GlobalRoot(..) => {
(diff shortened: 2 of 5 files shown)


junkil-park (commented):
Thanks! I have a clarification question.

> In summary, after this PR, data invariants will be checked:
> 1. on write-backs to local root, arised from let r = &mut s where s is a local variable
> 2. on write-backs to global root, arised from let r = borrow_global_mut<S>(addr)
> 3. on write-backs to &mut parameter.

I am wondering what rules has already existed and what rules are added in this PR?

Also, in the test cases in `data_invariant_for_mut_ref_arg.move`, what rule causes `push_1` and `push_2` to fail, respectively?

meng-xu-cs:
- Rule 1 and  2 already exist and are correctly enforced.
- Rule 3 is introduced in this PR

Rule 3 is the rule that causes `push_1` and `push_2` fail in the newly introduced test case.

junkil-park:
> Rule 3 is the rule that causes `push_1` and `push_2` fail in the newly introduced test case.

In `push_1`, does the write-back to the `&mut` parameter happen right after exiting the function call for `push_back`?

```
struct S {
    f: u8
}
spec S {
    invariant f != 0;
}

fun foo(s: &mut S) {
  s.f = 0;
}
```
In the example above, the write-back will happen after the line `s.f = 0` because it uses an implicit mutable reference?

meng-xu-cs:
```move
struct S {
    f: u8
}
spec S {
    invariant f != 0;
}

fun foo(s: &mut S) {
  s.f = 0;
  // no write-back as `s` is still alive
  s.f = 1;
  // no write-back as `s` is still alive
  s.f = 2;
  // write-back here as `s` is destroyed
  // data invariant asserted here, immediately after the write-back
}
```

junkil-park:
I see. Just to make sure, the data invariant is not asserted in `foo` because `s` is not destroyed?
```
struct S {
    f: u8
}
spec S {
    invariant f != 0;
}

fun foo(s: &mut S): &mut S {
  s.f = 0;
  s.f = 1;
  s.f = 2;
  s
}
```

meng-xu-cs:
I think the data invariant should be instrumented as the return value is assigned to a new variable and hence, s is still considered as destroyed. But this can be confirmed by adding a test case and see the results, I'll add one to confirm.

junkil-park:
That would be nice. Thanks!
By the way, I am curious if the write-back happening when a mutable reference is destroyed was already there before this PR? Or, is this newly added to this PR for the sake of enforcing data invariants?

meng-xu-cs:
It's already happening before this PR. This can been seen in the diff as well.

junkil-park (approved):

meng-xu-cs:
Added two test cases @junkil-park suggested in the comments, which also helps with the illustration on when write-backs happens.

PR #793, opened 2023-01-03
Cherry picks TypeTag bounds from old aptos branch (#791)

wrwg:
* [types] make type tag nesting limited

There's limited value and a lot of wasted resources by leaving this limited to serde's limits. Recursion is dangerous.

* [types] limit nesting of types via state (#532)

Cherry picked from #529.

This is an alternative version of what #528 and the previous for limitation of nesting provides, having a significant lower footprint.

* [types] limit parsing of nested TypeTag

There's limited value and a lot of wasted resources by leaving this limited to serde's limits. Recursion is dangerous.

Co-authored-by: David Wolinsky <isaac.wolinsky@gmail.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2819,9 +2819,9 @@ checksum = "7ffc5c5338469d4d3ea17d269fa8ea3512ad247247c30bd2df69e68309ed0a08"
 
 [[package]]
 name = "memchr"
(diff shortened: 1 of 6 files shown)


vgao1996 (approved):
These branches are surely confusing!

PR #795, opened 2023-01-04
[inlining] Expansion of inlined functions

wrwg:
This PR implements the core logic for expanding inlined functions. Expansion is happening as a new phase after typing and before hlir lowering.  The inlining phase is a function `f(&mut typing::ast::Program)`. It will inline functions as they are encountered. When a function with lambda arguments is inlined, those lambdas are further expanded, using basically the same algorithm as for conventional inlining.

There are a few tests added which cover basic behavior. More tests have to be added as the feature evolves, and more type checks have to be added. Specifically, currently inlining can create code which does not pass the bytecode verifier because 'private' calls are inserted at inlining context. This is not captured by the Move compiler, but the bytecode verifier will flag this. (With a terrible DevX.)

### DIFF
--- a/language/move-compiler/src/command_line/compiler.rs
+++ b/language/move-compiler/src/command_line/compiler.rs
@@ -8,7 +8,7 @@ use crate::{
     compiled_unit,
     compiled_unit::AnnotatedCompiledUnit,
     diagnostics::{codes::Severity, *},
-    expansion, hlir, interface_generator, naming, parser,
+    expansion, hlir, inlining, interface_generator, naming, parser,
     parser::{comments::*, *},
     shared::{
         CompilationEnv, Flags, IndexedPackagePath, NamedAddressMap, NamedAddressMaps,
@@ -56,9 +56,10 @@ pub const PASS_PARSER: Pass = 1;
 pub const PASS_EXPANSION: Pass = 2;
 pub const PASS_NAMING: Pass = 3;
 pub const PASS_TYPING: Pass = 4;
-pub const PASS_HLIR: Pass = 5;
-pub const PASS_CFGIR: Pass = 6;
-pub const PASS_COMPILATION: Pass = 7;
+pub const PASS_INLINING: Pass = 5;
+pub const PASS_HLIR: Pass = 6;
+pub const PASS_CFGIR: Pass = 7;
+pub const PASS_COMPILATION: Pass = 8;
 
 #[derive(Debug)]
 enum PassResult {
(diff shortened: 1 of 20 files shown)


wrwg:
FYI @sblackshear @tnowacki

language/move-compiler/src/inlining/translate.rs
```diff
@@ -0,0 +1,476 @@
...
+    }
+}
+
+fn get_params_from_decls(decls: &LValueList) -> Vec<Symbol> {
+    decls
+        .value
+        .iter()
+        .flat_map(|lv| match &lv.value {
+            LValue_::Var(v, _) => vec![v.0.value],
+            //_ => panic!("ICE unexpected parameter {:?}", lv.value),
```
meng-xu-cs, line 469:
```suggestion
```

meng-xu-cs (approved):
Thank you for this effort! This is a great feature and very nice implementation!

As mentioned in the PR desc, the test cases are a bit thin but this PR itself is a good foundation for iteration so stamping to unblock future (and concurrent) effort.

language/move-compiler/src/command_line/compiler.rs
```diff
@@ -56,16 +56,18 @@ pub const PASS_PARSER: Pass = 1;
...
+pub const PASS_CFGIR: Pass = 7;
+pub const PASS_COMPILATION: Pass = 8;
 
 #[derive(Debug)]
 enum PassResult {
     Parser(parser::ast::Program),
     Expansion(expansion::ast::Program),
     Naming(naming::ast::Program),
     Typing(typing::ast::Program),
+    Inlining(typing::ast::Program),
```
meng-xu-cs, line 70: A style consistent to the current compiler implementation would be to define a new `struct inlining::ast::Program` with pretty much everything similar to `struct typing::ast::Program` (and the associated structs) except in `enum Exp` where there is no `Lambda`-related variants.

However, the above style might not be the most time-economical way of implementing a transformation pass (in contrast to a compilation phase). Personally I am fine with either approaches and this PR proved its point on how succinctly this feature can be implemented without duplicating much of data structure (and the associated translation logic).

wrwg: This is a good point. However, I believe even a strict interpretation of this design would still entail to reuse an AST between two phases when there is no significant change in the AST structure (or normal form) of the program. This is the case for inlining. The only difference in the two ASTs is that after the inlining pass, no functions marked as 'inline' will be anymore in the AST. Wondering whether @tnowacki has an opinion here. Still I think I go forward with this.

PR #798, opened 2023-01-05
Define configurable limits for struct and function instantiations

dariorussi:
## Motivation
**Disclaimer**: most of the code in this PR is test code in a new file. It should be pretty straightforward to review the changes otherwise.

This PR makes instantiation limits for functions and types driven by config. And it also fixes a problem with the previous hard-coded solution. The code before counted the total number of "nodes" among the type to be instantiated and the instantiation "context" which leads to inconsistent behavior and it only partially controls a limit.
Consider the following code
```
struct S<T, W, Z> {...}
fun f<T> { let s: S<T, T, T> }
```
an instantiation of `f<T>` with a "deep" type (say `vector<vector<vector<T>>>`) would account for 8 nodes in the previous code instead of 13 which is what the type would be instantiated into. Bottom line depending on the combination of type, function and signature, some type could be created in some context and not in others.
This PR should unify the behavior and correctly check that either a function or type instantiation is "capped" to the limit defined in the config.

@wrwg @runtian-zhou @vgao1996 @rahxephon89 
Aptos VM implementation has checks also in `subst()` (besides the one in function and type instantiation) which also partially covers for this issue, making type instantiation controlled to some extent. This code should cover for limits better and more consistently **but** it will create a conflict when merging in either direction. I hope that is not too much hassle and we can converge to this solution.

Yes

## Test Plan
Added a unit test

### DIFF
--- a/language/move-vm/integration-tests/src/tests/limits_tests.rs
+++ b/language/move-vm/integration-tests/src/tests/limits_tests.rs
@@ -1,0 +1,2788 @@
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+#![allow(dead_code)]
+#![allow(unused_must_use)]
+#![allow(unused_imports)]
+
+use move_binary_format::{
+    errors::VMResult,
+    file_format::{
+        AbilitySet, AddressIdentifierIndex, Bytecode, Bytecode::*, CodeUnit, CompiledModule,
+        Constant, ConstantPoolIndex, FieldDefinition, FunctionDefinition, FunctionHandle,
+        FunctionHandleIndex, FunctionInstantiation, FunctionInstantiationIndex, IdentifierIndex,
+        ModuleHandle, ModuleHandleIndex, Signature, SignatureIndex, SignatureToken,
+        SignatureToken::*, StructDefInstantiation, StructDefInstantiationIndex, StructDefinition,
+        StructDefinitionIndex, StructFieldInformation, StructHandle, StructHandleIndex,
+        StructTypeParameter, TypeSignature, Visibility::*,
+    },
+};
+use move_bytecode_verifier::{verify_module, VerifierConfig};
+use move_core_types::{
+    account_address::AccountAddress,
+    identifier::{IdentStr, Identifier},
+    language_storage::{ModuleId, StructTag, TypeTag},
+    vm_status::StatusCode,
+};
+use move_vm_runtime::{
+    config::VMConfig,
+    move_vm::MoveVM,
+    session::{SerializedReturnValues, Session},
+};
+use move_vm_test_utils::{
+    gas_schedule::{Gas, GasStatus, INITIAL_COST_SCHEDULE},
+    InMemoryStorage,
+};
+use move_vm_types::loaded_data::runtime_types::Type;
+use once_cell::sync::Lazy;
+use std::time::Instant;
+
+const MODULE_NAME: &str = "Mod";
+const OUTER_NAME: &str = "Outer";
+const INNER_NAME: &str = "Inner";
+const FIELD_NAME: &str = "inner";
+const ENTRY_POINT_NAME_1: &str = "entry_point";
+const ENTRY_POINT_NAME_2: &str = "entry_point_one_ty_arg";
+const ENTRY_POINT_NAME_3: &str = "entry_point_mul_ty_args";
+
+const RECURSIVE_NAME: &str = "recursive";
+const EMPTY_NAME: &str = "empty";
+
+static DEFAULT_SIGNATURES: Lazy<Vec<Signature>> =
+    Lazy::new(|| vec![Signature(vec![]), Signature(vec![TypeParameter(0)])]);
+
+#[test]
+fn test_limit_vector() {
+    let verifier = VerifierConfig {
+        max_type_instantiation_size: Some(128),
+        ..Default::default()
+    };
+    // instantiated tests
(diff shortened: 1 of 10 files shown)


language/move-vm/runtime/src/loader.rs
```diff
@@ -753,6 +753,17 @@ impl Loader {
         self.verify_ty_args(func.type_parameters(), &type_arguments)
             .map_err(|e| e.finish(Location::Module(module_id.clone())))?;
 
+        if !func.type_parameters.is_empty() {
+            if let Some(max_count) = self.vm_config.verifier.max_function_instantiation_size {
+                let mut type_params = vec![];
+                for idx in 0..func.type_parameters.len() {
+                    type_params.push(Type::TyParam(idx))
+                }
```
tnowacki, line 761: For fancy pants points, consider `let type_params = (0..func.type_parameters.len()).map(Type::TyParam).collect::<Vec<_>>();`

language/move-vm/runtime/src/loader.rs
```diff
@@ -1367,6 +1378,58 @@ impl Loader {
...
+    ) -> PartialVMResult<usize> {
+        let mut type_params_count = vec![0; ty_args.len()];
+        let mut count_inst = self.count_type_nodes_and_params(type_, &mut type_params_count)?;
+        if count_inst > max_count {
+            return Err(PartialVMError::new(StatusCode::TOO_MANY_TYPE_NODES));
+        }
+        for (idx, count) in type_params_count.into_iter().enumerate() {
+            if count > 0 {
+                count_inst += count
+                    * match ty_args.get(idx) {
```
tnowacki, line 1400: `* match` with this big, multi-line match made me double take. Totally fine after reading it, but might be more clear if it was bound to its own local. 
Not a big deal either way, just wanted to share that (it is also a bit late for me fwiw... so maybe thats part of it)

language/move-vm/runtime/src/loader.rs
```diff
@@ -2535,6 +2602,51 @@ impl Loader {
...
+                Type::TyParam(idx) => match type_params.get_mut(*idx) {
+                    Some(value) => *value += 1,
+                    None => {
+                        return Err(PartialVMError::new(
+                            StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR,
+                        )
+                        .with_message(format!("no type param {} for {:?}", idx, type_)));
+                    }
+                },
+                _ => {
```
tnowacki, line 2640: nit, exhaustive match here. Helps a lot when doing refactors :)

language/move-vm/integration-tests/src/tests/limits_tests.rs
```diff
@@ -0,0 +1,2788 @@
...
+
+#[test]
+fn test_limit_vector() {
+    let verifier = VerifierConfig {
+        max_type_instantiation_size: Some(128),
+        ..Default::default()
+    };
+    // instantiated tests
+    let res = run_with_module(&verifier, vector_instantiated_2_nodes);
+    res.expect("max_type_instantiation_size(128) vector_instantiated_2_nodes failed");
```
tnowacki, line 62: I worry a bit about maintaining these tests. I also worry for auditing in that I find it hard to grok what tests are here. Thoughts on these issues? Do you share any concerns here? 

I think we could cook up a Rust macro (just a normy one, nothing procedural) for stamping out these cases. What would you think about that? Or do you have anything else in mind?

language/move-vm/runtime/src/loader.rs
dariorussi: very true and I may do that, I was just so sick of it after my 14th attempt....

language/move-vm/integration-tests/src/tests/limits_tests.rs
dariorussi: yeah those tests are a workbench for someone that wanted to look into building some semi-automatic test generation around instantiation. Which ruled and rules out the macros for now.
In other words, I want a simple abstraction over the instructions that do instantiations but I did not know exactly what that would look like. So all those tests are an attempt to understand the different scenarios. And they are all still there because I am still thinking about them :) 
There are several ways they can be unified, still thinking about how to break it up. And allow generation of test case with proptest and/or fuzzying.
So super happy to have deep conversations about this if they move it forward faster. Otherwise wear your seatbelt and let's move on.
I was surprised that I could build so simple test case (in terms of instruction sequences) for complex instantiations, and I am still thinking if they cover all we want.
Macros would set things in stone too much and make you lose focus on the real problem you are trying to solve. They still may be there eventually.

dariorussi:
addressed some issues and replied to comments

language/move-vm/runtime/src/loader.rs
dariorussi: done

language/move-vm/integration-tests/src/tests/limits_tests.rs
dariorussi: maybe it is also worth to point out they are testing all generic opcodes in different instantiations tests.
I know it does not help a lot, but I am hoping they are reasonably well covered and exposed real issues.

dariorussi: and finally, the instantiation tests of the other PR is also something that will follow a final re-org of the instantiation testing code.

I think finally we should have something comprehensible to all

dariorussi:
moving it back to draft, so I can work on few things and resolve the conflicts as well

PR #807, opened 2023-01-09
[inlining] Adding combinators to vector + fixing bugs

wrwg:
This fixes a few more bugs in inlining discovered while adding foreach/fold/map/filter to `vector.move`. With this, test coverage of this feature should be fairly complete. The new vector (and option) functions will go into an AIP.


wrwg:
> Sorry to off-topic a bit, but I can't do `public inline fun hi( f: ||) {f();}` that f is return for none and no argument going be pass though, any solution about that? what if `f` does return value? how write the syntax to pass the compiler?

Try `| |` instead, and please open an issue. Note that this feature is unstable.

wrwg:
Closing for an alternative implementation

PR #808, opened 2023-01-09
[movey] Scrub integrations into Movey APIs from Move codebase

amnn:
With the introduction of external package resolvers in #790, it is no longer necessary to embed calls to Movey's package resolution and authentication API into the Move CLI.

Instead, Movey will maintain its own CLI that gets called during resolution, and that tool can communicate with Movey's servers and deal with Movey authentication.

## Test Plan

```
$ cargo x lint && cargo xfmt && cargo xclippy --all-targets
$ cargo-nextest nextest run
$ cargo-nextest nextest run --features evm-backend
```


sblackshear (approved):

sblackshear:
yay red diffs

gregnazario (approved):

PR #810, opened 2023-01-10
Aw/prover custom memory

awelc:
**Please note that this is very much a draft at this point.**

## Motivation

This PR aims to allow other Move flavors to use the Prover despite differences in the memory model. For an example on how this extension can be used please see this PR in the Sui repository: https://github.com/MystenLabs/sui/pull/7275

The idea for this PR is pretty simple. We assume that the custom memory will be represented by a type similar to `$Memory` which represents global memory of core Move. Consequently:
- we allow changing the name of the type representing memory
- we re-use generation of global variables representing memory for each user-defined type
- we introduce a new built-in that return a variable representing memory for a given type so that the memory model can be queried using native spec functions defined by the alternative Move flavor

As already remarked in the comments, this draft PR completely disables quant translation as it generates invariants using existing global memory operations. Additional work would be required to disable generation of only memory-related invariants rather than disabling all of them.

Yes

### DIFF
--- a/language/move-model/src/ast.rs
+++ b/language/move-model/src/ast.rs
@@ -1362,7 +1362,7 @@ impl<'a> fmt::Display for OperationDisplay<'a> {
                 }
                 Ok(())
             }
-            Exists(label_opt) => {
+            Memory(label_opt) => {
                 write!(f, "memory")?;
                 if let Some(label) = label_opt {
                     write!(f, "[{}]", label)?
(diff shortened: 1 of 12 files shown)


awelc:
Modified custom memory support to remove custom memory naming and bringing back quant translation at the cost of adding an additional builtin to expose user-defined Boogie-level global variables. This allows re-using existing global memory mapping (if suitable) and defining additional mappings that can be accessed from specs.

PR #814, opened 2023-01-10
Cherry-pick the fix of limit instantiation recursion for all type instantiation onto aptos-main

rahxephon89:


runtian-zhou (approved):
LGTM! Thanks for patchting this

PR #816, opened 2023-01-11
Cherry-pick the fix of limit instantiation recursion for all type instantiation onto main

rahxephon89:


rahxephon89:
@dariorussi I disabled a failed check in `[instantiation_tests.rs](https://github.com/move-language/move/pull/816/files#diff-5044214598ad2c43accebabcd4aca93a8c2304d6545ca7986195acb40d05155b)`. I think it may related to a code change [here](https://github.com/move-language/move/blob/fcffff879bb776710300546b44772c9c97810752/language/move-vm/runtime/src/loader.rs#LL1286C25-L1286C25). Could you take a look and give a hint on how to fix it? Thanks

language/move-vm/integration-tests/src/tests/instantiation_tests.rs
```diff
@@ -208,7 +208,7 @@ fn test_runner(
     let (res, time) = run_with_module(&mut gas, entry_spec);
     let err = res.err().unwrap().major_status();
     println!("* {}: {}ms - Status: {:?}", test_name, time, err);
-    assert_eq!(err, StatusCode::OUT_OF_GAS, "Must finish OutOfGas");
+    // assert_eq!(err, StatusCode::OUT_OF_GAS, "Must finish OutOfGas");
```
dariorussi, line 211: feel free to comment that out for now, thanks!

dariorussi (changes requested):
@rahxephon89 sorry I did not see the mention :( 
I think https://github.com/move-language/move/pull/798 is a better fix for that problem and I would use that. Would love to hear your thoughts.
The test comment and the leak test fix seem a good thing to cherry pick

language/move-vm/integration-tests/src/tests/instantiation_tests.rs
Aseechan84: Thank you

dariorussi:
or if that unblocks you faster I can just accept and I'll do the cleanup later.
Still let's talk about it, but maybe I'll just accept your PR....

dariorussi (approved):
approving if you need to unblock yourself

rahxephon89:
> 

Thanks!

PR #820, opened 2023-01-13
[move-prover] check data invariants for a &mut param ...

meng-xu-cs:
... when and only when the param goes out of scope.

Previously, the approach is to check data invariants **on every write-back to a &mut param**. This is by definition more restrictive as this is essentially saying that data invariants are not allowed to be violated even temporarily in funciton body, which might not be a practical assumption (as shown in the newly added test case.)

## Motivation

Intuitive data invariant checking

Yes

## Test Plan

CI, test cases updated


rahxephon89 (approved):
Thanks!

PR #821, opened 2023-01-13
[move-prover] model `type_name::get` in stdlib

### LINKED ISSUE #803: [move] Missing Boogie implementation of `type::name_get` function from stdlib
# 🐛 Bug

When adding a new native function in the `type` stdlib module (sha: https://github.com/move-language/move/commit/e9b0aeda778c1c46bd12eefaafdad7be3dd93209), the equivalent Boogie implementation has not been added.

Assigning myself to this issue but also @sblackshear in case he has an immediate idea what the Boogie side of things should look like.

meng-xu-cs:
The majority of the Prover work was firstly introduced when Aptos introduced the `type_info` Move module. Now that the stdlib has a similar module `type_name`, we should bridge the effort here.

This PR is still a bit rough as it is using hardcoded module address / function names to identify the type-reflection schemes. We should come up with more robust / extensible approach later.

NOTE: it is hard to unify them though. `type_info::type_name<T>()` and `type_name::get<T>()` are not equivalent, primarily due to the format of module address:
- in `type_info::type_name()`, the address is `0x1` while
- in `type_name::get()`, the address is `00000000000000000000000000000001`

## Motivation

Fix #803

Yes

## Test Plan

CI, new tests added

### DIFF
--- a/language/move-prover/tests/sources/functional/nonlinear_arithm.move
+++ b/language/move-prover/tests/sources/functional/nonlinear_arithm.move
@@ -1,6 +1,6 @@
 // exclude_for: simplify
 // exclude_for: cvc5
-// simplify and cvc5 are exculded due to timeout
+// flag: --timeout=160
 module 0x42::TestNonlinearArithmetic {
 
     spec module {
--- a/language/move-model/src/ast.rs
+++ b/language/move-model/src/ast.rs
@@ -6,14 +6,22 @@
 //! Note that in this crate, specs are represented in AST form, whereas code is represented
 //! as bytecodes. Therefore we do not need an AST for the Move code itself.
 
+use std::{
+    borrow::Borrow,
+    cell::RefCell,
+    collections::{BTreeMap, BTreeSet, HashSet},
+    fmt,
+    fmt::{Debug, Error, Formatter},
+    hash::Hash,
+    ops::Deref,
+};
+
+use internment::LocalIntern;
+use itertools::Itertools;
 use num::{BigInt, BigUint, Num};
+use once_cell::sync::Lazy;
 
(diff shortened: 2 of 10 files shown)


meng-xu-cs:
The unit test failure on this PR seems irrelevant to the changes in this PR. Tagging @junkil-park to see this has happened before and in those cases, what are the fixes.

language/move-prover/tests/sources/functional/nonlinear_arithm.move
```diff
@@ -1,6 +1,6 @@
 // exclude_for: simplify
 // exclude_for: cvc5
-// simplify and cvc5 are exculded due to timeout
+// flag: --timeout=160
```
meng-xu-cs, line 3: Added to avoid mysterious timeout / aborts not covered error

junkil-park:
@meng-xu-cs , I don't see any unit test failure in this PR. Is it all good now?

awelc (approved):
LGTM. This solution went a bit different route than the one I was preliminarily exploring which was to try and have the same algorithm to create a string representation of a type to be used for both native method implementation and for the model in the prover. I can see that this might have been overly complicated, though, due to both string computation algorithms working on different type representations (`Type` and 'TypeTag` which are not easily convertible to one another without appropriate context) and so I am happy with this solution to unblock some folks hitting problems in their specs with `type_name::get` not being supported.

zengxinhai:
Great, nice to see this pr merged. Do I need to update the move prover?

awelc:
> Great, nice to see this pr merged. Do I need to update the move prover?

If you use the Prover via Sui CLI then you'd have to wait a bit for Sui to pull a new version of core Move dependency but no action on your side should be necessary.

PR #822, opened 2023-01-14
[inlining] Revised inlining implementation

wrwg:
This is a new implementation of inlining which uses eager (call-by-value) parameter passing. This makes inline functions semantically fully compatible with regular functions. The new implementation also fixes some name clash problems.

Functionals for vector have been added and tested. Moreover, documentation has been added to the Move reference.

The related AIP is here: https://github.com/aptos-foundation/AIPs/pull/36


language/move-compiler/transactional-tests/tests/inlining/shadowing.move
```diff
@@ -0,0 +1,21 @@
+//# publish
+module 0x42::Test {
+
+    public inline fun foo(f:|u64|) {
+        let _x = 3;
+        f(_x);
+    }
+
+    public fun test_shadowing() {
```
meng-xu-cs, line 9: This is a mind-tweaking (in a good way) test case....

meng-xu-cs (approved):
Stamping to unblock.

This PR could enjoy a bit more testing on the new addition of inferred ability part although the testing could also come later.

wrwg:
> This PR could enjoy a bit more testing on the new addition of inferred ability part although the testing could also come later.

This actually fundamental in the entire working of this PR. There is e.g. code where we need do the following:

```
 let mut ty = Type_::multiple(loc, tys);
 self.infer_abilities(&mut ty);
```

Subsequent phases expect that `Type_::Apply` has fully populated abilities. This information is populated by the typing phase, but we run after it. We need to maintain the state of typed AST, and without this inference, nothing would work.

PR #824, opened 2023-01-17
[absint] Bring back associated error type

tnowacki:
- The associated type for errors is necessary for custom verifier passes using absint

## Motivation

Custom adapter verifier passes struggle when they want to use the same absint infrastructure, but with a custom error type. This adds back the type that was previously there.

## Test Plan

- Non functional change. ran tests


amnn (approved):
Seems reasonable

PR #836, opened 2023-01-24
[inline functions] Extending consistency checks and enabling memory management

wrwg:
This PR refines the inline implementation to peform various tests for consistency:

- report an error if `return` is used in the inlined functions (currently not supported)
- report errors if inlining introduces inconsistency in the expansion context (calling private functions, accessing module memory, etc.)
- re-arrange the acquires analysis, allowing inline functions to access memory without decl of `acquires`. Rather, check `acquires` after inlining.

The last step, specifically, also allows to write generic inline functions which deal with memory. This has been excercised in the [objects.move](https://github.com/wrwg/move-lang/blob/lambda2/language/move-compiler/transactional-tests/tests/inlining/objects.move) testcase.

The change in the acquires leads to a different style of acquire error messages (also less followup-errors) which results in baseline changes.

### DIFF
--- a/language/move-compiler/src/inlining/translate.rs
+++ b/language/move-compiler/src/inlining/translate.rs
@@ -3,15 +3,16 @@
 
 use crate::{
     diag,
-    expansion::ast::{AbilitySet, ModuleIdent, ModuleIdent_},
+    expansion::ast::{AbilitySet, ModuleIdent, ModuleIdent_, Visibility},
     inlining::visitor::{Dispatcher, Visitor, VisitorContinuation},
-    naming::ast::{StructDefinition, StructTypeParameter, TParamID, Type, Type_},
-    parser::ast::{StructName, Var},
+    naming,
+    naming::ast::{StructDefinition, StructTypeParameter, TParamID, Type, TypeName_, Type_},
+    parser::ast::{Ability_, StructName, Var},
     shared::{ast_debug, unique_map::UniqueMap, CompilationEnv, Identifier, Name},
     typing::{
         ast::{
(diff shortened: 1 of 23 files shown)


language/move-compiler/src/inlining/translate.rs
```diff
@@ -83,11 +85,12 @@ impl<'l> Inliner<'l> {
                 if !fdef.inline {
                     let mut visitor = OuterVisitor { inliner };
                     Dispatcher::new(&mut visitor).function(fdef);
+                    //Self::eprint_fdef(&format!("fun {} ", _name), fdef);
```
meng-xu-cs, line 88:
```suggestion
```

language/move-compiler/transactional-tests/tests/inlining/objects.move
```diff
@@ -0,0 +1,115 @@
...
+
+    public fun writer_addr_of<T: key>(r: &WriterRef<T>): address {
+        r.addr
+    }
+
+    // ================================================
+    // Basic Operations, inlined in the object definition context
+
+    public inline fun create<T: key>(signer: &signer, _ref: &OwnerRef, val: T) {
+        // In reality, we should ge the signer via the ref
```
meng-xu-cs, line 51:
```suggestion
        // In reality, we should get the signer via the ref
```

meng-xu-cs (approved):

language/move-compiler/src/inlining/translate.rs
wrwg: Prefer to keep -- as discussed for last PR.

PR #837, opened 2023-01-25
[borrow v2] Alternative design sketch using a checked reference types

wrwg:
This sketches an alternative design to #835 based on new types `global::Ref<T>` and `global::RefMut<T>`. Those types are considered to be known to the compiler, but some checking is defererred to runtime.

I'm not sure yet whether this design is correct, but I think it points at least in the right direction.

### DIFF
--- a/language/move-stdlib/sources/global.move
+++ b/language/move-stdlib/sources/global.move
@@ -1,0 +1,35 @@
+/// Represent references to global memory which can freely passed around functions. In core Move, references
+/// like `&R` and `&mut R` can only be passed downwards the stack if obtained from global memory. The types here
+/// relax this restriction, for the cost deferring some checks to runtime instead of compile time.
+module std::global {
+
+    /// Represents an immutable reference to a global. This value can be copied and dropped but not stored.
+    struct Ref<phantom T> has drop, copy {
+        handle: u64
+    }
+
+    /// Represents a mutable reference to global. This value is linear and cannot be stored.
+    struct RefMut<phantom T> has drop {
+        handle: u64
+    }
+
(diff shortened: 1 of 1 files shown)


wrwg:
@chen-robert

chen-robert:
Minor typo nit, I think `ref_mut` has an extra reference on `rc` right? 
>     public native fun ref_mut<T>(rc: &RefMut<T>): &mut T /* acquires T */;
should be
>     public native fun ref_mut<T>(rc: RefMut<T>): &mut T /* acquires T */;

I also wonder if it would make more sense to have these as bytecode instructions instead of native functions (keeping in mind that it would be a far more invasive change). It seems like, in order to add this functionality, there need to be large changes in the bytecode verifier. For example, the `acquires` semantics in the borrow checker and link-time checks for `T` ownership. 

Also, I think `T` cannot be a runtime type generic right?

language/move-stdlib/sources/global.move
```diff
@@ -0,0 +1,35 @@
...
+
+    /// Represents a mutable reference to global. This value is linear and cannot be stored.
+    struct RefMut<phantom T> has drop {
+        handle: u64
+    }
+
+    /// Borrows the resource at `addr` and returns a `Ref<T>` for it.
+    /// - This function will not type check if there is an active `&T` (indicated by the informal acquires)
+    /// - This function aborts if there is any active `RefMut<T>` for this location.
+    public native fun borrow<T:key>(addr: address): Ref<T> /* acquires T */;
```
jolestar, line 19: Is this method breaking the `borrow_global<T>` to the `T` limit? Such as, `T` must be defined in the caller's current module.

PR #844, opened 2023-01-28
[non functional change][compiler] Move local variable resolution to Naming

tnowacki:
- Move local variable resolution to naming
- Helps simplify logic for translations post naming so they can easily deal with scoping

## Test Plan

- Updated tests


language/move-compiler/src/typing/core.rs
```diff
@@ -72,7 +72,7 @@ pub struct Context<'env> {
     pub current_function: Option<FunctionName>,
     pub current_script_constants: Option<UniqueMap<ConstantName, ConstantInfo>>,
     pub return_type: Option<Type>,
-    locals: UniqueMap<Var, Type>,
+    locals: BTreeMap<Var, Type>,
```
amnn, line 75: Just for my own information -- why the switch away from `UniqueMap` here?  IIUC, `Var`s will still be unique coming into this mapping.

language/move-compiler/src/naming/translate.rs
```diff
@@ -495,10 +543,15 @@ fn function(
         body,
         specs: _,
     } = ef;
+    assert!(context.local_scopes.is_empty());
+    assert!(context.local_count.is_empty());
+    assert!(context.used_locals.is_empty());
+    context.local_scopes = vec![BTreeMap::new()];
+    context.local_count = BTreeMap::new();
```
amnn, line 550: Is there a need to partition this count by the name, or would it be okay to just count the number of locals in the function?  I.e. instead of this:

```
fun foo(): u64 {
  let x = 39;
  let y = x + 1;
  let x = y + 1;
  let y = x + 1;
  y
}
```

Lowering to:

```
fun foo(): u64 {
  let x#0#0 = 39;
  let y#0#0 = x#0#0 + 1;
  let x#1#0 = y#0#0 + 1;
  let y#1#0 = x#1#0 + 1;
  y#1#0
}
```

It becomes:

```
fun foo(): u64 {
  let x#0#0 = 39;
  let y#1#0 = x#0#0 + 1;
  let x#2#0 = y#1#0 + 1;
  let y#3#0 = x#2#0 + 1;
  y#3#0
}
```

(I think you could also potentially simplify away `color` using this approach by bumping all the vars in a macro expansion or inline by the number of locals you've used so far in the function body)

language/move-compiler/src/naming/translate.rs
```diff
@@ -507,15 +560,37 @@ fn function(
         body,
     };
     fake_natives::function(context.env, module_opt, name, &f);
+    let used_locals = std::mem::take(&mut context.used_locals);
+    remove_unused_bindings_function(context, &used_locals, &mut f);
+    context.local_scopes = vec![];
```
amnn, line 565: There's a potential risk that some function-specific field could get added and then not reset here -- is it worth grouping these fields into a single struct, which you `std::mem::take` at the end here, to guarantee that they all get reset?

language/move-compiler/src/hlir/translate.rs
```diff
@@ -962,12 +953,30 @@ fn exp_<'env>(
                 let tbool = N::Type_::bool(loc);
                 let tu64 = N::Type_::u64(loc);
                 let tunit = sp(loc, N::Type_::Unit);
-                let vcond = Var(sp(loc, new_temp_name(stack.context)));
-                let vcode = Var(sp(loc, new_temp_name(stack.context)));
+                // name guaranteed to be unique, so we can skip the depth/color
+                let vcond_ = N::Var_ {
+                    name: new_temp_name(stack.context),
```
amnn, line 958: Hmm, okay, this pattern might make it more difficult to implement the suggestion of a single count for both depth and colour, shared across all variable names...

amnn: Although I notice that the temp name is derived from a counter in the context -- this could start where the proposed count from naming leaves off.

amnn (commented):
I noticed a couple of cases where this change introduces duplicate errors in the test cases that might be worth looking at.  Also a potential simplification for how names are uniqued, but otherwise, looks great.

language/move-compiler/src/typing/core.rs
tnowacki: I was being lazy and didn't add TName for Var. I should really do that

language/move-compiler/src/naming/translate.rs
tnowacki: I was wondering if there was some value in being able to tell which instance of a given name it is. But that probably isn't true. 

The color thing is still really helpful (though not strictly necessary) when expanding macros as it lets you swap between "scopes" a bit easier, which is something needed when expanding the blocks/lambdas in macros.

tnowacki: Actually, just went to change this. I might leave it as is? Even though it isn't necessary, I rely on the count being `0` for parameters for ever so slightly better error messages. 

Though I could still set the count to 0 for parameters.

Interestingly though, if I do just user the number and increment it each time, I don't even really need the name except for display purposes. I'm not sure that's helpful anywhere, but it is interesting.

language/move-compiler/src/naming/translate.rs
tnowacki: > is it worth grouping these fields into a single struct,

Yes... and I've been really bad about this throughout the compiler. I might want to fix this everywhere. Maybe another PR?

language/move-compiler/src/hlir/translate.rs
tnowacki: Yeah that would definitely work too. I'm not sure either really feels simpler to me conceptually at first glance

language/move-compiler/tests/move_check/typing/declare_duplicate_binding.exp
```diff
@@ -26,6 +26,12 @@ error[E02001]: duplicate declaration, item, or annotation
   │              │     
   │              Previously declared here
 
+warning[W09002]: unused variable
+  ┌─ tests/move_check/typing/declare_duplicate_binding.move:6:19
+  │
+6 │         let (f, R{f}, f);
```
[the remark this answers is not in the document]
tnowacki: It's because I assign to `f`. Is this a confusing error? Not sure how common it will be.

With the numbers expanded it's something like
```
let (f#1, R { f: f#2 }, f#3);
f#3 = 0; f#3;
```

language/move-compiler/tests/move_check/liveness/unused_assignment.exp
```diff
@@ -1,54 +1,66 @@
+warning[W09002]: unused variable
```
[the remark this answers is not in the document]
tnowacki: The errors are _slightly_ different. It is an unused assignment vs an unused declaration. I've suppressed the error for this case since it will get an unused assignment error.

language/move-compiler/src/naming/translate.rs
amnn: Noted on `color` -- I think it also becomes helpful if/when macros are able to define new macros (vague recollection on my part, I may be wrong).

Re: detecting parameters, you could do this by making sure the first `p` indices in a p-arity function are reserved for its parameters. This does break down if you are relying on this property during inlining though (you won't necessarily know how many parameters there are in the function the body was inlined from).

The aim here was to simplify things slightly, so not a dealbreaker to leave it.

language/move-compiler/src/naming/translate.rs
amnn: Another PR sounds good to me.

amnn (approved):
LGTM!

Issue #460, opened 2022-09-08
[Proposed feature] Deprecate CompiledScript. Extend VM API to execute a module's function without publishing


sblackshear:
Sponsor: @sblackshear 

## Rationale
Move development involves writing a lot of code of the form `MyType::foo(&mytype)`, which can get quite verbose and inhibits readability.

## Design
If a struct `MyType` and function `foo` are both declared in the same module, and `foo`’s signature is something like `foo_mut(&mut MyType, …)` or `foo_immut(&MyType, …)`, the compiler should support the call syntax `t.foo_mut(...)` or `t.foo_immut(...). Specifically: 
- If `t: MyType`, either `t.foo_mut(...)` or `t.foo_immut(...)` can be called
- If `t: &mut MyType, either `t.foo_mut(...)` or `t.foo_immut(...)` can be called
- If `t: &MyType`,  `t.foo_immut(...)` can be called

jolestar:
I like this feature, but if multiple structs exist in the same Module, they may have conflicting function names. For example

1. there are two struct Foo{name:String}, struct Bar{name:String}. 
2. we want to get the name by foo.name(), bar.name().
3. so we define two functions: `public fun name(f:&Foo)` and `public fun name(b:&Bar)`. These two functions will have naming conflicts.

So is it possible to use implement syntax for struct function declarations like rust, e.g.

```move
module MyModule{
  struct Foo{
    name:String,
  }
  struct Bar{
    name: String,
  }
  
  impl Foo{
    public fun name(f:&Foo): String{
      f.name
    }
  }
  
  impl Bar{
    public fun name(b:&Bar): String{
      b.name
    }
  }
}
```

or do we support function overloading in Move?

```move
module MyModule{
  struct Foo{
    name:String,
  }
  struct Bar{
    name: String,
  }
  
  public fun name(f:&Foo): String{
    f.name
  }
  
  public fun name(b:&Bar): String{
    b.name
  }

}
```

sblackshear:
I would suggest that we keep this simple and continue requiring all functions declared in the same module have distinct names. For examples like the above, this would either mean:
- `Foo` and `Bar` should be declared in separate modules (which I think is typical Move--one "main" type per module)
- The functions would need to be called `foo_name`/`bar_name`, or `name`/`bar_name`

kcsongor:
An alternative syntax could be
```
(&t).foo_immut(...)
```
and
```
(&mut t).foo_mut(...)
```

[...]

PR #848, opened 2023-01-29
[method style syntax] Resurrected 49788a86d8c3d4f9b8f83e796625e086adcc02d6

tnowacki:
- Resurrected dot call syntax
- Next PR will add self and a few other additions (or I might add it to this PR as a separate commit)

- Built on #844


wrwg:
Any reason why this does not go towards `main`? As far as I remember, this is a feature we all want and have agree upon a while ago. @sblackshear

wrwg:
> Any reason why this does not go towards `main`? As far as I remember, this is a feature we all want and have agree upon a while ago. @sblackshear

https://github.com/move-language/move/issues/460

tnowacki:
> Any reason why this does not go towards `main`? As far as I remember, this is a feature we all want and have agree upon a while ago. @sblackshear

Just a place to iterate, we will merge back to main when it stabilizes.

PR #849, opened 2023-01-29
[move-stdlib] add `next_char_boundary` function

bytedeveloperr:
## Motivation
Motivation for this PR can be found here: #ISSUE-1560080403

Yes

## Test Plan
Added unit tests.

in the move-stdlib folder, run
```
cargo run -p df-cli -- test
```

### DIFF
--- a/language/move-stdlib/sources/string.move
+++ b/language/move-stdlib/sources/string.move
@@ -85,10 +85,14 @@ module std::string {
         internal_index_of(&s.bytes, &r.bytes)
     }
 
+    public fun next_char_boundary(s: &String, i: u64): u64 {
+        internal_next_char_boundary(&s.bytes, i)
+    }
 
     // Native API
     native fun internal_check_utf8(v: &vector<u8>): bool;
     native fun internal_is_char_boundary(v: &vector<u8>, i: u64): bool;
     native fun internal_sub_string(v: &vector<u8>, i: u64, j: u64): vector<u8>;
     native fun internal_index_of(v: &vector<u8>, r: &vector<u8>): u64;
+    native fun internal_next_char_boundary(s: &vector<u8>, i: u64): u64;
 }
--- a/language/move-stdlib/src/natives/string.rs
+++ b/language/move-stdlib/src/natives/string.rs
@@ -191,6 +191,57 @@ pub fn make_native_index_of(gas_params: IndexOfGasParameters) -> NativeFunction 
 }
 
 /***************************************************************************************************
+ * native fun internal_next_char_boundary
+ *
+ *   gas cost: base_cost + unit_cost * bytes_iterated
+ *
+ **************************************************************************************************/
+#[derive(Debug, Clone)]
+pub struct NextCharBoundaryGasParameters {
+    pub base: InternalGas,
+    pub per_byte: InternalGasPerByte,
+}
+
+fn native_next_char_boundary(
+    gas_params: &NextCharBoundaryGasParameters,
+    _context: &mut NativeContext,
+    _ty_args: Vec<Type>,
+    mut args: VecDeque<Value>,
+) -> PartialVMResult<NativeResult> {
+    debug_assert!(args.len() == 2);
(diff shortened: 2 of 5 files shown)


language/move-stdlib/src/natives/string.rs
```diff
@@ -190,6 +190,57 @@ pub fn make_native_index_of(gas_params: IndexOfGasParameters) -> NativeFunction
...
+
+    let s_arg = pop_arg!(args, VectorRef);
+    let s_ref = s_arg.as_bytes_ref();
+    let s_str = unsafe {
+        // This is safe because we guarantee the bytes to be utf8.
+        std::str::from_utf8_unchecked(s_ref.as_slice())
+    };
+
+    if i >= s_ref.len() {
+        // TODO: what abort code should we use here?
```
wrwg, line 223: Uhuh. That TODO is also in substring, and should have been long resolved. Can you use 2 for now?

language/move-stdlib/tests/string_tests.move
```diff
@@ -75,4 +75,19 @@ module std::string_tests {
         string::insert(&mut s, 1, string::utf8(b"xy"));
         assert!(s == string::utf8(b"axybcd"), 22)
     }
+
+    #[test]
+    fun test_next_char_boundary() {
+        let s = string::utf8(b"abcdefgh");
```
wrwg, line 81: Probably should have some real utf8 byte string here for testing

wrwg (approved):
This looks OK (besides the abort code problem which is inherited), but landing this will not yet make this part of Aptos or Sui. For that, it need to be cherry picked in sui-move and/or aptos-main, and probably also need some feature gating in the Move adapters of both systems. I'm not sure who is dealing with stdlib things at Sui, so I added Todd as a reviewer to potentially forward this.

language/move-stdlib/tests/string_tests.move
```diff
@@ -75,4 +75,75 @@ module std::string_tests {
...
+        let s = string::utf8(b"\xf0\x9f\x99\x82 move?");
+
+        let sub1 = string::sub_string_char(&s, 7, 7);
+        assert!(string::is_empty(&sub1), 22);
+
+        let sub2 = string::sub_string_char(&s, 7, 10);
+        assert!(string::is_empty(&sub2), 22);
+
+        let sub3 = string::sub_string_char(&s, 5, 3);
+        assert!(string::is_empty(&sub3), 22)
```
PaulFidika, line 128: Originally I thought the expected behavior for this should be that if you try to access an out-of-bounds character, an abort will occur, but this is not the case in Rust.
```
    let s = "Hello, wãrld!";
    let ss: String = s.chars().skip(7).take(99).collect();
```
works fine for example

PaulFidika (commented):
Looks good to me

language/move-stdlib/tests/string_tests.move
bytedeveloperr: yeah, it'll work fine; aborting depends on the operation you're performing on the string.

language/move-stdlib/sources/string.move
```diff
@@ -85,10 +85,37 @@ module std::string {
...
+            bytes: internal_sub_string_char(&s.bytes, i, j)
+        }
+    }
+
+
+    /// Insert the other string at a character index in given string.
+    public fun insert_char(s: &mut String, at: u64, c: String) {
+        let result = internal_insert_char(&s.bytes, at, &c.bytes);
+        s.bytes = result
+    }
```
tnowacki, line 109: I wonder if we need/want these functions. Just a thought but I wonder if what you want instead is a `Char` type and then expose something like `to_chars` and `from_chars` that work over `vector<Char>`. If you want to do inserts around unicode scalar boundaries, that feels like the right place to do it. (And it is a bit more generalized). 

And if you really want the functions he added, you can compose `insert` and `next_char_boundary` (thought it will needlessly rerun utf validation?

PR #859, opened 2023-01-31
sui-move: [12/x][move-package/lock] Use DependencyGraph in download_dependency_repos

amnn:
Another occasion where the previous implementation recursively traversed the transitive dependency graph -- the new approach discovers the graph and then iterates through its nodes to download them.

The second step is currently technically unnecessary, but becomes relevant in the presence of externally resolved dependencies where third-party resolvers can introduce sub-graphs into the `DependencyGraph` without having to explore them locally.

## Test Plan

```
```

## Stack

- #741 
- #745 
- #753 
- #754 
- #759 
- #760 
- #854 
- #855 
- #856 
- #857 
- #858 

See also: #788 for main.

### DIFF
--- a/language/tools/move-package/src/resolution/mod.rs
+++ b/language/tools/move-package/src/resolution/mod.rs
@@ -2,12 +2,11 @@
 // Copyright (c) The Move Contributors
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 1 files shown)


tzakian (approved):
Nice!

dariorussi (approved):

PR #860, opened 2023-01-31
sui-move: [13/x][move-package/lock] Externally resolved dependencies

amnn:
Adds support for dependencies that are resolved by calling out to third-party binaries.  The invocation is expected to return the transitive dependency sub-graph as a lock file, which is deserialized and merged into the main graph.

Externally resolved dependencies can appear in manifests but not lock files, and it is an error for a package to appear in the outputs of multiple resolvers (including the internal resolver).  To differentiate this case from the previous case where the internal resolver encounters the same package via another path, packages in the `DependencyGraph` track their resolver via an optional field (`None` for internally resolved packages).

Another consequence of this change is that the check that a dependency's name matches its package name moves from the creation of `DependencyGraph` to the creation of `ResolvedGraph`, to pick up cases where the name mismatch occurs in an externally resolved dependency.

## Clean-ups

- Also adds support for "progress" tests (snapshot tests for progress output).
- Stricter warnings during dependency parsing, e.g. warning when a `rev` field is supplied for a local dependency and other similar cases.
- Delete a stale (unused) snapshot file.

## Test Plan

New tests, including unit tests for the merging operation, and snapshot tests for external resolution lock file and resolved graph output:

```
```

## Stack

- #741 
- #745 
- #753 
- #754 
- #759 
- #760 
- #854 
- #855 
- #856 
- #857 
- #858 
- #859 

See also: #789 for main.


language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -336,6 +353,78 @@ impl DependencyGraph {
...
 
+    /// Add the graph in `other` to `self` consuming it in the process.  Assumes the root of `other`
+    /// is the only shared node between the two, and fails if this is not the case.  Labels packages
+    /// coming from `other` as being resolved by `resolver`.
+    ///
+    /// It is an error to attempt to merge into `self` after its `always_deps` (the set of packages
+    /// that are always transitive dependencies of its root, regardless of mode) has been
+    /// calculated.  This usually happens when the graph is created, so this function is intended
+    /// primarily for internal use, but is exposed for testing.
+    pub fn merge(&mut self, other: DependencyGraph, resolver: Symbol) -> Result<()> {
```
tzakian, line 364: Super minor nit (that you can ignore if you disagree): but `other` here (and the use of `other_<name>` below) while a good name doesn't  feel quite as clear as it could be. What would you think of something like `external`?

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -365,52 +454,180 @@ impl DependencyGraph {
...
+        if !output.status.success() {
+            let err_msg = format!(
+                "'{resolver}' failed to resolve {mode_label} for dependency '{to}' of package \
+                 '{from}'"
             );
+
+            if let Some(code) = output.status.code() {
+                bail!("{err_msg}. Exited with code: {code}.");
+            } else {
+                bail!("{err_msg}. Terminated by signal.");
```
tzakian, line 570: Other `bail` messages don't finish with a period, so to keep things consistent probably best to remove them here.

language/tools/move-package/src/source_package/parsed_manifest.rs
```diff
@@ -43,7 +43,14 @@ pub struct PackageInfo {
...
 #[derive(Debug, Clone, Eq, PartialEq)]
-pub struct Dependency {
+pub enum Dependency {
+    /// Parametrised by the binary that will resolve packages for this dependency.
+    External(Symbol),
+    Internal(IntDependency),
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct IntDependency {
```
tzakian, line 53: This name is confusing IMO. `InternalDependency` is much clearer and worth the extra typing :)

tzakian (approved):
This looks good to me. A couple nits mainly around naming. The main thing I'd like to discuss/have addressed before landing is the `IntDependency` rename.

language/tools/move-package/src/resolution/dependency_graph.rs
amnn: Is the inspiration here that the `other` graph is coming from external resolution?  That's definitely true for how `merge` is currently used, but I don't think it's intrinsic to the operation.  I won't defend `other` as a great name though -- what about ... `sub_graph`, or ... `extension`?  This recognises the fact that `other` needs to attach into a point in `self`.

dariorussi (approved):

PR #861, opened 2023-01-31
sui-move: [14/x][move-package/lock] Clean-up: Avoid explicit PackageLock::unlock

amnn:
Remove the explicit unlock function and rely on the drop handler always.  The only semantic difference is the order in which the locks are unlocked, but this should not be an issue (from a deadlock perspective), because the order in which they are locked is the always the same.

Motivated by how easy (especially when using the `?`/try syntax) it is to accidentally skip the explicit calls to `PackageLock::unlock`.  The reason why this still works is that the guards that it contains automatically unlock when they are dropped, so `PackageLock`'s drop handler should do the right thing, without need for an explicit `unlock` call.

## Test Plan

```
```

## Stack

- #741 
- #745 
- #753 
- #754 
- #759 
- #760 
- #854 
- #855 
- #856 
- #857 
- #858 
- #859 
- #860 

See also: #790 for main.

### DIFF
--- a/language/tools/move-package/src/lib.rs
+++ b/language/tools/move-package/src/lib.rs
@@ -154,7 +154,7 @@ impl BuildConfig {
     /// failure.
     pub fn compile_package<W: Write>(self, path: &Path, writer: &mut W) -> Result<CompiledPackage> {
         let resolved_graph = self.resolution_graph_for_package(path, writer)?;
-        let _mutx = PackageLock::lock();
+        let _mutx = PackageLock::lock(); // held until function returns
         BuildPlan::create(resolved_graph)?.compile(writer)
     }
 
@@ -166,7 +166,7 @@ impl BuildConfig {
         writer: &mut W,
     ) -> Result<CompiledPackage> {
         let resolved_graph = self.resolution_graph_for_package(path, writer)?;
-        let _mutx = PackageLock::lock();
+        let _mutx = PackageLock::lock(); // held until function returns
         BuildPlan::create(resolved_graph)?.compile_no_exit(writer)
     }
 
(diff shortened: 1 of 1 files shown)


tzakian (approved):
Looks good. I'm probably guilty of adding this in way back when so am glad to see this change :)

language/tools/move-package/src/lib.rs
```diff
@@ -168,21 +166,17 @@ impl BuildConfig {
...
     pub fn compile_package_evm<W: Write>(self, path: &Path, writer: &mut W) -> Result<()> {
         // resolution graph diagnostics are only needed for CLI commands so ignore them by passing a
         // vector as the writer
         let resolved_graph = self.resolution_graph_for_package(path, &mut Vec::new())?;
-        let mutx = PackageLock::lock();
-        let ret = BuildPlan::create(resolved_graph)?.compile_evm(writer);
-        mutx.unlock();
-        ret
+        let _mutx = PackageLock::lock();
+        BuildPlan::create(resolved_graph)?.compile_evm(writer)
```
tnowacki, line 179: What are the semantics for dropping unused locals like this in Rust? Are they well defined enough to rely on this?

tnowacki, line 179: Would it be safer to do something like this?
```suggestion
        let mutx = PackageLock::lock();
        BuildPlan::create(resolved_graph)?.compile_evm(writer)
        drop(mutx)
```

amnn: Yes, it's well defined -- drops get called in reverse order of declaration.  The gotchas are related to moves and the _ pattern -- if you move into a variable, you drop what was in the variable before and if you use the _ pattern instead of a variable name (tempting seeing as you don't need to refer to the guard), the guard will be dropped at assignment because it's treated as being never assigned.

amnn: Here's a good resource that runs through the drop rules: https://vojtechkral.github.io/blag/rust-drop-order/

dariorussi (approved):

language/tools/move-package/src/lib.rs
amnn: Talked about this offline, and settled on commenting the `lock()` call instead -- the issue with the explicit `drop` is similar to the explicit `unlock` in that it gives the false impression that all exit paths flow through that point in the function, when that isn't true (e.g. the return edge for the `?` will not flow through the explicit `drop`).

PR #865, opened 2023-02-01
[Move-package] Enable bytecode as dependencies

rahxephon89:
## Motivation

Re-enable the feature to use bytecode as dependencies when compiling Move programs.

Yes

## Test Plan

Add later

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3456,6 +3456,7 @@ dependencies = [
  "log",
  "move-binary-format",
  "move-command-line-common",
+ "move-compiler",
  "move-core-types",
  "move-model",
  "move-stackless-bytecode",
--- a/language/move-compiler/src/command_line/compiler.rs
+++ b/language/move-compiler/src/command_line/compiler.rs
@@ -607,6 +607,8 @@ fn generate_interface_files_for_deps(
     let interface_files_paths =
         generate_interface_files(deps, interface_files_dir_opt, module_to_named_address, true)?;
     deps.extend(interface_files_paths);
+    // Remove bytecode files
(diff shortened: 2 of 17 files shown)


rahxephon89:
Hi @tnowacki @tzakian, although it will not be landed on main, could you take a look and give some feedback? Thanks!

language/move-compiler/src/parser/mod.rs
```diff
@@ -63,6 +63,8 @@ pub(crate) fn parse_program(
 
     let targets = find_move_filenames_with_address_mapping(targets)?;
     let mut deps = find_move_filenames_with_address_mapping(deps)?;
+    // Remove bytecode files
+    deps.retain(|p| !p.path.as_str().ends_with(MOVE_COMPILED_EXTENSION));
```
tnowacki, line 67: Could you explain the reasoning to remove bytecode files from deps? 

Targets means that the compiler is intended to produce bytecode modules for those sources files, which I'm not sure makes sense here

rahxephon89: At this point, deps may already contain paths to mv files that are used as external dependencies, which will cause panic later when calling the function parse_file.

tnowacki: So the "interface" files have been generated at this point right? Would it make sense to just remove the `.mv` files from deps after generating the source files for them?

rahxephon89: Thanks for the advice Todd! I moved the code to [here]( https://github.com/move-language/move/blob/4d24d3df379cf655eb43ef7fccd7db3e71472860/language/move-compiler/src/command_line/compiler.rs#L611). I am wondering whether it makes more sense?

wrwg (approved):

PR #868, opened 2023-02-03
[natives] New function `exists_at`

wrwg:
This adds a new native function `globals::exists_at` which allows to check existance of a resource outside of the module which defines this resource. This functionality is needed for Move Objects to be able to build typed object references.

This is an alternative to #852. Preview state, no tests added yet, and the gas formula is incomplete.

### DIFF
--- a/language/move-stdlib/sources/globals.move
+++ b/language/move-stdlib/sources/globals.move
@@ -1,0 +1,6 @@
+/// Utilities for working with global memory.
+module std::globals {
+
+    /// Tests whether the resource exists in global memory.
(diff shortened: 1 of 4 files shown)


davidiw:
I see, I hadn't realized I could pop out a Type, thus my approach had to do some weird reading of storage. This looks good, only concern is that we cannot constrain the exists_at function to where we want it, we move quickly in opening it up to the world. My only thought was do we want to move to this so quickly? I don't really mind either way, but would like to land an approach to unblock the next phase of object code ;)

PR #873, opened 2023-02-06
[loader] Expose TypeTag to Type conversion

awelc:
## Motivation

Currently, we have the ability to convert `Type` to `TypeTag` in native context but not vice versa which may be useful if in the implementation of a native method you need to created types (struct types specifically) out of thin air, such as in https://github.com/MystenLabs/sui/pull/8120.

Yes


awelc:
Converted to draft as apparently exposing this is a big no-no...

PR #876, opened 2023-02-07
[move-compiler] allow the inlining origin to be embedded in spec marker

meng-xu-cs:
This commit extends `UnannotatedExp_::Spec` as following:

```diff
- Spec(SpecId, BTreeMap<Var, Type>),
+ Spec(SpecId, Option<SpecIdent>, BTreeMap<Var, Type>),
```

Where `SpecIdent` marks the origin of the spec block (before inlining)

```rust
pub struct SpecIdent {
    pub module: Option<ModuleIdent_>,  // None means a script
    pub function: Symbol,
    pub id: SpecId,
}
```

This commit also re-assigns `SpecId` for each `UnannotatedExp_::Spec` inlined to maintain that each inline spec block has a unique ID. This is needed later in the compilation stage.

This behavior is tested with a `spec_inlining.move` test case added in the `move_check` folder. Previously, this test case hits an assertion in the compiler.

## Motivation

More on spec inlining

Yes

## Test Plan

- CI
- New test case

### DIFF
--- a/language/move-compiler/src/cfgir/borrows/mod.rs
+++ b/language/move-compiler/src/cfgir/borrows/mod.rs
@@ -2,18 +2,21 @@
 // Copyright (c) The Move Contributors
 // SPDX-License-Identifier: Apache-2.0
 
-mod state;
+use std::collections::BTreeMap;
 
-use super::absint::*;
+use move_ir_types::location::*;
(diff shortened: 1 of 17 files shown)


meng-xu-cs:
Closing as the changes are enclosed in #879

PR #881, opened 2023-02-08
[compiler] warn on unused `&mut` qualifier in parameters

sblackshear:
Add new warning type: unused `&mut` qualifier. A qualifier is considered unused if the corresponding variable is not:
- mutated
- used in an expression that mutably borrows a field
- passed as an argument to a call that requires a parameter of `&mut` type
- returned from a function with a `&mut` return type

Added tests for each of these cases.

- Note 1: there are 45 existing compiler tests that are failing because they now (correctly) get flagged with unused `&mut` warnings. My inclination is to to update the expected values rather than modifying the tests, but let me know if that doesn't sound right.
- Note 2: I added this warning inside translate.rs because that's where very similar unused variables warning lives, but the file is perhaps getting a bit crowded. Maybe both of these warnings want to be separate passes in the future, but I'm not sure.


language/move-compiler/src/hlir/translate.rs
```diff
@@ -1323,10 +1363,26 @@ fn exp_impl(
...
+                    _ => (),
+                }
+            }
             HE::Borrow(mut_, e, f)
         }
         TE::TempBorrow(mut_, te) => {
             let eb = exp_(context, result, None, *te);
+            if mut_ {
+                match &eb.exp.value {
+                    HE::Move { var, .. } | HE::Copy { var, .. } => {
```
tnowacki, line 1380: For all these, I think you can do `if let HE::Move { var, .. } | HE::Copy { var, .. } = &eb.exp.value`

I'm not sure one is more readable than the other though

language/move-compiler/src/hlir/translate.rs
```diff
@@ -1187,9 +1216,20 @@ fn exp_impl(
...
                 acquires,
             } = *call;
-            let expected_type = H::Type_::from_vec(eloc, single_types(context, parameter_types));
+            let expected_single_types = single_types(context, parameter_types);
+            let expected_type = H::Type_::from_vec(eloc, expected_single_types.clone());
             let htys = base_types(context, type_arguments);
             let harg = exp(context, result, Some(&expected_type), *arguments);
+            for (i, ty) in expected_single_types.into_iter().enumerate() {
+                if ty.value.is_mut_ref() {
+                    match harg.exp_at_index(i).exp.value {
```
tnowacki, line 1225: TooManyArgs is not a blocking error, so these might not line up. (I think this `harg.exp_at_index(i)` might fail).

language/move-compiler/src/hlir/translate.rs
```diff
@@ -1141,6 +1162,14 @@ fn exp_impl(
         TE::Mutate(tl, tr) => {
             let er = exp(context, result, None, *tr);
             let el = exp(context, result, None, *tl);
+            if el.ty.value.is_mut_ref() {
```
tnowacki, line 1165: This should always be true?

language/move-compiler/src/hlir/translate.rs
```diff
@@ -1921,6 +1977,41 @@ fn check_trailing_unit_statement(context: &mut Context, sp!(_, s_): &mut H::Stat
...
+fn check_unused_mut_qualifier(context: &mut Context, locals: &UniqueMap<Var, H::SingleType>) {
+    let signature = context
+        .signature
+        .as_ref()
+        .expect("ICE Signature should always be defined when checking a function body");
+    for (var, ty) in locals.key_cloned_iter() {
+        match ty.value {
+            H::SingleType_::Ref(is_mut, _) if is_mut => {
+                if !context.mutably_used_locals.contains(&var)
+                    && signature.is_parameter(&var)
```
tnowacki, line 1993: What about for non parameters?

tnowacki (commented):
I wonder if it might be easier to do this as a pass after hlir translation instead of during. 

A lot of this logic is looking for cases that will be covered by `maybe_freeze`. 

To state this slightly differently, hlir::translate has to basically already do these checks because subtyping of references doesn't exist at runtime. 

So if you wait until after the translation is done, you should be able to just track locals of type `&mut _` that are used in the positions:
- `Mutate(Move|Copy var, _)`
- `Borrow(/* mut */ true, Move|Copy var, _)`
- `ModuleCall { args: ExpList(Move | Copy var, _) ... }`
(Sorry for the pseudo code but I think you get the idea... I hope)

It should make this a tad easier to write, and it should be able to cover everything, not just parameters

PR #887, opened 2023-02-10
[Move-prover] Cherry-pick prover related PRs to aptos-main

rahxephon89:
## Motivation

Cherry-pick recent prover-related PRs onto aptos-main

### DIFF
--- a/language/documentation/examples/diem-framework/move-packages/experimental/sources/Vote.move
+++ b/language/documentation/examples/diem-framework/move-packages/experimental/sources/Vote.move
@@ -329,6 +329,7 @@ module ExperimentalFramework::Vote {
         let removed_ballots = vector::empty();
         while ({
             spec {
+                invariant unique_ballots(ballots);
                 invariant no_expired_ballots(ballots, DiemTimestamp::spec_now_seconds(), i);
                 invariant vector_subset(ballots, old(ballot_data).ballots);
                 invariant i <= len(ballots);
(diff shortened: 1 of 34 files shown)


rahxephon89:
@meng-xu-cs @junkil-park After cherry picking recent prover PRs onto `aptos-main`, verification of `move-stdlib` will time out. Do you have any idea on the reason why this happens?

meng-xu-cs:
Seems that the timeout is about arithmetic operations:

```
error: verification out of resources/timeout (global timeout set to 80s)
    ┌─ ./sources/fixed_point32.move:265:5
    │  
265 │ ╭     public fun round(num: FixedPoint32): u64 {
266 │ │         let floored_num = floor(num) << 32;
267 │ │         let boundary = floored_num + ((1 << 32) / 2);
268 │ │         if (num.value < boundary) {
    · │
272 │ │         }
273 │ │     }
    │ ╰─────^
```

Deferring it to @junkil-park for some quick comment.

In the meanwhile, we can try to add `pragma timeout = 120` or sth more to the spec of `round`, if giving it more time helps with the issue.

rahxephon89:
@meng-xu-cs Could you double check that all your recent prover-related PRs (after Jan 1) landed on `main` are cherry-picked here? Thanks

meng-xu-cs:
@rahxephon89 confirmed that all my commits are there.

PR #889, opened 2023-02-13
[Move-prover] Cherry-pick #878

rahxephon89:
## Motivation

Cherry-pick #878 onto aptos-main


meng-xu-cs (approved):
stamping as this is cherry-pick

PR #890, opened 2023-02-13
[compiler] Preserve module handle ordering across address changes

amnn:
Ensure the order of module handles in the compiler output does not change when a package's numerical address changes.  This property is helpful for implementing source validation (confirming the source that a binary originated from, by re-building that source and checking the resulting bytecode matches the binary).

Requires a module's `dependency_order` (calculated during expansion) to pay attention to its package name, and for the module handles to be ordered with respect to this during lowering, which already uses this ordering for sorting dependencies.

Relies on the fact that `petgraph::DiGraphMap`'s iteration order is sensitive to insertion order (which is what is actually changed).

## Test Plan

New test for module-handle ordering:

```
$ cargo nextest
```


tnowacki (approved):
LGTM, might just want to swap the clone to copied, but I doubt it matters other than being nice to read :)

PR #891, opened 2023-02-13
[move-compiler] collect information on inlined lambda for spec

meng-xu-cs:
This commit includes a complete revamp of the anchor point information for
inline spec.

More prominently, if an inline spec uses a lambda function pointer which is
passed in a function arguments (i.e., the body of the lambda is defined outside
of this inline function), the compiler will create an implied function to capture
the lambda.

In a concrete example:

```move
module 0x42::Test {
    inline fun apply(v: u64, predicate: |u64| bool): bool {
        spec {
            assert v >= 42;
            assert predicate(v);
        };
        predicate(v)
    }

    public fun test_apply(a1: u64, a2: u64) {
        apply(0, |v| v >= 0);
        apply(0, |v| v != a1 + a2);
    }
}
```

When compiling for verification, the result module of
`0x42::Test` will include not only the bytecode for
`test_apply`, but also two additional functions:

```move
fun test_apply_predicate_0(v: u64): bool {
  v >= 0
}

fun test_apply_predicate_1(v: u64, a1: u64, a2: u64): bool {
  v != a1 + a2
}
```

Both above functions are automatically generated and
added to the existing list of functions in HLIR phase.

The best way to review this commit is by following the compilation flow:
- expansion/ast.rs
- expansion/translate.rs
- expansion/*.rs
- naming/ast.rs
- naming/translate.rs
- typing/ast.rs
- typing/translate.rs
- typing/*.rs
- inlining/visitor.rs
- inlining/translate.rs
- hlir/ast.rs
- hlir/translate.rs
- cfgir/*.rs
- to_bytecode/context.rs
- to_bytecode/translate.rs

## Motivation

Continued effort on supporting inline function in prover

Yes

## Test Plan

CI


language/move-compiler/src/inlining/translate.rs
```diff
@@ -285,6 +302,108 @@ impl<'l, 'r> Visitor for SubstitutionVisitor<'l, 'r> {
                     });
                     *id = SpecId::new(self.inliner.current_spec_block_counter);
                     self.inliner.current_spec_block_counter += 1;
+
+                    // after inlining, we should
+                    // - remove the function pointers in `used_locals`
```
wrwg, line 307: Can we pull the logic here in an independent function? (And perhaps section in this file?)

meng-xu-cs: Done

- The spec anchor rewriting logic is extracted to a separate function
- The lambda lifting logic is further extracted to a separate section in the same file

wrwg (approved):

PR #896, opened 2023-02-14
[bug fix] Fix doc comment start

tnowacki:
- Fixed panic caused by bad lexing of doc comments

Thanks @otter-sec for the fix!
    
## Motivation

- panics are not good ux

## Test Plan

- added a test


tzakian (approved):
Change looks good to me.

PR #897, opened 2023-02-15
[sanity fix] Declaration order of structs/constants/functions used for definition tables

tnowacki:
- The definition tables in the CompiledModule were populated via the alphanumeric ordering
- This might be unexpected as other constructs, namely fields, do not do this.
- This fix sorts the definitions based on the source file declaration order

## Test Plan

- new test and updated tests

### DIFF
--- a/language/move-compiler/src/to_bytecode/translate.rs
+++ b/language/move-compiler/src/to_bytecode/translate.rs
@@ -448,16 +448,12 @@ fn struct_defs(
     m: &ModuleIdent,
     structs: UniqueMap<StructName, H::StructDefinition>,
 ) -> Vec<IR::StructDefinition> {
-    let sorting: HashMap<Symbol, usize> = structs
-        .iter()
-        .map(|(_, sname, sdef)| (*sname, sdef.index))
-        .collect();
-    let mut structs_vec = structs
+    let mut structs = structs.into_iter().collect::<Vec<_>>();
+    structs.sort_by_key(|(_, s)| s.index);
+    structs
         .into_iter()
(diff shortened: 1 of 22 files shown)


language/move-compiler/src/to_bytecode/translate.rs
```diff
@@ -465,13 +443,31 @@ fn var_info(
...
+) -> Vec<IR::StructDefinition> {
+    let sorting: HashMap<Symbol, usize> = structs
+        .iter()
+        .map(|(_, sname, sdef)| (*sname, sdef.index))
+        .collect();
+    let mut structs_vec = structs
+        .into_iter()
+        .map(|(s, sdef)| struct_def(context, &m, s, sdef))
+        .collect::<Vec<_>>();
+    structs_vec.sort_by_key(|sdef| sorting[&sdef.value.name.0]);
```
tzakian, line 459: Could this and the sorting done for constants/functions below be written without needing this `sorting` table and indexing into it? i.e., something like this:

```rust
let mut sorted_structs = structs.into_iter().collect();
sorted_structs.sort_by_key(|(_, sdef)| sdef.index);
sorted_structs.into_iter().map(|(s, sdef)| struct_def(context, &m, s, sdef)).collect()
```

tzakian (approved):
LGTM, although you might want to consider changing how the sorting is done in `to_bytecode`.

language/move-compiler/src/to_bytecode/translate.rs
tnowacki: ah, yeah that feels a bit better doesn't it. Thanks!

PR #898, opened 2023-02-15
Add abstract gas sizing for StructTag, TypeTag, Value

oxade:
## Motivation

We need some way of deterministically sizing types for gas metering.
For now we're okay handrolling but will use macros to ensure structs are properly sized.

Yes
## Test Plan

Unit tests

### DIFF
--- a/language/move-core/types/src/account_address.rs
+++ b/language/move-core/types/src/account_address.rs
@@ -116,7 +116,7 @@ impl AccountAddress {
 
     /// TODO (ade): use macro to enfornce determinism
     pub fn abstract_size_for_gas_metering(&self) -> AbstractMemorySize {
-        AbstractMemorySize::new(std::mem::size_of::<Self>() as u64)
+        AbstractMemorySize::new(Self::LENGTH as u64)
     }
 }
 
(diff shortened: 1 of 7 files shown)


language/move-core/types/src/account_address.rs
```diff
@@ -111,6 +113,11 @@ impl AccountAddress {
             .map_err(|_| AccountAddressParseError)
             .map(Self)
     }
+
+    /// TODO (ade): use macro to enfornce determinism
+    pub fn abstract_size_for_gas_metering(&self) -> AbstractMemorySize {
+        AbstractMemorySize::new(std::mem::size_of::<Self>() as u64)
```
tzakian, line 119: Nit: They'll probably be the same, but can this be just `Self::LENGTH`?

tzakian (approved):
This looks good for now (modulo that one nit possibly). 

One thing I'm also wondering is if before the macro-approach if it would be good to impose some structure on what is/is not abstract sizeable and have a trait for getting the abstract size? e.g.,

```rust
trait AbstractSize {
 fn abstract_size(&self) -> AbstractMemorySize;
}

impl AbstractSize for AccountAddress {
  fn abstract_size(&self) -> AbstractMemorySize {
    ....
  }
}
```

oxade:
> This looks good for now (modulo that one nit possibly).
> 
> One thing I'm also wondering is if before the macro-approach if it would be good to impose some structure on what is/is not abstract sizeable and have a trait for getting the abstract size? e.g.,
> 
> ```rust
> trait AbstractSize {
>  fn abstract_size(&self) -> AbstractMemorySize;
> }
> 
> impl AbstractSize for AccountAddress {
>   fn abstract_size(&self) -> AbstractMemorySize {
>     ....
>   }
> }
> ```

Makes sense. Can do in subsequent PR.

PR #899, opened 2023-02-15
[move-prover] table intrinsic: bug fix and new intrinsic function

meng-xu-cs:
This commit addresses two issues:

1. Added a new intrinsic spec function `spec_map_new` to allow the following native spec function to be emulated: `spec native fun spec_new<K, V>(): Table<K, V>;` This funciton will return an empty table on the spec side.

2. While testing the newly added intrinsic funciton, several issues are discovered.

   - One issue fixed in this commit is the equality encoding of the `table` intrinsic, evident by the fact that the newly added test `create_and_insert_fail2` will pass the verification which it really should not.

     The table equality check is now formulated in a manner symmetric to both tables (the previous formulate seems to be biased to one table, although I don't claim that I fully understand the difference).

  - Another issue not fixed in this commit is shown in the newly added test case `create_and_insert_fail_due_to_typed_key_encoding`.

    The root cause is that for tables, the key encoding is type-dependent. Hence, an encoding of `u8` and `u258` of the same numeric key is different. I.e., there is no correlation between `EncodeKey'u8'(1)` and `EncodeKey'u258'(1)` although both take the signature of `EncodeKey'<type>'(int): int`.

    Nevertheless, this subtle discrepancy is captured in the test case and hopefully can be solved later.

## Motivation

feature addition and bug fix

Yes

## Test Plan

- CI
- new test cases

### DIFF
--- a/language/move-prover/boogie-backend/src/prelude/native.bpl
+++ b/language/move-prover/boogie-backend/src/prelude/native.bpl
@@ -269,12 +269,9 @@ function $IsEqual'{{Type}}{{S}}'(t1: {{Self}}, t2: {{Self}}): bool {
 {%- else -%}
 function $IsEqual'{{Type}}{{S}}'(t1: {{Self}}, t2: {{Self}}): bool {
     LenTable(t1) == LenTable(t2) &&
-    (forall k: int :: (
-        ContainsTable(t1, k) ==> (
-            ContainsTable(t2, k) && GetTable(t1, k) == GetTable(t2, k)
-        ) &&
-        ContainsTable(t2, k) ==> ContainsTable(t1, k)
-    ))
+    (forall k: int :: ContainsTable(t1, k) <==> ContainsTable(t2, k)) &&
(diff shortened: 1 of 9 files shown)


rahxephon89 (approved):
Could you cherry-pick this onto `aptos-main` after landing this? Thanks

language/move-prover/boogie-backend/src/prelude/native.bpl
```diff
@@ -425,10 +422,11 @@ function {:inline} {{impl.fun_spec_has_key}}{{S}}(t: ({{Self}}), k: {{K}}): bool
 
 {%- if impl.fun_spec_set != "" %}
 function {:inline} {{impl.fun_spec_set}}{{S}}(t: {{Self}}, k: {{K}}, v: {{V}}): {{Self}} {
-    if (ContainsTable(t, {{ENC}}(k))) then
-        UpdateTable(t, {{ENC}}(k), v)
+    (var enc_k := {{ENC}}(k);
```
junkil-park, line 425: I am just curious if this change was also necessary to resolve the inconsistency issue.

PR #900, opened 2023-02-15
[move-vm] Change gas model within the VM to be tier-based

tzakian:
This is the preliminary changes to the way that we account for gas in the Move VM. 

Note that we do not calculate the size changes to the operand stack yet!

Note that this PR is stacked on a PR adding a CSV output option to unit test statistics (see #903) so only look at the top commit.


wrwg:
I wonder whether the intention of this PR is to deliberately go away from the model which was co-developed by Aptos and Mysten in the last fall, making merging to upstream impossible. If this is the case, should this really be in the Move repo? We also internally discussed to create our own flavor of the VM at Aptos (not started yet), but we do not think this would live in the Move repo, which is a community project in the first place.

tzakian:
Hey Wolfgang,

The idea here is to experiment and see where to move the codebase and hopefully arrive at a place where both us (Mysten) and the community (everybody else including Aptos) would benefit from the change, and with a backward compatibility story in place. 

I agree that this PR as is looks pretty breaking but we think there are ways where we could eventually put this work on gas back in to `main` in a sane way that would allow a choice as to which type of metering scheme is used (instruction-based or a tier-based model). Or, for there to be different interpreter implementations that can be chosen.

language/move-vm/test-utils/src/gas_schedule.rs
```diff
@@ -137,7 +235,145 @@ impl<'a> GasStatus<'a> {
...
+                    self.cost_table.stack_height_tier(self.stack_height_current);
+                self.stack_height_current_tier_mult = next_mul;
+                self.stack_height_next_tier_start = next_tier;
+            }
+        }
+
+        Ok(())
+    }
+
+    pub fn pop_stack(&mut self, pushes: u64) -> PartialVMResult<()> {
```
dariorussi, line 276: wonder if the arg name pushes is what you wanted...

language/move-vm/test-utils/src/gas_schedule.rs
```diff
@@ -137,7 +235,145 @@ impl<'a> GasStatus<'a> {
...
+        Ok(())
+    }
+
+    pub fn decrease_stack_size(&mut self, size_amount: u64) -> PartialVMResult<()> {
+        match self.stack_size_current.checked_sub(size_amount) {
+            None => {
+                self.stack_size_current = 0;
+            }
+            Some(new_size) => {
+                if new_size > self.stack_size_high_water_mark {
```
dariorussi, line 337: is this ever possible?

language/move-vm/test-utils/src/gas_schedule.rs
```diff
@@ -168,122 +404,112 @@ impl<'a> GasStatus<'a> {
...
     pub fn set_metering(&mut self, enabled: bool) {
         self.charge = enabled
     }
 }
 
-fn get_simple_instruction_opcode(instr: SimpleInstruction) -> Opcodes {
-    use Opcodes::*;
+fn get_simple_instruction_stack_change(
+    instr: SimpleInstruction,
+) -> (u64, u64, AbstractMemorySize, AbstractMemorySize) {
```
dariorussi, line 414: a comment as to what the return values are would be good

language/move-vm/test-utils/src/gas_schedule.rs
```diff
@@ -168,122 +404,112 @@ impl<'a> GasStatus<'a> {
...
+        let size_increase = ret_vals
+            .map(|ret_vals| {
+                ret_vals.fold(AbstractMemorySize::zero(), |acc, elem| {
+                    acc + elem.legacy_abstract_memory_size()
+                })
+            })
+            .unwrap_or_else(AbstractMemorySize::zero);
+        // Charge for the stack operations. We don't count this as an "instruction" since we
+        // already accounted for the `Call` instruction in the
+        // `charge_native_function_before_execution` call.
```
dariorussi, line 491: should this be uncommented at some point?

language/move-vm/test-utils/src/gas_schedule.rs
```diff
@@ -293,238 +519,231 @@ impl<'b> GasMeter for GasStatus<'b> {
...
+        });
+        // Charge for the pops, no pushes, and account for the stack size decrease. Also track the
+        // `CallGeneric` instruction we must have encountered for this.
+        self.charge(1, 0, pops, 0, stack_reduction_size.into())
     }
 
     fn charge_ld_const(&mut self, size: NumBytes) -> PartialVMResult<()> {
-        self.charge_instr_with_size(Opcodes::LD_CONST, u64::from(size).into())
+        // Charge for the load from the locals onto the stack.
+        self.charge(1, 1, 0, u64::from(size), 0)
```
dariorussi, line 553: what is the 1, 1 for push and pop?

language/move-vm/test-utils/src/gas_schedule.rs
```diff
@@ -293,238 +519,231 @@ impl<'b> GasMeter for GasStatus<'b> {
...
         _val: impl ValueView,
     ) -> PartialVMResult<()> {
+        // We already charged for this based on the bytes that we're loading so don't charge again.
         Ok(())
     }
 
     fn charge_copy_loc(&mut self, val: impl ValueView) -> PartialVMResult<()> {
-        self.charge_instr_with_size(Opcodes::COPY_LOC, val.legacy_abstract_memory_size())
+        // Charge for the copy of the local onto the stack.
+        self.charge(1, 1, 0, val.legacy_abstract_memory_size().into(), 0)
```
dariorussi, line 566: here and below too not clear about the 1,1 for pop, push

dariorussi (commented):
I wonder if it would be better to have a completely new file and get a feature flag to compile in one or the other...

dariorussi (approved):
great thanks!

PR #902, opened 2023-02-16
[cherry-pick][prover] table intrinsic

meng-xu-cs:
This is to cherry-pick two commits landed in main from PR #899.


awelc (approved):
LGTM

PR #904, opened 2023-02-17
[compiler] Enforce a canonical order for module handles

amnn:
Add a pass to the compiler, immediately after conversion from IR to bytecode, to re-order module handles to guarantee a stable ordering.

## Test Plan

New tests for canonical order:

```
$ cargo nextest -- canonicalize_
```

### DIFF
--- a/language/move-compiler/src/to_bytecode/canonicalize_handles.rs
+++ b/language/move-compiler/src/to_bytecode/canonicalize_handles.rs
@@ -6,7 +6,8 @@ use std::collections::HashMap;
 use move_binary_format::{
     access::{ModuleAccess, ScriptAccess},
     file_format::{
-        CompiledScript, ModuleHandleIndex, Signature, SignatureToken, StructDefinition,
+        Bytecode, CodeUnit, CompiledScript, FunctionDefinition, FunctionDefinitionIndex,
+        FunctionHandleIndex, ModuleHandleIndex, Signature, SignatureToken, StructDefinition,
         StructDefinitionIndex, StructFieldInformation, StructHandleIndex, TableIndex,
     },
     internals::ModuleIndex,
@@ -105,10 +106,9 @@ pub fn in_module(
         } else {
             // Order the remaining handles afterwards, in lexicographical order of module, then
             // struct name.
-            let name = module.identifier_at(handle.name).as_str();
             ReferenceKey::External {
                 module: handle.module,
-                name,
+                name: module.identifier_at(handle.name).as_str(),
             }
         }
     });
@@ -131,6 +131,40 @@ pub fn in_module(
 
     // 2 (c). Update ordering for struct handles.
     apply_permutation(&mut module.struct_handles, structs);
+
+    // 3 (a). Choose ordering for function handles
+    let function_defs = function_definition_order(&module.function_defs);
+    let functions = permutation(&module.function_handles, |ix, handle| {
+        if handle.module == module.self_handle_idx() {
+            // Order functions from this module first, and in definition order
+            let Some(def_position) = function_defs.get(&FunctionHandleIndex(ix)) else {
+                panic!("ICE function handle from module without definition: {handle:?}");
+            };
+            ReferenceKey::Internal(def_position.0)
+        } else {
+            // Order the remaining handles afterwards, in lexicographical order of module, then
+            // struct name.
+            ReferenceKey::External {
+                module: handle.module,
+                name: module.identifier_at(handle.name).as_str(),
+            }
+        }
+    });
+
+    // 3 (b). Update references to function handles
+    for inst in &mut module.function_instantiations {
+        permute!(FunctionHandleIndex, inst.handle, functions);
+    }
+
+    for def in &mut module.function_defs {
+        permute!(FunctionHandleIndex, def.function, functions);
+        if let Some(code) = &mut def.code {
+            permute_code(code, &functions);
+        }
+    }
+
+    // 3 (c). Update ordering for function handles.
+    apply_permutation(&mut module.function_handles, functions);
 }
 
 /// Apply canonicalization to a compiled script.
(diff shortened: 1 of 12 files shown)


language/tools/move-cli/src/base/disassemble.rs
```diff
@@ -21,6 +21,9 @@ pub struct Disassemble {
     /// The name of the module or script in the package to disassemble
     #[clap(long = "name")]
     pub module_or_script_name: String,
+    /// Also print the raw disassembly using Rust's Debug output, at the end.
+    #[clap(long = "Xdebug")]
+    pub debug: bool,
```
amnn, line 26: This is not the main change, but it adds a flag to the disassembler (`--Xdebug`) that adds the raw `Debug` output of the `CompiledModule` or `CompiledScript` to the end of the usual non-interactive output.

Ideally, I would have added this to the transactional-test commands, but `move-cli/tests/build_tests` doesn't use this, and these tests are ideally placed there, because we need to use named addresses with numerical assignments.

amnn:
(Just tested a bump on `sui` and we will also need to canonicalize the identifier pool).

amnn:
Alright, we're good to go, this passes the regression test in https://github.com/MystenLabs/sui/pull/8248 and the repro instructions for building packages in https://github.com/capsule-craft/capsules from @PaulFidika.

language/tools/move-cli/src/base/disassemble.rs
tnowacki: What's the reasoning behind the name `Xdebug`?

amnn: I'm used to the `-X` flags on compilers being the hidden ones that you're not really supposed to use as a user of the compiler (i.e. they are more intended for the compiler authors) -- I think it's a JVM convention?

They are often also hidden from the help/usage output as well, but I couldn't get that to work with `clap_derive`.

I don't have a strong preference on it though -- if we're alright with this being here, we can drop the "X".

language/move-compiler/src/to_bytecode/canonicalize_handles.rs
```diff
@@ -0,0 +1,466 @@
...
+    for (ix, jx) in inverse.into_iter().enumerate() {
+        permutation[jx as usize] = ix as TableIndex;
+    }
+
+    permutation
+}
+
+/// Re-order `pool` according to the `permutation` array.  `permutation[i]` is the new location of
+/// `pool[i].
+fn apply_permutation<T>(pool: &mut Vec<T>, mut permutation: Vec<TableIndex>) {
```
tnowacki, line 415: It took me a little bit realize how this works lol, feels like a tortuous interview problem

tnowacki: But its a bit fun :) I only really understood how it worked when I sat down and said "how would I write this" and you start with the outer loop and quickly realize this is the only thing to do

language/move-compiler/src/to_bytecode/canonicalize_handles.rs
```diff
@@ -0,0 +1,466 @@
...
+        name: IdentifierIndex,
+    },
+    Unnamed,
+}
+
+/// Key for ordering function and struct handles, distinguishing handles for definitions in
+/// the module and handles for externally defined functions and structs.
+#[derive(Eq, PartialEq, Ord, PartialOrd)]
+enum ReferenceKey {
+    Internal(TableIndex),
```
tnowacki, line 52: Thanks <3

tnowacki (commented):
I think looks good, just some nits that might have helped with my immediate understanding of the code. 

Any thoughts though about assertions we can do to make sure nothing was missed? I just worry someone will add some new usage of a given table index and then not come update this code

language/tools/move-cli/src/base/disassemble.rs
tnowacki: Ah, that's a neat convention. I'm fine with it, we can always change it too. 
But maybe @tzakian has some thoughts?

tnowacki (approved):

language/move-compiler/src/to_bytecode/translate.rs
```diff
@@ -323,6 +325,24 @@ fn script(
     }))
 }
 
+/// Generate a mapping from numerical address and module name to named address, for module's whose
```
rvantonder, line 328:
```suggestion
/// Generate a mapping from numerical address and module name to named address, for modules whose
```

rvantonder (approved):
LGTM, I also tried some variations on the test input locally to check my thinking and validate. Appreciate the detailed comments in `canonicalize_headers`!

language/move-compiler/src/to_bytecode/translate.rs
```diff
@@ -298,17 +299,18 @@ fn script(
...
-        Err(e) => {
-            compilation_env.add_diag(diag!(
-                Bug::BytecodeGeneration,
-                (loc, format!("IR ERROR: {}", e))
-            ));
-            return None;
-        }
-    };
+    let (mut script, source_map) =
+        match move_ir_to_bytecode::compiler::compile_script(ir_script, deps) {
```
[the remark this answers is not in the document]
amnn: The trouble is that the `else` case needs to access the error anyway, so I don't think it helps a whole lot.  I do love `let ... else { }` though.

language/move-compiler/src/to_bytecode/canonicalize_handles.rs
```diff
@@ -0,0 +1,466 @@
...
+    External {
+        module: ModuleHandleIndex,
+        name: IdentifierIndex,
+    },
+}
+
+/// Forward the index at `$ix`, of type `$Ix` to its new location according to the `$perm`utation
+/// array.
+macro_rules! permute {
```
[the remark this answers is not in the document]
amnn: I can do `remap` -- I already use `apply_permutation` for something else.

language/move-compiler/src/to_bytecode/canonicalize_handles.rs
amnn: Stack Overflow suggests a different flavour of this algorithm: https://stackoverflow.com/questions/16501424/algorithm-to-apply-permutation-in-constant-memory-space

But, I found it easier to reason about this one.  It does have a flavour of an interview question -- when I did coding interview, I used to ask "Graph Clone", which is not too different from what we're doing here, except that we're cloning the graph onto itself!

amnn:
Thanks both!  Recording a summary of a conversation that I had with @tnowacki about the scenario where a new field gets added:

Yep, this is a real issue, and the best solution I can come up with is to adopt a visitor pattern that's used here and everywhere else that traverses bytecode indices in this way, so that folks only have to update one place.  Unfortunately, it seems like this kind of traversal isn't super common, so it doesn't buy us much to use a visitor pattern just here.

That being said, whenever I made a mistake while building this pass, I got a test failure in the Move testsuite, so we do have pretty good test coverage (usually it's a verifier test that fails which is reassuring, albeit with an invariant violation).

language/move-compiler/src/to_bytecode/translate.rs
```diff
@@ -323,6 +325,24 @@ fn script(
     }))
 }
 
+/// Generate a mapping from numerical address and module name to named address, for module's whose
+/// identities contained a named address.
+fn address_names<'a>(
+    dependencies: impl Iterator<Item = &'a ModuleIdent>,
+) -> HashMap<(MoveAddress, &'a str), Symbol> {
```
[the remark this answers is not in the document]
amnn: That's right, just querying here -- in general, I use `HashMap` when order shouldn't matter, and `BTreeMap` otherwise.  I feel like we could write a clippy rule or something for iterating over HashMap and HashSet?  I'll look into it.

PR #909, opened 2023-02-17
[move-prover] improve the way custom natives are handled in the workflow

meng-xu-cs:
This PR contains three changes:

- Refactored the native inst filter on Boogie template rendering, especially,
  this is to enable instantiations of two (or more) type parameters to pass
  to the template rendering engine.
- More coherent and natural flow of assigning the correct borrow edge,
  especially when encountering custom borrow primitives.
- Do not clean and optimize intrinsic borrow primitives.

## Motivation

Code consistency and coherence

### Have you read the [Contributing Guidelines on pull requests](https://github.com/move-language/move/blob/main/CONTRIBUTING.md#developer-workflow)?

Yes

## Test Plan

CI

PR #912, opened 2023-02-19
[cherry-pick][move-prover] better support for custom natives

meng-xu-cs:
This is to cherry-pick #909 which was landed on main. This is to further add prover support on the storage model.


awelc (approved):
LGTM!

PR #914, opened 2023-02-20
[move] Avoid using git for already fetched dependencies

awelc:
## Motivation

The goal of this PR is to resolve https://github.com/MystenLabs/sui/issues/8346 - it caches information about packages fetched during a given dependency graph construction to avoid unnecessarily calling into Git and, at least in some cases, avoid re-fetching the same dependency.

The assumption here is that custom package hooks are registered once before the dependency graph is constructed which ( I hope) seems reasonable, but if not, we can always exclude custom deps from this type of caching.

Yes

### DIFF
--- a/language/tools/move-package/src/resolution/dependency_cache.rs
+++ b/language/tools/move-package/src/resolution/dependency_cache.rs
@@ -1,0 +1,224 @@
+// Copyright (c) The Move Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+use anyhow::Result;
+use colored::Colorize;
+use std::{
+    collections::BTreeSet,
+    ffi::OsStr,
+    io::Write,
+    path::PathBuf,
+    process::{Command, Stdio},
+};
+
+use crate::{
+    package_hooks,
+    source_package::parsed_manifest::{DependencyKind, GitInfo, PackageName},
+};
+
+use super::repository_path;
+
+/// Fetches remote dependencies and caches information about those already fetched when building a
+/// given package.
+#[derive(Debug, Clone)]
+pub struct DependencyCache {
+    /// A path for remote dependencies that have already been fetched
+    fetched_deps: BTreeSet<PathBuf>,
+
+    /// Should a dependency fetched when building a different package be refreshed to the newest
+    /// version when building a new package
+    skip_fetch_latest_git_deps: bool,
+}
+
+impl DependencyCache {
+    pub fn new(skip_fetch_latest_git_deps: bool) -> DependencyCache {
+        let fetched_deps = BTreeSet::new();
+        DependencyCache {
+            fetched_deps,
+            skip_fetch_latest_git_deps,
+        }
+    }
+
+    pub fn download_and_update_if_remote<Progress: Write>(
+        &mut self,
+        dep_name: PackageName,
+        kind: &DependencyKind,
+        progress_output: &mut Progress,
+    ) -> Result<()> {
+        match kind {
+            DependencyKind::Local(_) => Ok(()),
+
+            DependencyKind::Custom(node_info) => {
+                // check if a give dependency type has already been fetched
+                if !self.fetched_deps.insert(repository_path(kind)) {
+                    return Ok(());
+                }
+                package_hooks::resolve_custom_dependency(dep_name, node_info)
+            }
+
+            DependencyKind::Git(GitInfo {
+                git_url,
+                git_rev,
+                subdir: _,
+            }) => {
+                let repository_path = repository_path(kind);
+                // check if a give dependency type has already been fetched
+                if !self.fetched_deps.insert(repository_path.clone()) {
+                    return Ok(());
+                }
+                let git_path = repository_path;
+                let os_git_url = OsStr::new(git_url.as_str());
+                let os_git_rev = OsStr::new(git_rev.as_str());
+
+                if !git_path.exists() {
+                    writeln!(
+                        progress_output,
+                        "{} {}",
+                        "FETCHING GIT DEPENDENCY".bold().green(),
+                        git_url,
+                    )?;
+
+                    // If the cached folder does not exist, download and clone accordingly
+                    Command::new("git")
+                        .args([OsStr::new("clone"), os_git_url, git_path.as_os_str()])
+                        .output()
+                        .map_err(|_| {
+                            anyhow::anyhow!(
+                                "Failed to clone Git repository for package '{}'",
+                                dep_name
+                            )
+                        })?;
+
+                    Command::new("git")
+                        .args([
+                            OsStr::new("-C"),
+                            git_path.as_os_str(),
+                            OsStr::new("checkout"),
+                            os_git_rev,
+                        ])
+                        .output()
+                        .map_err(|_| {
+                            anyhow::anyhow!(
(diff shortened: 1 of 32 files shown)


awelc:
Updated test output results to reflect the new caching data and most look good other than a few, for example this [one](Updated test output results to reflect the new caching data and most look good other than this [one](https://github.com/move-language/move/pull/914/files#diff-f0126da913fe5eb32ab0475f5885e3e7246ef7ea23ef47d0bbcaf319edd8e8b7) which I am not 100% sure about) or this [one](https://github.com/move-language/move/pull/914/files#diff-0387d7f89a909ca0f26a62a84d1a0d4071b4be4bdce42de4c6bc014be21b8373). Will look into these tomorrow.

language/tools/move-package/tests/test_sources/nested_deps_bad_parent/Move.resolved
```diff
@@ -1 +1 @@
-Failed to resolve dependencies for package 'NestedDeps': Resolving dependencies for package 'Nested': Path cannot access parent of current directory: ../out_of_repo
+Failed to resolve dependencies for package 'NestedDeps': Fetching 'Nested': Failed to reset to latest Git state 'ed613ee8e' for package 'Nested', to skip set --skip-fetch-latest-git-deps | Exit status: exit status: 128
```
amnn, line 1: Yep, this seems not correct (as you mentioned in your comment).

language/tools/move-package/tests/test_sources/nested_deps_git_local/Move.resolved
```diff
@@ -55,6 +55,22 @@ ResolvedGraph {
             "MoveStdlib",
             "NestedDeps",
         },
+        fetched_deps: {
+            Git(
+                GitInfo {
+                    git_url: "https://github.com/move-language/move",
+                    git_rev: "781c844",
+                    subdir: "language/move-stdlib",
```
amnn, line 63: The `subdir` should not be part of your cache key, only the URL of the remote, and its revision.  `download_and_update_if_remote` already calculates (in effect) a kind of string-based cache key, to form the directory name to download the remote dependencies into, I would suggest re-using that.

language/tools/move-package/tests/test_sources/external_dev_dep/Move.progress
```diff
@@ -1,11 +1,11 @@
-RESOLVING DEPENDENCIES IN A FROM Root WITH ../resolvers/successful.sh
-../resolvers/successful.sh stderr:
+[1;32mRESOLVING DEPENDENCIES IN[0m A [1;32mFROM[0m Root [1;32mWITH[0m ../resolvers/successful.sh
```
amnn, line 1: Also agreed on this one -- not sure what has caused it to change its output (it is in a sense legitimate, because the escape codes do exist to colour things green, but I don't know why your change has prompted them to start appearing in tests).

amnn (commented):
Thanks for making a start on this!  I think it's broadly there, but in terms of code organising I would suggest:

- `DependencyGraph` is the wrong place to store the set of fetched dependencies.
- Consider constructing a new module and type to store this (`PackageCache`?), which takes on the responsibility of `download_and_update_if_remote`.  The `skip_fetch_latest_git_deps` configuration should also go to it, and I believe the `writer` is also intended for it as well? 
- This type can be passed by mutable reference to both `DependencyGraph` and `ResolutionGraph`, and then discarded.
- Rather than using `DependencyKind` as the cache key, use the names of the directories that remote packages get fetched into, this avoids re-fetching when two packages come from the same git repo/revision, or redundantly keeping track of local dependencies which are not being fetched at all.

Also you're right that some of the test changes are fishy -- thanks for catching that.

language/tools/move-package/tests/test_sources/nested_deps_git_local/Move.resolved
awelc: I re-used the entire `DependencyKind` as it seemed like we may have to distinguish between `Custom` and `Local` variants in case they happen to point to the same location, and simply using enum as the key seemed simplest. Looking more closely, particularly at the `repository_path` function whose result could be used at key, it seems like we don't have to worry about that.

awelc:
> * `DependencyGraph` is the wrong place to store the set of fetched dependencies.

Why is it so bad? It's available in all places where we need to access the cache and, conceptually, it deals with dependencies so keeping this cache there does not seem that bad. It will indeed be kept until the dependency graph itself is discarded but at the same time we don't have to worry about explicitly managing its lifetime.

amnn:
>> - DependencyGraph is the wrong place to store the set of fetched dependencies.
>
> Why is it so bad?

It mostly boils down to separation of concerns, but that manifests in a couple of ways:

- `DependencyGraph` is the in-memory representation of a lock file -- so there is a 1:1 correspondence between those two things, which breaks with the addition of `fetched_deps`.  Not a big deal, but then every special case we add requires explanation, and their existence invites bugs from missing correspondences.
- `ResolutionGraph` also needs to be able to fetch dependencies, so it seems strange to encapsulate the logic for that within one of its immutable inputs (Something that I think you noticed and worked around by cloning the fetched dependencies).

awelc:
Clearly you have it thought over quite a bit more than I do :-) I separated the cache from the dependency graph but left the writer as argument to `download_and_update_if_remote` as it is also used in some random places during the resolution process so could not be isolated inside `DependencyCache` anyway.

language/tools/move-package/src/resolution/dependency_cache.rs
```diff
@@ -0,0 +1,224 @@
...
+    source_package::parsed_manifest::{DependencyKind, GitInfo, PackageName},
+};
+
+use super::repository_path;
+
+/// Fetches remote dependencies and caches information about those already fetched when building a
+/// given package.
+#[derive(Debug, Clone)]
+pub struct DependencyCache {
+    /// A path for remote dependencies that have already been fetched
```
amnn, line 25: nit

```suggestion
    /// A set of paths for remote dependencies that have already been fetched
```

language/tools/move-package/src/source_package/parsed_manifest.rs
```diff
@@ -57,14 +57,14 @@ pub struct InternalDependency {
     pub digest: Option<PackageDigest>,
 }
 
-#[derive(Debug, Clone, Eq, PartialEq)]
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd)]
```
amnn, line 60: Are these derives necessary anymore?

amnn (approved):
This looks great, thanks for working on this Adam :)

> but left the writer as argument to download_and_update_if_remote as it is also used in some random places during the resolution process so could not be isolated inside DependencyCache anyway

Yes, I think that is the right call -- I unfortunately use the same progress output for outputting progress from external package resolution, and it seems like it's not straightforward to, e.g. clone the kinds of writers we care about (e.g. `Stderr`).

language/tools/move-package/src/source_package/parsed_manifest.rs
awelc: Sorry, missed this during cleanup - clearly no longer necessary as we don't use `DependencyKind` as a key anymore. Thank you for catching this!

PR #915, opened 2023-02-21
[status code] add EXECUTION_LIMIT_REACHED, IO_LIMIT_REACHED and STORAGE_LIMIT_REACHED

vgao1996:
This adds a few status codes that can be used by downstream clients to represent errors caused transactions exceeding certain resource limits.

### DIFF
--- a/language/move-core/types/src/vm_status.rs
+++ b/language/move-core/types/src/vm_status.rs
@@ -159,6 +159,20 @@ impl VMStatus {
                 ..
             }
             | VMStatus::Error(StatusCode::OUT_OF_GAS) => Ok(KeptVMStatus::OutOfGas),
+
+            VMStatus::ExecutionFailure {
+                status_code:
+                    StatusCode::EXECUTION_LIMIT_REACHED
+                    | StatusCode::IO_LIMIT_REACHED
+                    | StatusCode::STORAGE_LIMIT_REACHED,
+                ..
+            }
+            | VMStatus::Error(
+                StatusCode::EXECUTION_LIMIT_REACHED
+                | StatusCode::IO_LIMIT_REACHED
+                | StatusCode::STORAGE_LIMIT_REACHED,
+            ) => Ok(KeptVMStatus::MiscellaneousError),
+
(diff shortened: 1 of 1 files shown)


wrwg:
I wonder as you are with this, could we also do https://github.com/aptos-labs/aptos-core/issues/6347. It is rather important and I wish we had done it already a while ago.

movekevin:
> I wonder as you are with this, could we also do [aptos-labs/aptos-core#6347](https://github.com/aptos-labs/aptos-core/issues/6347). It is rather important and I wish we had done it already a while ago.

+1

vgao1996:
@wrwg @movekevin we are already under a tight schedule (branch cut on Wednesday) to land the mission-critical gas changes so I'd suggest we try not to squeeze this in. I like the idea of having adapter-reserved status codes though.

vgao1996:
In other words, I have a feeling that the ADAPTER_CODEs may require some special conversion logic + special handling by the clients (wallet, indexer etc.). Let's study those a bit more and then come up with a proper implementation.

wrwg:
> @wrwg @movekevin we are already under a tight schedule (branch cut on Wednesday) to land the mission-critical gas changes so I'd suggest we try not to squeeze it in this time. I like the idea of having adapter-reserved status codes though.

Fixing this bug I linked is a release blocker by itself. But if you do not want to fold it in here, sure go ahead and someone else will do it.

language/move-core/types/src/vm_status.rs
```diff
@@ -679,6 +693,9 @@ pub enum StatusCode {
     STORAGE_WRITE_LIMIT_REACHED = 4027,
     MEMORY_LIMIT_EXCEEDED = 4028,
     VM_MAX_TYPE_NODES_REACHED = 4029,
+    EXECUTION_LIMIT_REACHED = 4030,
```
wrwg, line 696: FWIW new status codes need to be flag protected as once they are used, we cannot roll back (had an SEV for this). I guess you are aware.

wrwg (approved):

msmouse (approved):

PR #917, opened 2023-02-21
 [printing] fix bug in StructTag::struct_to_canonical_string

sblackshear:
Previously, a type instantiation w/ >1 type parameters was printed without a separator (e.g., `S<u64bool>` instead of `S<u64,bool>`for `S<T1,T2>` instantiated with `u64` and `bool`). This PR adds a comma separator and a test.


awelc (approved):
LGTM

oxade (approved):

Issue #916, opened 2023-02-21
[Bug] std::type_name::into_string does not handle multiple generics well


PaulFidika:
When I define a struct like this:

`struct SillyStruct<phantom A, phantom B, phantom C> has drop { }`

And then try to get the type name, like this:

`sui_utils::encode::type_name<SillyStruct<SUI, u64, vector<u8>>>()`

I get an output like this:

`77799ff1bc7ca73c5d9e8b5ff85362a5681963d4::encode_test::SillyStruct<0000000000000000000000000000000000000002::sui::SUIu64vector<u8>>`

The problem is that the generics are all smooshed together; you would expect "SUI, u64, vector<u8>" (spaces) or "SUI,u64,vector<u8>" (no spaces). But as it is now it's impossible to parse the generics, which is something I'm trying to do. This is only an issue for structs with multiple generics, which are uncommon, which is probably why it hasn't been encountered yet.

sui_utils can be found here, although it's simply calling into the existing std::type_name::get<T>() function: https://github.com/capsule-craft/capsules/blob/master/packages/sui_utils/sources/encode.move

Thanks.

PR #918, opened 2023-02-21
[printing] fix bug in StructTag::struct_to_canonical_string

sblackshear:
Previously, a type instantiation w/ >1 type parameters was printed without a comma separator (e.g., `S<u64bool>` instead of `S<u64,bool>`for `S<T1,T2>` instantiated with `u64` and `bool`). This PR adds a comma separator and a test.

(reported in https://github.com/move-language/move/issues/916)

Adding Aptos folks as reviewers to advise on whether this fix is ok as-is or creates a breaking change that should be managed.

### DIFF
--- a/language/move-core/types/src/language_storage.rs
+++ b/language/move-core/types/src/language_storage.rs
@@ -126,8 +126,9 @@ impl StructTag {
 
     /// Return a canonical string representation of the struct.
     /// Struct types are represented as fully qualified type names; e.g.
-    /// `00000000000000000000000000000001::string::String` or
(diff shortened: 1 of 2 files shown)


vgao1996 (approved):
LGTM @alinush not sure if this is related to your work

alinush:
I'll have to check when I get back home. From a quick glance, I don't remember if I implemented that function or not.

alinush:
Indeed, not related to my work; see my (unrelated) changes to `language_storage.rs` [here](https://github.com/move-language/move/pull/630/files#diff-87f239c60b70ccf3a079ee9dc228bed08bbdbbdb1801d52f6df0986163ab738b).

So LGTM

PR #906, opened 2023-02-17
[Cherry-pick][Move-package] Enable bytecode as dependencies

rahxephon89:
## Motivation

Cherry-pick #865 onto `main`

### Have you read the [Contributing Guidelines on pull requests](https://github.com/move-language/move/blob/main/CONTRIBUTING.md#developer-workflow)?

Yes

PR #923, opened 2023-02-22
[Move-package][Cherry-pick] Restrict searching bytecode within the build folder

rahxephon89:
## Motivation

Cherry-pick #922, which is a small fix from #906 to only search bytecode within the build folder.


amnn (approved):
LGTM

PR #924, opened 2023-02-22
[Move-prover] Fix duplicated table function instances in boogie

rahxephon89:
## Motivation

Close the issue [here](https://github.com/aptos-labs/aptos-core/issues/6723).


junkil-park (approved):
Great! I resolves the issue. Thanks!

PR #929, opened 2023-02-24
[move-package] Support overlapping resolution graphs

awelc:
## Motivation

In the current implementation of external package resolution, the dependency graphs produced by each resolver invocation (each external resolver, and the internal resolver) must be pairwise disjoint.

This restriction needs to be loosened to allow overlap between the outputs of resolution invocations when the overlap is consistent to avoid problems in case the same important package (e.g., framework package) is a dependency of both a package provided by third-party resolver and a dependency explicitly specified by the programmer

Yes

## Test Plan

Tests added and modified accordingly


language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -670,15 +664,56 @@ impl DependencyGraph {
         let manifest = parse_move_manifest_from_file(&pkg_path)
             .with_context(|| format!("Parsing manifest for '{}'", name))?;
 
-        let kind = pkg.kind.clone();
-        self.extend_graph(
-            &kind,
+        let sub_graph = DependencyGraph::new(
```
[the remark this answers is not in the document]
awelc: My thinking was that we'd want to a "disjoint" graph for comparison rather than the entire "merged" one. Wasn't sure it would make a difference but was erring on the side of caution.

awelc:
> I think the approach you're taking here _almost_ works, but

Yeah, I kind of realized this once I prototyped it and only then started thinking about cycles...

> let me know what you think:

Sounds simple once the spark materializes so I am glad it happened :-) - trying it now. For better or worse I like working this way - prototype something and see if it works and use it to come up with something better if need be.

language/tools/move-package/tests/test_dependency_graph.rs
```diff
@@ -293,11 +293,7 @@ fn merge_overlapping() {
     )
     .expect("Reading inner");
 
-    let Err(err) = outer.merge(inner, Symbol::from("")) else {
-        panic!("Outer mentions package A, and so does inner.");
-    };
-
-    assert_error_contains!(err, "Conflicting dependencies found");
+    assert!(outer.merge(inner, Symbol::from("")).is_ok());
```
amnn, line 296: Could you also add a test where the merge does genuinely fail?

language/tools/move-package/tests/test_sources/external_overlap_fail/Move.resolved
```diff
@@ -0,0 +1,3 @@
+Failed to resolve dependencies for package 'Root': Adding dependencies from ../resolvers/successful.sh for dependency 'A' in 'Root': Conflicting dependencies found:
+ADep = { local = "deps_only/ADep" }
+ADep = { local = "./deps_only/ADep" } # Resolved by ../resolvers/successful.sh
```
amnn, line 3: We should probably bolster this error now -- the issue is not with `ADep` but the fact that its own dependencies differ, so we should share that information to developers.

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -904,3 +945,23 @@ fn str_escape(s: &str) -> Result<String, fmt::Error> {
 fn path_escape(p: &Path) -> Result<String, fmt::Error> {
     str_escape(p.to_str().ok_or(fmt::Error)?)
 }
+
+/// Checks if dependencies of a given package in two different dependency graph maps are the
+/// same.
+fn pkg_deps_equal_new(
```
amnn, line 951: To support the comment below about more robust error message, this function will need to return the delta between the outbound edges in these graphs, rather than just a boolean.

```
fn diff_package_dependencies(
    pkg_name: PM::PackageName,
    graph_a: &DiGraphMap<PM::PackageName, Dependency>,
    graph_b: &DiGraphMap<PM::PackageName, Dependency>,
) -> /* ... not sure what a good return type would be here ... */
```

amnn (commented):
I think this has come out quite cleanly, thanks for working on the fix-ups!  We'll need to improve the error message when we can't overlap the graphs, but then it'll be good to go.

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -402,16 +433,19 @@ impl DependencyGraph {
...
-                        PackageWithResolverTOML(&ext_pkg),
-                    );
+                // Seeing the same package in `extension` is OK only if it has the same set of
+                // dependencies as the existing one.i
+                Entry::Occupied(_) => {
+                    let (self_deps, ext_deps) =
+                        pkg_deps_equal(ext_name, &self.package_graph, &ext_graph);
+                    if self_deps != ext_deps {
+                        bail!(
+                            "An already resolved package {} encountered during external resolution but with different dependencies:\n existing deps: {:?}\n external deps: {:?}",
```
amnn, line 443: The error message should contain the following pieces of information:

- The list of mismatched dependencies on each side, rather than just the mismatched package being depended on. See `DependencyTOML` for a wrapper type that formats this kind of data (package name + dependency) in a human intelligible way.
- The external resolver being used.
- The package that's depending on these other packages.

Because otherwise there won't be sufficient context for someone to know where to look to fix the error.  Here's a sample error message to use as a template, that kind of matches what we had before, but with the additional information:

```
Conflicting dependencies found for 'ADep' during external resolution by 'resolver':
  Existing dependencies not found:
    ADepDep = { ... }

  New external dependencies:
    ADepDep = { ... }
```

And we should not display either of `Existing dependencies not found` or `New external dependencies` if the list is empty.

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -904,3 +947,31 @@ fn str_escape(s: &str) -> Result<String, fmt::Error> {
...
+        .difference(&intersection)
+        .cloned()
+        .map(|(_, pkg)| pkg)
+        .collect();
+    let other_deps = other_edges
+        .difference(&intersection)
+        .cloned()
+        .map(|(_, pkg)| pkg)
+        .collect();
+    (pkg_deps, other_deps)
```
amnn, line 976: nit: Might be neater to use `symmetric_difference` and `partition`?

```suggestion
    let pkg_edges = BTreeSet::from_iter(pkg_graph.edges(pkg_name).map(|(_, pkg, dep)| (dep, pkg));
    let other_edges = 
        BTreeSet::from_iter(other_graph.edges(pkg_name).map(|(_, pkg, dep)| (dep, pkg);
    
    pkg_edges
        .symmetric_difference(other_edges)
        .partition(|dep| pkg_edges.contain(dep));
```

Also, I would recommend returning the whole dependency (both package name and the dependency), because the mismatch could be in the `Dependency` part, in which case we want to avoid a confusing error message where it says there's a mismatch but shows no mismatched packages.

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -904,3 +946,41 @@ fn str_escape(s: &str) -> Result<String, fmt::Error> {
...
+    let other_edges =
+        BTreeSet::from_iter(other_graph.edges(pkg_name).map(|(_, pkg, dep)| (dep, pkg)));
+
+    let (pkg_deps, other_deps): (Vec<_>, Vec<_>) = pkg_edges
+        .symmetric_difference(&other_edges)
+        .partition(|dep| pkg_edges.contains(dep));
+    (
+        pkg_deps
+            .iter()
+            .map(|(dep, pkg)| format!("{}", DependencyTOML(*pkg, dep)))
```
amnn, line 979: nit: Seeing as you have a separate function for formatting, I would maybe put the logic of calling `format!` into there.

amnn (approved):
Looks beautiful!  Left a small nit about where the formatting logic goes, but otherwise, this is good to go, thanks for working on this @awelc !

language/tools/move-package/src/resolution/dependency_graph.rs
awelc: My Rust mojo definitely still has gaps and I could not easily return `Dependency` from a function, presumably because it contains (optional) `BTreeMap`s...

awelc: Never mind, I think I figure it out.

amnn: It's possible that you could get into issues with temporaries if returning by value, but I think returning by reference is the right call here, in any case!

PR #935, opened 2023-02-26
[chore] implement GetModule for all ModuleResolvers

sblackshear:
This allows us to use GetModule with less boilerplate in some other places where it's needed--needed to unblock some testing in the `sui` repo.


rvantonder (approved):
Looks like a straightforward change to me, approving to unblock!

PR #936, opened 2023-02-26
[vm errors] Propagate status message to adapters

wrwg:
This makes the internal status message of VM errors available in `VMStatus` and accessible by adapters. The purpose is to allow adapters to show more detail of errors like they appear e.g. during bytecode verification. These errors happen more often in practice as anticipated, mainly because uses operate with different versions of networks and tools, e.g. deploying code to some chain where it is not deserializable, or using wrong dependencies on disk which do not match those on the network. Currently, the errors produced by the system are as unspecific as 'LINKER_ERROR'.

### DIFF
--- a/language/move-core/types/src/vm_status.rs
+++ b/language/move-core/types/src/vm_status.rs
@@ -57,7 +57,7 @@ pub enum VMStatus {
     /// Indicates an error from the VM, e.g. OUT_OF_GAS, INVALID_AUTH_KEY, RET_TYPE_MISMATCH_ERROR
     /// etc.
     /// The code will neither EXECUTED nor ABORTED
-    Error(StatusCode),
+    Error(StatusCode, Option<String>),
 
     /// Indicates an `abort` from inside Move code. Contains the location of the abort and the code
     MoveAbort(AbortLocation, /* code */ u64),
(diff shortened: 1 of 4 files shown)


wrwg:
Related issues:

https://github.com/aptos-labs/aptos-core/issues/5864
https://github.com/aptos-labs/aptos-core/issues/6232
https://github.com/aptos-labs/aptos-core/issues/6041

msmouse (approved):

language/move-core/types/src/vm_status.rs
```diff
@@ -57,7 +57,7 @@ pub enum VMStatus {
     /// Indicates an error from the VM, e.g. OUT_OF_GAS, INVALID_AUTH_KEY, RET_TYPE_MISMATCH_ERROR
     /// etc.
     /// The code will neither EXECUTED nor ABORTED
-    Error(StatusCode),
+    Error(StatusCode, Option<String>),
```
davidiw, line 60: Do we need to roll out a backwards compatible version of this? Or change the serialization behaviors so we don't embed the String?

davidiw (commented):
nice

language/move-core/types/src/vm_status.rs
wrwg: Yes, but this will happen in aptos-core, not in this repo.

movekevin: How would we gate this on the aptos side? Aren't many of these errors thrown inside the verifier or interpreter?

wrwg: Yep, we don't. We are converting this to our representation. We can decide on our side what we do with this new piece of information.

zekun000 (approved):

movekevin (approved):

PR #940, opened 2023-02-27
Charge native inline within budget

oxade:
## Motivation

This PR sets the budget for native function and breaks down charging to ensure the native functions don't do work which the client cannot afford.

Yes
## Test Plan

N/A

### DIFF
--- a/language/move-vm/runtime/src/native_functions.rs
+++ b/language/move-vm/runtime/src/native_functions.rs
@@ -8,6 +8,7 @@ use crate::{
 use move_binary_format::errors::{ExecutionState, PartialVMError, PartialVMResult};
 use move_core_types::{
     account_address::AccountAddress,
+    gas_algebra::InternalGas,
     identifier::Identifier,
     language_storage::TypeTag,
     value::MoveTypeLayout,
@@ -94,6 +95,7 @@ pub struct NativeContext<'a, 'b> {
     data_store: &'a mut dyn DataStore,
     resolver: &'a Resolver<'a>,
     extensions: &'a mut NativeContextExtensions<'b>,
+    gas_budget: InternalGas,
 }
 
 impl<'a, 'b> NativeContext<'a, 'b> {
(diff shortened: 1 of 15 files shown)


language/move-vm/runtime/src/native_functions.rs
```diff
@@ -172,4 +176,35 @@ impl<'a, 'b> NativeContext<'a, 'b> {
...
+    }};
+}
+
+/// Total cost of a native call so far
+#[macro_export]
+macro_rules! native_gas_total_cost {
+    ($native_context:ident, $gas_left:ident) => {{
+        // Its okay to unwrap because the budget can never be less than the gas left
+        $native_context.gas_budget().checked_sub($gas_left).unwrap()
+    }};
```
dariorussi, line 209: so the model is that 
1. we grab all gas left on entry (say gas budget `100`) 
2. we call `native_charge_gas_early_exit` for operations we want to meter, and if that goes out of gas we stop. So say we call it twice with 10, we end up with `gas_left` being 80. If we pass `100` we fail right there
3. at the end of a native we call `native_gas_total_cost` to report the cost outside the function into the interpreter. And the cost is budget - whatever is left, in the example 100 - 80 = 20 which is the cost.

Is that right?

dariorussi (approved):
Looks good, I left a comment to make sure I understood correctly

language/move-vm/runtime/src/native_functions.rs
oxade: > so the model is that
> 
> 1. we grab all gas left on entry (say gas budget `100`)
> 2. we call `native_charge_gas_early_exit` for operations we want to meter, and if that goes out of gas we stop. So say we call it twice with 10, we end up with `gas_left` being 80. If we pass `100` we fail right there
> 3. at the end of a native we call `native_gas_total_cost` to report the cost outside the function into the interpreter. And the cost is budget - whatever is left, in the example 100 - 80 = 20 which is the cost.
> 
> Is that right?

Precisely!

oxade: > so the model is that
> 
> 1. we grab all gas left on entry (say gas budget `100`)
> 2. we call `native_charge_gas_early_exit` for operations we want to meter, and if that goes out of gas we stop. So say we call it twice with 10, we end up with `gas_left` being 80. If we pass `100` we fail right there
> 3. at the end of a native we call `native_gas_total_cost` to report the cost outside the function into the interpreter. And the cost is budget - whatever is left, in the example 100 - 80 = 20 which is the cost.
> 
> Is that right?

Now on further thought, I can the `gas_left` inside `NativeContext`, which will make things cleaner.
I can tackle this in a subsequent PR

PR #952, opened 2023-03-01
Revert "Rollforward "[vm errors] Propagate status message to adapters""

davidiw:
Reverts move-language/move#949

Let's wait until we cleanly make it through the branch cut before adding in new features :pray:


movekevin (approved):

vgao1996 (approved):

wrwg:
This is the wrong approach. Instead we should create aptos-1.3 similar as we did with 1.2. Then all this roll back/forward/back is not needed.

PR #954, opened 2023-03-02
[gas] fix NativeContext::exists_at & introduce new gas parameter to fix buggy gas formula in the table extension

vgao1996:
This fixes two issues:
1. `NativeContext::exists_at` should not have flattened the nested options. 
2. The table extension does not waive the base fee even when the resource being loaded is already in the cache.


msmouse (approved):
So we can't distinguish between execution and io here?

vgao1996:
Sadly we cannot

PR #963, opened 2023-03-04
add VMRuntimeLimitsConfig and vector len limit

oxade:
## Motivation

Add `VMRuntimeLimitsConfig` which exposes `vector_len_max` and allows for bounding vec len at runtime. 
Add verifier check to ensure constant vector sizes limit.

Yes
## Test Plan

push_back tests of limit at and above


dariorussi (approved):
great!

PR #964, opened 2023-03-04
[move-package] Support dependency overrides

awelc:
## Motivation

This PR addresses problem with package versioning introducing potential for diamond problems (two sibling dependencies with a shared dependency whose version does not match).

Yes

## Test Plan

Respective tests have been added


language/tools/move-package/src/source_package/parsed_manifest.rs
```diff
@@ -55,15 +56,31 @@ pub struct InternalDependency {
...
     Git(GitInfo),
     Custom(CustomDepInfo),
 }
 
+/// Custom implementation to normalize local paths
+impl PartialEq for DependencyKind {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (&DependencyKind::Local(ref p), &DependencyKind::Local(ref op)) => {
+                normalize_path(p, true).unwrap() == normalize_path(op, true).unwrap()
```
amnn, line 74: Is it possible to ensure local paths are stored normalized?  No worries if not, but I remember some other occasions where I ran into this, and that's how I solved it (by calling `normalize_path`).

language/tools/move-package/src/resolution/lock_file/schema.rs
```diff
@@ -58,6 +58,9 @@ pub struct Dependency {
 
     /// Expected hash for the source and manifest of the package being depended upon.
     pub digest: Option<String>,
+
+    /// Optional dependency override to handle dependency conflicts.
+    pub dep_override: Option<bool>,
```
amnn, line 63: I don't think dep overrides should be necessary in the lock file -- i.e. the lock file should only contain one package for each package name.

language/tools/move-package/src/source_package/parsed_manifest.rs
```diff
@@ -55,15 +56,31 @@ pub struct InternalDependency {
     pub subst: Option<Substitution>,
     pub version: Option<Version>,
     pub digest: Option<PackageDigest>,
+    pub dep_override: Option<DepOverride>,
```
amnn, line 59: nit: This doesn't need to be an `Option<DepOverride>` we can just have it be a `bool` and parse `None` as `false` (and display `false` as no field, and `true` as `override = true`).

Also, I don't feel strongly about this, but I found the alias of `DepOverride = bool` a little confusing while reading -- I was expecting a bool but found what I thought was a struct, and was left wondering what other information it needed to carry.

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -415,6 +438,23 @@ impl DependencyGraph {
             bail!("Merging dependencies into a graph after calculating its 'always' dependencies");
         }
 
+        // Because all the packages in `extensions`'s package table didn't exist in `self`'s, all
```
amnn, line 441: After your change to handle overlapping graphs, is this still the case?  I think the doc-comment on `merge` is now out-of-date.

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -428,41 +468,53 @@ impl DependencyGraph {
...
-                        pkg_deps_equal(ext_name, &self.package_graph, &ext_graph);
-                    if self_deps != ext_deps {
-                        bail!(
-                            "Conflicting dependencies found for '{ext_name}' during external resolution by '{resolver}':\n{}{}",
-                            format_deps("\nExternal dependencies not found:", self_deps),
-                            format_deps("\nNew external dependencies:", ext_deps),
-                        );
+            let pkg = match self.package_table.get(&ext_name) {
+                None => ext_pkg,
+                Some(existing_pkg) => {
```
amnn, line 473: nit: You should still be able to use the `entry` API for this -- `Entry::Occupied` contains the existing package.

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -428,41 +468,53 @@ impl DependencyGraph {
...
-        // without worrying about introducing duplicate edges.
-        for (from, to, dep) in ext_graph.all_edges() {
-            self.package_graph.add_edge(from, to, dep.clone());
+        // finalize all edges
+        for (from, to, _) in ext_graph.all_edges() {
+            self.package_graph
+                .edge_weight_mut(from, to)
+                .unwrap()
+                .finalized = true;
         }
```
amnn, line 510: nit: there is an [`all_edges_mut`](https://docs.rs/petgraph/latest/petgraph/graphmap/struct.GraphMap.html#method.all_edges_mut):

```suggestion
        for (_, _, e) in ext_graph.all_edges_mut() {
            e.finalized = true;
        }
```

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -684,22 +743,24 @@ impl DependencyGraph {
         external_requests: &mut Vec<ExternalRequest>,
         progress_output: &mut Progress,
     ) -> Result<()> {
-        let pkg = match self.package_table.entry(name) {
-            Entry::Vacant(entry) => entry.insert(pkg),
+        let pkg = match self.package_table.get(&name) {
+            None => pkg,
```
amnn, line 747: ditto to the comment above -- could you elaborate on why you needed to move away from the `entry` API?  I may be missing a detail here, but this case I think could also still use `entry`.

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -969,12 +1045,183 @@ fn pkg_deps_equal<'a>(
...
+    );
+}
+
+/// Find the closest valid (dominating) override for a given package.
+fn find_override(
+    graph: &Graph<Symbol, Dependency>,
+    root_node: NodeIndex<u32>,
+    pkg_node: NodeIndex<u32>,
+) -> Result<Option<NodeIndex>> {
+    let all_dominators = algo::dominators::simple_fast(&graph, root_node);
```
amnn, line 1127: This is going to result in calculating a lot of dominator trees.  Calculating dominators is in the best case almost linear time, but I think the number of times this gets called makes the overall cost in the algorithm quadratic, which I think is avoidable.

amnn (commented):
Left some implementation-level thoughts and questions as inline comments, but high level this has also served as some inspiration for a potentially simpler and more efficient solution, let me know what you think:

Instead of processing dependencies and then calculating dominators and using that to figure out which override is in effect, why not leverage the order of traversal to keep track of a set of overrides that are currently in effect? I.e.

- When extending a graph with its dependencies, record which of its dependencies are overrides first, and maintain them in a set (if a package is already present in that, then don't add it again, because it means that it's already been overriden higher up).
- When processing a dependency, if it is already in the overriden set, don't process it -- we know that it will be met by an override.
- After processing all the dependencies of a package, remove its overrides from the transient set, because they are no longer in effect.
- When merging graphs, we will need to heed the set of overrides -- i.e. remove the sub-graphs rooted at overriden packages (it may be possible to do this by just removing the overridden packages in the graph to be merged in, and doing a final pass to remove unreachable packages).
- Logic for handling package conflicts should be able to remain the same, same is true for the cycle check.

I think this achieves the same thing without having to recompute dominators when a potential conflict is found, and it should be a smaller change, because all the information is held in a transient data structure, not maintained in terms of `finalized` flags.

The most interesting edge case to look at (and I haven't fully fleshed this out yet) is this one:

```
           ┌─A─┐
           ▼   ▼
         ┌─B   C
         │ │   │
         │ └►D◄┘
override │   │
         │   ▼
         └──►E
```

With what I'm proposing if we process `B` first, we will override `E` as a dependency of `D`, but `D` and `E` are not dominated by `B`, so we will need some mechanism for `C` to know that it needs to check that `D` and `E` are consistent with its own dependency sub-graph (i.e. `C` should not simply skip `D` because it already exists in the graph).

language/tools/move-package/src/source_package/parsed_manifest.rs
awelc: > Is it possible to ensure local paths are stored normalized?

That was my first thought, but this I believe would require changing the code parsing the manifest file and I wasn't sure we wanted to go there. The comparison change seemed less invasive, but I am happy to go the other route if it looks better.

language/tools/move-package/src/resolution/lock_file/schema.rs
awelc: The thought here (though not fully implemented yet) was that there may be an override encoded in the dependency graph coming from an external resolver. In particular, for the following manifest for a package involving a diamond problem down the line (e.g., https://github.com/move-language/move/pull/964/files#diff-8c1d4a846826c7a3c78ecd708834e6d958c922a948298f9279e402ab3b30ea37) , instead of overriding `B` an external resolver could conceivably contain an override for `A`.

If we don't worry about this case (I think the external override would only make sense at the level of external root package) then we indeed don't need overrides in Move.lock files.

language/tools/move-package/src/source_package/parsed_manifest.rs
awelc: Makes sense!

language/tools/move-package/src/resolution/dependency_graph.rs
awelc: I am not sure which part of `merge`'s doc comment you have in mind. The one on line 441 above is still correct I think as we just moved adding edges from extension graph to the end of graph processing so the invariant about edges not existing in the original graph still holds.

language/tools/move-package/src/resolution/dependency_graph.rs
awelc: I was hitting some problems related to borrowing and at some point I gave up. If we decide to use the current solution and this code stays, I will try harder!!!

awelc: I went ahead and changed it back. Clearly this was some problem mid-hacking as it was a rather easy change...

language/tools/move-package/src/source_package/parsed_manifest.rs
awelc: Let's settle on normalizing local paths.

language/tools/move-package/src/resolution/lock_file/schema.rs
awelc: Let's not worry about external resolvers specifying overrides (for now).

language/tools/move-package/src/resolution/dependency_graph.rs
awelc: Document for future cleanup.

awelc:
Overrides are specified explicitly, rather than what the previous solution assumed, that is implicitly. In other words to override some dependency `C` that `A` and `B` depend on (and conflict on), we specify a version of `C` to be used in the root package's manifest file:
```
[dependencies]
A = { local = "./deps_only/A" }
B = { local = "./deps_only/B" }
C = { local = "./deps_only/C", version = "1.0.0", override = true }
```
rather than, what was assumed before, specifying which dependency path is supposed to designate the "winning" version of the package:
```
[dependencies]
A = { local = "./deps_only/A", override = true }
B = { local = "./deps_only/B" }
```
The updated algorithm is as follows (repeated here from the doc-comment for the `override_pkg` function.

The algorithm relies on the fact that this check is performed when a new  package is added to the graph already containing package with the same name - this allows us to resolve the conflicts pair-wise even if ultimately a package with a given name is a dependency of more than two other packages.
                                                                                               
First we check if package already existing in the graph has been inserted there as a result of  an override. If it was, then either:                                                           
- the new package is dominated by the same override                                            
or                                                                                             
- no valid override can be found and error must be reported                                    
                                                                                               
Because the overridden dependencies are processed first for each package, it is guaranteed that when searching for overrides for a child (new package), a correct override will be found if it exists because by construction it must be already present in the graph.                        
                                                                                               
If package already existing in the graph has not been inserted as a result of an override then  either:                                                                                        
- the new package has been inserted as a result of an override and:                            
    - it's dominating override for the existing package then new package is the winner
    or
    - it's not a dominating override for the existing package then an error must be reported   
or                                                                                             
- the new package has not been inserted as a result of an override and an error must be reported
                                                                                               
Correctly finding dominators only works because when trying to add a conflicting package through a new dependency edge, we add this edge early in an non-finalized state (before the package is fully processed), finalizing the edges only after package processing is done.

language/tools/move-package/src/resolution/dependency_graph.rs
amnn: This also means you probably don't need to make these free functions -- you can put them in the `impl` block for `DependencyGraph` and always refer to `&self.package_graph`.

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -969,12 +1085,125 @@ fn pkg_deps_equal<'a>(
...
+/// - the new package has been inserted as a result of an override and:
+///     - it's dominating override for the existing package then new package is the winner
+///     or
+///     - it's not a dominating override for the existing package then an error must be reported
+/// or
+/// - the new package has not been inserted as a result of an override and an error must be reported
+///
+/// Correctly finding dominators only works because when trying to add a conflicting package through
+/// a new dependency edge, we add this edge early in an non-finalized state (before the package is
+/// fully processed), finalizing the edges only after package processing is done.
```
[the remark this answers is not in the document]
amnn: Additionally, now I've read the full scope of changes, I'm a little concerned that adding finalized edges may break a legitimate dominating override.  The example I have in mind goes like this:

```
     ┌───R───┐
     │   │   │
     │   │o  │
     ▼   ▼   ▼
 ┌───C   B◄──A
 │   │       │
o│   │       │
 ▼   ▼       │
 D◄──E◄──B'◄─┘
     │
     │
     ▼
     D'
```

This is a legitimate dependency graph where `R` override's `A`'s dependency on `B` (moving it from `B'` to `B`), and `C` overrides `E`'s dependency on `D` similarly.  But if you include non-finalized edges when computing dominators, you get into a situation where `C`'s override of `D` is not recognised because `C` does not dominate `E` (there's a separate path to `E` via the non-finalized `A -> B'` edge).

amnn (changes requested):
I think there may still be issues with the core algorithm -- check out the comment on `override_pkg`.

PR #970, opened 2023-03-08
[loader] Cache CompiledModule separately from Module

amnn:
Separates verification, which blesses `CompiledModule`s from loading which caches `Module`s:

- Renames the existing cache as `loaded_modules`.
- Moves `CompiledModule` out of `Module` and into its own cache -- `compiled_modules` -- populated during verification (which now no longer populates `loaded_modules`).
- `ModuleCache::insert` now recursively adds `Module`s, `Function`s, and `Struct`s (ensuring dependency modules are loaded first), assuming the all dependencies have been verified (failing otherwise).
- Public APIs into the loader that previously returned `Arc<Module>` now also return a `Arc<CompiledModule>`.

This change is motivated by the fact that in the presence of package upgrades, a single `CompiledModule` can end up being stamped out as multiple `Module`s

## Test Plan

```
move$ rm -rf ~/.move && cargo nextest
```

### DIFF
--- a/language/move-vm/runtime/src/loader.rs
+++ b/language/move-vm/runtime/src/loader.rs
@@ -684,26 +684,15 @@ impl Loader {
 
         match self.verify_script(&script) {
             Ok(_) => {
-                // load dependencies
-                let immediate_dependencies = script.immediate_dependencies();
-                for dep in &immediate_dependencies {
-                    self.load_module(dep, data_store)?;
-                }
+                // verify and load dependencies, fetching the verified compiled module.
+                let deps: VMResult<Vec<_>> = script
+                    .immediate_dependencies()
+                    .into_iter()
+                    .map(|dep| Ok(self.load_module(&dep, data_store)?.0))
+                    .collect();
 
-                // fetch compiled modules from cache
-                let dep_compiled_modules: Vec<_> = {
-                    let module_cache = self.module_cache.read();
-                    immediate_dependencies
-                        .into_iter()
-                        .map(|module_id| module_cache.compiled_module_at(&module_id).unwrap())
-                        .collect()
-                };
-
-                // verify dependencies
-                dependencies::verify_script(
-                    &script,
-                    dep_compiled_modules.iter().map(|module| module.as_ref()),
-                )?;
+                // verify script linkage
+                dependencies::verify_script(&script, deps?.iter().map(Arc::as_ref))?;
 
                 Ok(script)
             }
--- a/language/move-vm/runtime/src/loader.rs
+++ b/language/move-vm/runtime/src/loader.rs
@@ -121,20 +121,28 @@ impl ScriptCache {
     }
 }
 
-// A ModuleCache is the core structure in the Loader.
-// It holds all Modules, Types and Functions loaded.
-// Types and Functions are pushed globally to the ModuleCache.
-// All accesses to the ModuleCache are under lock (exclusive).
+/// The ModuleCache holds all verified modules as well as loaded modules, structs, and functions.
+/// Structs and functions are pushed into a global structure, handles in compiled modules are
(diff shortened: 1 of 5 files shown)


language/move-vm/runtime/src/loader.rs
```diff
@@ -990,35 +1020,63 @@ impl Loader {
...
-        }
+    ) -> VMResult<(Arc<CompiledModule>, Arc<Module>)> {
+        let compiled = {
+            let locked_cache = self.module_cache.read();
+            match (
+                locked_cache.compiled_module_at(id),
+                locked_cache.loaded_module_at(id),
+            ) {
+                (None, Some(_loaded)) => {
+                    unreachable!("Loaded module without verified compiled module");
```
dariorussi, line 1031: this is ok but something we may want to keep in mind. A binary module may exist without its compiled module being around (though a future to be defined). But given we have split the `CompiledModule` and `Module` in the API (both returned) we may want to consider that option. Otherwise it could make sense to return a `CompiledModule` from a `BinaryModule`/`Module`

dariorussi (commented):
we'll talk tomorrow though this seems fine and a needed change to move forward. I'll approve after our talk and another look tomorrow (if others approve sooner feel free to go in).
As a general comment, with likely no action item, I was hoping to move to a place where we first collect all modules (load from storage), then we verify all modules internally (verification with no dependencies), then we compile/link them (binary module). And have those 3 steps much more clear in the code, in the different caching layers (think module_cache), and operationally. Also with much less recursion built in.
I mention just in case it gives you or others some input for thoughts....

language/move-vm/runtime/src/loader.rs
```diff
@@ -121,20 +121,28 @@ impl ScriptCache {
...
+/// replaced with indices into these global structures in loaded modules.  All access to the
+/// ModuleCache via the Loader is under an RWLock.
 pub struct ModuleCache {
-    modules: BinaryCache<ModuleId, Module>,
+    /// Compiled modules go in this cache once they have been verified (individually and linkage
+    /// with dependencies and friends)
+    compiled_modules: BinaryCache<ModuleId, CompiledModule>,
+    /// Loaded modules go in this cache once their structs and functions have populated `structs`
+    /// and `functions` below.
+    loaded_modules: BinaryCache<ModuleId, Module>,
```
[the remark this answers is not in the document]
amnn: I like the change of `Module` -> `BinaryModule` (I wanted to do that myself, but worried that I was just putting my own fingerprint on the codebase for no reason!) but the distinction between `compiled` and `binary` is less clear to me than `compiled` and `loaded` -- let's chat about it.

language/move-vm/runtime/src/loader.rs
```diff
@@ -169,18 +183,29 @@ impl ModuleCache {
...
+        for dep in module.immediate_dependencies() {
+            let compiled_dep = self.compiled_module_at(&dep).ok_or_else(|| {
+                PartialVMError::new(StatusCode::MISSING_DEPENDENCY)
+                    .with_message(format!("Cannot load dependency {dep} of {id}"))
+                    .finish(Location::Undefined)
+            })?;
+
+            self.insert(natives, dep, compiled_dep.as_ref())?;
+        }
+
```
dariorussi, line 202: is all of this under lock now?

amnn: Yes, it is.  The same work was happening under the write lock before, but we would release and re-acquire the lock between each `insert` of a dependency module.

dariorussi (approved):
the idea is to move forward with this and build on the different caches model and possibly moving to a phased loading.
I mention briefly just for the broader audience.
We will talk a bit more extensively within the team about where this is going

Issue #966, opened 2023-03-07
Replace Or Supplement std::type_name with std::struct_tag


PaulFidika:
So currently if you want to do type introspection, you use the std::type_name::get() function to get a TypeName, then convert it into a string, and then parse that string to the type-name up into one of its component pieces (i.e., package-id or module-name). We've written a bunch of utility functions that do this parsing:

https://github.com/capsule-craft/capsules/blob/master/packages/sui_utils/sources/encode.move

However, what would be simpler and more useful would be to skip strings altogether, and simply return StructTags, like this:

struct StructTag has store, copy, drop {
        package_id: address,
        module_name: string::String,
        struct_name: string::String,
        generics: vector<string::String>
    }
    
Note that it might have been interesting to make make struct_tag.generics be of type vector<StructTag> instead, but this sort of recursive type definition is not allowed, so the generics will just have to be strings.

For Sui Move, it might be interesting to have package_id be a sui::object::ID instead of an address, but for the Move core IDs do not exist.

We've implemented StructTags here:

https://github.com/capsule-craft/capsules/blob/master/packages/sui_utils/sources/struct_tag.move

Currently these are created using string-parsing, but if StructTags were created natively within Move it would presumably be a lot cheaper to construct them than having to parse ascii strings. These StructTags can be used to compare object-types rather easily; i.e., are they from the same package, the same module, are they the same struct but with different generics, etc.

In addition to being faster and simpler to work with, than TypeNames StructTags are also smaller to store. Addresses are encoded as hex in ascii, meaning that a 32 byte address uses up 64 bytes of storage as a hex-ascii string. I honestly can't really think of any downsides.

sblackshear:
Quick thoughts: 
- We did think about something similar to this when building `type_name`, but wanted to bias toward reflection API's that encourage simple, easy-to-reason about usage (e.g., comparison of typenames). We also wanted a representation that would work for both structs and primitive types--felt a bit uncomfortable with patterns like introducing "dummy struct tags" for primitive types
- Clearly, the existence of `type_name::into_string` makes it possible to go deeper down the reflection rabbit hole, even if it's (somewhat intentionally) inconvenient
- I'm not necessarily opposed to something like an in-Move representation of `StructTag`, but I'd like to understand what sort of operations you'd like to do on a `StructTag` (if there are only a few) and see if we can implement those more directly via native functions on `TypeName` if so. If that's possible, I think it will make it much easier for programmers and static analysis tools to reason about our reflection APIs

[...]

PR #971, opened 2023-03-10
Struct tag

bytedeveloperr:
## Motivation

This adds struct decomposition into move natively. 

While working with move, The need to work with decomposed struct arise and we had to write our solution on top of `std::type_name` which is not efficient and involve unnessary pain while trying to get what we need.

This introduces a new native module called `struct_tag` and a struct called `StructTag` which represent the decomposed struct.

```move
    struct StructTag has copy, store, drop {
        address_: address,
        module_name: String,
        struct_name: String,
        generics: vector<String>
    }
```
For more information about this please see https://github.com/move-language/move/issues/966

Yes

## Test Plan
after switching current dir to `move/language/move-stdlib`, run

```sh
cargo run -p df-cli -- test
```

### DIFF
--- a/language/move-stdlib/sources/struct_tag.move
+++ b/language/move-stdlib/sources/struct_tag.move
@@ -4,10 +4,10 @@ module std::struct_tag {
 
     struct StructTag has copy, store, drop {
         /// Address of the entity that the struct belongs to.
-        /// taking `00000000000000000000000000000001::option::Option<u64>` for example, 
+        /// taking `00000000000000000000000000000001::option::Option<u64>` for example,
         /// the address will be `00000000000000000000000000000001`
         address_: address,
-        /// the name of the module where the struct is defined. 
+        /// the name of the module where the struct is defined.
         /// using the example struct above the module name should be `option`
         module_name: String,
         /// the name of the struct itself.
@@ -28,11 +28,11 @@ module std::struct_tag {
 
     /// Returns the module authority for the struct of type `T`
     public fun module_authority<T>(): StructTag {
-        let StructTag { 
-            address_, 
-            module_name, 
-            struct_name: _, 
-            generics: _ 
+        let StructTag {
+            address_,
+            module_name,
+            struct_name: _,
+            generics: _
         } = get<T>();
 
         StructTag {
--- a/language/move-stdlib/sources/struct_tag.move
+++ b/language/move-stdlib/sources/struct_tag.move
@@ -1,12 +1,12 @@
 /// Module to decompose a move struct into it's components.
 module std::struct_tag {
-    use std::ascii::String;
+    use std::ascii::{Self, String};
 
     struct StructTag has copy, store, drop {
-        /// Address of the package that the struct belongs to.
+        /// Address of the entity that the struct belongs to.
         /// taking `00000000000000000000000000000001::option::Option<u64>` for example, 
-        /// it's package address will be `00000000000000000000000000000001`
-        package_address: address,
+        /// the address will be `00000000000000000000000000000001`
+        address_: address,
         /// the name of the module where the struct is defined. 
         /// using the example struct above the module name should be `option`
         module_name: String,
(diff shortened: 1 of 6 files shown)


PaulFidika:
Note that for struct_tag.address_, we could also go with other names, like 'addr', 'package', or 'account'. Aptos and Sui both have different naming conventions (Aptos calls them account-addresses whereas Sui calls them package-ids), and we wanted to be inclusive of both.

Additionally for generics, we went with vector<String>, but we could go with vector<TypeName> if people like that better?

language/move-stdlib/sources/struct_tag.move
```diff
@@ -0,0 +1,55 @@
...
+    /// Returns the tag of the struct of type `T`
+    public native fun get<T>(): StructTag;
+
+    // Converts `self` into a tuple of it's inner values
+    public fun into(self: &StructTag): (address, String, String, vector<String>) {
+        (self.address_, self.module_name, self.struct_name, self.generics)
+    }
+
+    /// Returns the module authority for the struct of type `T`
+    public fun module_authority<T>(): StructTag {
```
jolestar, line 30: How to understand the module authority? I see the struct_name is hardcode `Witness`, so is it a programming pattern?

PaulFidika: Yeah exactly! My intention is that every module should be able to assert their ownership in this way. For example you could do something like:

```
public fun do_something<T: drop>(witness: T, object: &Object) {
   assert!(struct_tag::module_authority<Object> == struct_tag::get<T>, ENO_MODULE_AUTHORITY);
...
```

This is stating that, in order for this function to continue, we need a witness, and that Witness must be the module-authority witness produced by the same module that produced the Object.

This way we can guarantee that the function-call originated from the same module that defined Object in the first place.

Suficio:
We've been using this pattern to authorize getting a mutable reference to an object from a type-erased container which should only be allowed in the context of the module defining said object. This PR would be very helpful to achieve that.

Here is a link to our documentation which should put it in context:
https://origin-byte.github.io/nft.html#borrow_domain_mut

sblackshear:
Apologies for being slow in replying to this. I left some [comments](https://github.com/move-language/move/pull/971) that I'll reproduce here for convenience

> Quick thoughts:
> 
> * We did think about something similar to this when building `type_name`, but wanted to bias toward reflection API's that encourage simple, easy-to-reason about usage (e.g., comparison of typenames). We also wanted a representation that would work for both structs and primitive types--felt a bit uncomfortable with patterns like introducing "dummy struct tags" for primitive types
> * Clearly, the existence of `type_name::into_string` makes it possible to go deeper down the reflection rabbit hole, even if it's (somewhat intentionally) inconvenient
> * I'm not necessarily opposed to something like an in-Move representation of `StructTag`, but I'd like to understand what sort of operations you'd like to do on a `StructTag` (if there are only a few) and see if we can implement those more directly via native functions on `TypeName` if so. If that's possible, I think it will make it much easier for programmers and static analysis tools to reason about our reflection APIs

@PaulFidika replied to the third point with some specific use-cases that make a lot of sense to support, but I think we can do so with a reflection API that is slightly less powerful, or at least uses the big hammer of converting types into strings only when it's needed. Keeping the artifacts we reflect over as opaque as possible will make life easier for static reasoning tools like the Prover--when we go into string-land, things get a lot trickier for such tools.

The two use-cases Paul suggested (also in the thread above) are:
1. Getting the package name for a type
2. Deriving a unique, fixed-length identifier for a type (e.g., by hashing it and converting the result into an `address`)

What do you think about adding functions specifically for these use-cases (e.g., `get_package<T>(): Option<MoveIdentifier>` and `get_unique_id<T>(): u256`) rather than a more generic API that lets folks hand-roll these? Definitely open adding others if we are missing other critical use-cases, or going with the approach in this PR if there are so many use-cases that the general library is the way to go. But reflection is very hard for static reasoning, and amenability to static reasoning is a key part of what makes it easy to build awesome tolling for Move + keep it safe, so I'm hesitant to go too far down that road if we can avoid it.

PR #973, opened 2023-03-10
[loader] Remove features not relevant for Sui

amnn:
Remove support for

- republishing an existing module or doing a compatibility check in that case in `session` and `loader`.  
- invalidating the loader because a module has been republished.
- checking for cycles among friends.
 
These codepaths are never used by Sui, which does not permit republishing at the same address except in very specific edge cases that are handled externally, and continuing to support these codepaths will further complicate support for package upgrades (i.e. republishing) on Sui.

This PR also removes any test that relies on or tests republishing, as well as support in the sandbox for republishing and/or compatibility checks

## Test Plan

```
move$ cargo nextest
move$ cargo test --workspace
```


dariorussi (approved):
let's think about how to reconcile this in main at some point.
The issue of upgrade and versioning is very specific to the different Move implementations and we need a way to be able to have whatever a vendor wants, not as it is today which is a bit of everything.
This is good in making it clear what we (mysten) need and in time we need to clean up the main repo/story

rvantonder (approved):
In favor of removing things that are not needed 🙂

PR #974, opened 2023-03-10
[loader] Do not traverse friends when verifying before load

amnn:
For Sui, packages are immutable and it is not practically possible to introduce a friendship across packages: the dependent package would need to be published first, and would need to know the address containing the friend at that time, but this is not possible because a package's address is decided on publish.

We can take advantage of these properties to simplify loading by assuming that a `friend ...::B` declaration in module `A` is only relevant if `B` depends on `A`, and will be factored into verification when verifying/loading `B` (in which case the visibility of `A`'s functions to `B` is relevant).

Furthermore, it should not matter if `A`'s friend module does not exist, and no part of `A`'s friends are referenced when `A` is loaded, so they do not need to be known to the loader.

## Test Plan

```
cargo nextest
```


dariorussi (approved):
this is a really good chunk of code disappearing... the less code in the loader the better we all are

PR #975, opened 2023-03-11
[bytecode verifier] Relax metering during tests

wrwg:
This aligns the default of the metering config to be unbounded, like with other constraints. Addresses https://github.com/aptos-labs/aptos-core/issues/7040

### DIFF
--- a/language/move-bytecode-verifier/src/verifier.rs
+++ b/language/move-bytecode-verifier/src/verifier.rs
@@ -184,10 +184,11 @@ impl Default for VerifierConfig {
             max_back_edges_per_function: None,
             max_back_edges_per_module: None,
             max_basic_blocks_in_script: None,
-            /// General metering for the verifier. This defaults to a bound which should align
-            /// with production, so all existing test cases apply it.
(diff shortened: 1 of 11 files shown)


gregnazario (approved):

language/move-bytecode-verifier/src/verifier.rs
```diff
@@ -201,4 +202,30 @@ impl VerifierConfig {
...
+            max_basic_blocks: Some(1024),
+            max_basic_blocks_in_script: Some(1024),
+            max_value_stack_size: 1024,
+            max_type_nodes: Some(256),
+            max_push_size: Some(10000),
+            max_dependency_depth: Some(100),
+            max_struct_definitions: Some(200),
+            max_fields_in_struct: Some(30),
+            max_function_definitions: Some(1000),
+
```
gregnazario, line 221: This won't conflict with anything that already exists right?

wrwg: It replaces some config used for testing locally to become available globally.

PR #977, opened 2023-03-11
[loader] Remove invalidation feature

amnn:
(Sui-specific) The way packages are republished/upgraded does not require loader invalidation, so this part of the loader can also now be removed.

## Test Plan

```
cargo nextest
```


rvantonder (approved):

dariorussi (approved):
nice to see this go, thanks

PR #983, opened 2023-03-14
port verifier metering

oxade:
## Motivation

Ports over verifier metering from https://github.com/move-language/move/pull/950
Next step is to test the logic and pare back what we don't need

Yes
## Test Plan

Existing tests


oxade:
> is this effectively off right now? I wonder if we want to commit this and if so it has to be off. and is this ready to go in?

Yes it will be off when we pick it up in Sui. And it is ready to go in. We will iterate on it as we go.

dariorussi (approved):
I suppose good luck to us all? ;)

PR #985, opened 2023-03-15
[interpreter] Fix native as entry point

tnowacki:
## Motivation

- interpreter entrypoint did not handle the case where the function is native

## Test Plan

- by value test, by ref test, type test


language/move-vm/runtime/src/interpreter.rs
```diff
@@ -101,15 +102,45 @@ impl Interpreter {
...
+                    Some(id) => e
+                        .at_code_offset(function.index(), 0)
+                        .finish(Location::Module(id.clone())),
+                    None => PartialVMError::new(StatusCode::UNKNOWN_INVARIANT_VIOLATION_ERROR)
+                        .with_message(
+                            "Unexpected native function not located in a module".to_owned(),
+                        )
+                        .finish(Location::Undefined),
+                })?;
+            Ok(return_values.into_iter().collect())
```
amnn, line 138: Is there some equivalent to the paranoid checks we do in the other code path that should go here?

amnn (approved):
Seems reasonable!

language/move-vm/runtime/src/interpreter.rs
```diff
@@ -101,15 +102,45 @@ impl Interpreter {
...
+        if function.is_native() {
+            for arg in args {
+                interpreter
+                    .operand_stack
+                    .push(arg)
+                    .map_err(|e| e.finish(Location::Undefined))?;
+            }
+            let resolver = function.get_resolver(loader);
+            let return_values = interpreter
+                .call_native_return_values(
```
dariorussi, line 120: just out of curiosity... I have a PR where I moved collecting the return values from the stack in this function.
If we had that we could have called directly the `call_native` as it was and fetch the value from the stack... or not?

oxade (approved):
good find

dariorussi (approved):
it does feel like we should review native invocation some day....

language/move-vm/runtime/src/interpreter.rs
tnowacki: Ah, yeah we could retrieve them from the stack. Do you want me to attempt to do that in this PR?

PR #989, opened 2023-03-17
[RFC][DRAFT][VM] Function pointer support in VM

runtian-zhou:
## Motivation

This is the very first initial draft of how MoveVM can potentially support function pointer. We believe function pointer could be a very important corner stone of how interface should be implemented in the future.

In this implemetnation, we add three new opcodes and one more SignatureToken type:
- GetFunctionPointer: Convert a `FunctionHandle` into a function pointer value on top of the stack
- GetFunctionPointerGeneric:  Convert a `FunctionInstantiation` into a function pointer value on top of the stack
- CallFunctionPointer: Instruction to invoke the function pointer

Right now it has the following limitations:
- No currying is allowed.
- Only function with no acquire list can be obtained as function pointer to avoid reentrency issue and major refactor in our borrow semantics.

Yes.

## Test Plan

This implementation only passed compilation and is no way close to merging into main. The intention of this PR is just to assess the safety of this change and to collect usability feedbacks. I'll try to see how I can mod the IR to try the new feature out in the compiler


language/move-vm/transactional-tests/tests/function_pointers/example/coin_interface/coin_interface.mvir
```diff
@@ -0,0 +1,125 @@
...
+    let interface: CoinInterface.T<BasicCoin1.Coin<CoinType.Foo>>;
+    let coin1: BasicCoin1.Coin<CoinType.Foo>;
+    let coin2: BasicCoin1.Coin<CoinType.Foo>;
+    let v: u64;
+label b0:
+    coin1 = BasicCoin1.mint<CoinType.Foo>(10);
+    coin2 = BasicCoin1.mint<CoinType.Foo>(20);
+    interface = BasicCoin1.coin_interface<CoinType.Foo>();
+
+    v = GenericAdder.add_coins<BasicCoin1.Coin<CoinType.Foo>>(&interface, &coin1, &coin2);
```
runtian-zhou, line 121: This is an example for how to implement interface with function pointer provided.

language/move-vm/types/src/loaded_data/runtime_types.rs
```diff
@@ -50,6 +50,7 @@ pub enum Type {
     U16,
     U32,
     U256,
+    Function,
```
vgao1996, line 53: Hmm... no type parameters?

runtian-zhou: So I intentionally hided them because the runtime already carried the type information so there's no need to duplicate it.

PR #990, opened 2023-03-17
Add `load_function_without_types` to VM session and loader

banool:
## Motivation

For Aptos, in order to expose whether a function is a view function in the API-based ABI today, we need to be able to load function metadata through the Aptos VM. The existing function exposed by the Session for this, `load_function`, loads information about arguments, generic types, and return types, which is more than is necessary for this use case. 

To that end I've exposed a version of this function that loads only the module and function, without this additional type information.

Yes.

## Test Plan
Not sure. As far as I can tell, `load_function` does not have unit tests either.


banool:
I'll re-open this in aptos-core soon once we have the sync working.

banool:
See https://github.com/aptos-labs/aptos-core/pull/7275.

PR #992, opened 2023-03-19
[move-prover] loop unrolling

meng-xu-cs:
Prover now supports bounded loop unrolling with
```move
invariant [unroll = <iteration-limit>] true;
```

The semantics of loop unrolling aligns with most of the bounded model
checking work, which can be illustrated below:

```rust
loop {
    invariant [unroll = N] true;
    assert C1;
}
assert C2;
```

- If C1 holds for iteration 0, 1, ..., N, the prover will not complain
  about C1, even when C1 fails in the K-th iteration (K > N).
- If C2 holds when loop exits within 0, 1, ... N iterations, the prover
  will not complain about C2, even when C2 fails when loop exits after K
  iterations (K > N).

This is evident in the `loop_unroll.move` test case.

## Motivation

(Write your answer here.)

## Test Plan


wrwg:
Do you mind filing this in third_party/move? You can also do here but this
is ready now

meng-xu-cs:
Will move it to third_party/move.

meng-xu-cs:
Run into some issues with `third_party/move` (mostly due to not being on a Mac). Plus, as I am not sure how test cases are run there, I'll develop here for a bit longer and once I get tests passing I'll patch up `third_party/move` and submit PRs from there.

wrwg:
> Run into some issues with `third_party/move` (mostly due to not being on a Mac). Plus, as I am not sure how test cases are run there, I'll develop here for a bit longer and once I get tests passing I'll patch up `third_party/move` and submit PRs from there.

Not sure what issues that could be. You do not need to run copybara, only 'admins' do this. Relevant tests in third_party/move are all run as part of standard CI. But no issue to develop here and then sync into aptos-core.

rahxephon89 (approved):
LGTM, I am wondering whether we need a warning message if loop unrolling is used.

language/move-model/src/pragmas.rs
```diff
@@ -287,6 +287,10 @@ pub const CONDITION_CHECK_ABORT_CODES_PROP: &str = "check";
 /// enabled disabled by the disable_invariant_in_body pragma
 pub const CONDITION_SUSPENDABLE_PROP: &str = "suspendable";
 
+/// A property that can be attached to a loop invariant to indicate that the loop needs to
+/// be unrolled to a certain depth, a typical relaxation in bounded model checking.
```
wrwg, line 291: This allows controlling the unrolling on the finest level of granularity, by loop. But can we also have a pragma (perhaps in a 2nd PR) which can be attached to function or module level, and applies to all loops? This avoids having to look at every single loop.

language/move-prover/tests/sources/functional/loop_unroll.move
```diff
@@ -0,0 +1,105 @@
...
+        ensures false;
+    }
+
+    fun t2_success(): u64 {
+        let i = 0;
+        while ({
+            // it is okay to unroll more than the loop iterations, it won't
+            // affect the correctness
+            spec {
+                invariant [unroll = 6] true;
```
wrwg, line 25: Together with the predicate `true`, this doesn't read very intuitive, but I guess we can improve this to the point where no spec blocks need to be attached to get the unrolling.

wrwg (approved):

junkil-park (approved):

junkil-park:
@meng-xu-cs , this PR seems to have conflicts.

PR #1000, opened 2023-03-22
[move-binary-format] Add additional information to the normalized module

tzakian:
This adds additional information to the normalized module. It adds:
* Dependencies
* Private functions
* Constants


dariorussi (approved):
it seems pretty straightforward

language/move-binary-format/src/compatibility.rs
```diff
@@ -268,3 +268,89 @@ fn type_parameter_phantom_decl_compatibile(
...
+impl InclusionCheck {
+    pub fn check(&self, old_module: &Module, new_module: &Module) -> PartialVMResult<()> {
+        let err = Err(PartialVMError::new(
+            StatusCode::BACKWARD_INCOMPATIBLE_MODULE_UPDATE,
+        ));
+
+        // Module checks
+        if old_module.address != new_module.address
+            || old_module.name != new_module.name
+            || old_module.file_format_version != new_module.file_format_version
```
amnn, line 289: file format is an interesting one... is that strictly necessary? (If this is a requirement for the `Compatibility` checks then feel free to ignore, but I could imagine an additive update that only uses new aspects of the file format in newly added functions, that could be considered a superset of the old package).

language/move-binary-format/src/compatibility.rs
```diff
@@ -268,3 +268,89 @@ fn type_parameter_phantom_decl_compatibile(
...
+/// A simpler, and stricter compatibility checker relating to the inclusion of the old module in
+/// the new.
+#[derive(Clone, Debug, Ord, PartialOrd, Eq, PartialEq)]
+pub enum InclusionCheck {
+    Subset,
+    Equal,
+}
+
+impl InclusionCheck {
+    pub fn check(&self, old_module: &Module, new_module: &Module) -> PartialVMResult<()> {
```
amnn, line 281: Nit: I assume that we're checking the `old_module` is a subset or equal to `new_module`, but we should document that explicitly.

language/move-binary-format/src/compatibility.rs
```diff
@@ -268,3 +268,89 @@ fn type_parameter_phantom_decl_compatibile(
...
+                .or_else(|| new_module.private_functions.get(name))
+            {
+                Some(new_func) if old_func == new_func => (),
+                _ => {
+                    return err;
+                }
+            }
+        }
+
+        // Friend checks
```
amnn, line 334: Structuraly speaking, I see why we have this, but are friendships strictly relevant for this check? (i.e. if friendships change but nothing else changed, has the package meaningfully changed, assuming it passes verification?)

language/move-binary-format/src/compatibility.rs
```diff
@@ -268,3 +268,89 @@ fn type_parameter_phantom_decl_compatibile(
...
+        if !old_module
+            .friends
+            .iter()
+            .all(|old_friend| new_friends.contains(old_friend))
+        {
+            return err;
+        }
+
+        // Constant checks
+        let new_constants: BTreeSet<_> = new_module.constants.iter().collect();
```
amnn, line 345: By a similar argument to the above for friendships, I'm not strictly sure we need to check the constant pool.  (If a constant gets added to the pool but nothing uses it, did the module change?)

language/move-binary-format/src/normalized.rs
```diff
@@ -420,3 +706,46 @@ impl std::fmt::Display for Type {
         }
     }
 }
+
+fn struct_name(m: &CompiledModule, idx: &StructDefinitionIndex) -> Identifier {
```
amnn, line 710: This is quite a common sequence of operations, I feel like we should add this to `ModuleAccess` or `BinaryIndexedView` or something like that (if it's not already there).

language/move-binary-format/src/unit_tests/compatibility_tests.rs
```diff
@@ -115,3 +204,30 @@ fn deprecated_add_script_visibility() {
         .check(&friend_module, &script_module)
         .is_err());
 }
+
+#[test]
+fn check_exact_and_unchange_same_module() {
```
amnn, line 209: It would also be great to include some tests that show that the check is permutation invariant (for all the pools).

amnn (approved):
Left a bunch of questions about relaxing various things and adding tests, but the core thesis of the PR looks good, thanks @tzakian !

language/move-binary-format/src/compatibility.rs
tzakian: I was going back and forth on this and whether we should allow `old < new` or not care. I think relaxing this to `old < new` makes sense, and in the case of an `Equal` match we will enforce that the file format is the same.

language/move-binary-format/src/compatibility.rs
tzakian: Structurally I felt better having these checks, but technically they don't need to be checked. I'll remove.

language/move-binary-format/src/compatibility.rs
tzakian: For the subset case agreed. But for the `Equal` check, I think _technically_ it's fine to not check this, but it feels kinda strange to me to not check it. I think I'll also remove these along with the friendship checks to relax things a bit, and we can tighten them down later if need be.

PR #1004, opened 2023-03-23
[loader] "defining module" support

amnn:
Add a new API to `LinkageResolver` to translate the fully-qualified name of a struct (including its runtime ID) to the defining ID (the ID of the package in storage that first defined the struct).

Use this API to store the defining ID as well as the runtime ID in the loaded `StructType`. (Note: This results in a situation where you can load a type from a type tag, get the type tag for that type and have that operation not be the identity, because it has translated the struct's runtime package ID to its defining package ID).

## Test Plan

New loader unit tests:

```
$ cargo nextest -- loader_tests::relink_defining_module
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3704,6 +3704,7 @@ name = "move-vm-integration-tests"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "fail",
  "memory-stats",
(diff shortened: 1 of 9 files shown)


tzakian (approved):
LGTM! Like the tests :)

PR #1006, opened 2023-03-24
[lock] BuildConfig::install_dir configures temporary lock file directory

amnn:
Previously, temporary lock files were always stored in a sub-directory of the package root because `LockFile` assumed that the `build` directory was always there, while the lock file itself could be committed elsewhere.

But the location of `/build/` is controlled by
`BuildConfig::install_dir` so we should heed that as well when finding where to store the temporary lock file.

## Test Plan

Existing tests:

```
cargo nextest
```


tzakian (approved):
Makes sense :)

PR #1009, opened 2023-03-27
[session] Accept MoveResolver by value

amnn:
Rather than accepting `MoveResolver` by reference accept it by value, but provide generic implementations for references of types that implement the traits.  Also expose a reference to it on `Session` and return it by value on `finish`.

This allows for the creation of resolvers that hold state that is scoped to the session, internally to themselves, rather than relying on that state's lifetime being maintained externally.

## Test Plan

```
$ cargo nextest
```

### DIFF
--- a/language/move-vm/runtime/src/data_cache.rs
+++ b/language/move-vm/runtime/src/data_cache.rs
@@ -72,9 +72,12 @@ impl<'l, S: MoveResolver> TransactionDataCache<'l, S> {
     /// published modules.
     ///
     /// Gives all proper guarantees on lifetime of global data as well.
-    pub(crate) fn into_effects(self) -> PartialVMResult<(ChangeSet, Vec<Event>, S)> {
+    pub(crate) fn into_effects(mut self) -> (PartialVMResult<(ChangeSet, Vec<Event>)>, S) {
+        (self.impl_into_effects(), self.remote)
+    }
+    fn impl_into_effects(&mut self) -> PartialVMResult<(ChangeSet, Vec<Event>)> {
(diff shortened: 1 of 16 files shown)


awelc (approved):
LGTM! I guess now having `0` tagged on the results in quite a few places is unavoidable (at least I can't think of a better way...)

language/move-vm/runtime/src/data_cache.rs
```diff
@@ -72,9 +72,12 @@ impl<'r, 'l, S: MoveResolver> TransactionDataCache<'r, 'l, S> {
...
     ///
     /// Gives all proper guarantees on lifetime of global data as well.
-    pub(crate) fn into_effects(self) -> PartialVMResult<(ChangeSet, Vec<Event>)> {
+    pub(crate) fn into_effects(mut self) -> (PartialVMResult<(ChangeSet, Vec<Event>)>, S) {
+        (self.impl_into_effects(), self.remote)
+    }
+    fn impl_into_effects(&mut self) -> PartialVMResult<(ChangeSet, Vec<Event>)> {
         let mut change_set = ChangeSet::new();
-        for (addr, account_data_cache) in self.account_map.into_iter() {
+        for (addr, account_data_cache) in std::mem::take(&mut self.account_map).into_iter() {
```
tzakian, line 80: Not sure if I understand the need for splitting this into two functions. It seems like taking the data cache by value and then destructuring it into its parts would still work fine here and remove the need for the `std::mem::take`s below.

tzakian (approved):
Overall changes look good to me as well! Only concern is around `into_effects` and if those changes can be simplified.

PR #1011, opened 2023-03-27
[tools] Add `try_exists` that surfaces other files issues if the file exists but has an error

tzakian:
`std::path::exists` collapses all errors to `false` however if we want to be able to return more fine-grained issues back we need to surface these errors as well.

This adds a `try_exist` function that only returns false if the file actually doesn't exist. Otherwise if there's a different error this will return an `Err` result.


amnn (approved):
LGTM!

PR #1012, opened 2023-03-27
[move-binary-format] Support changing of signatures for private entry functions

tzakian:
Adds a new flag `check_private_entry_linking` that when true checks the private entry function signatures cannot change. E.g.,

If we had

```
entry fun fn() { }
```

and

```
entry fun fn(_x: u64) { }
```

These would be compatible if `check_private_entry_linking` was set to `false` but ruled incompatible if `check_private_entry_linking` was set to `true`.

### DIFF
--- a/language/move-binary-format/src/compatibility.rs
+++ b/language/move-binary-format/src/compatibility.rs
@@ -14,19 +14,23 @@ use move_core_types::vm_status::StatusCode;
 
 /// The result of a linking and layout compatibility check. Here is what the different combinations. NOTE that if `check_struct_layout` is false, type safety over a series of upgrades cannot be guaranteed.
 /// mean:
-/// `{ check_struct_and_pub_function_linking: true, check_struct_layout: true, check_friend_linking: true }`: fully backward compatible
-/// `{ check_struct_and_pub_function_linking: true, check_struct_layout: true, check_friend_linking: false }`: Backward compatible, exclude the friend module declare and friend functions
-/// `{ check_struct_and_pub_function_linking: false, check_struct_layout: true, check_friend_linking: false }`: Dependent modules that reference functions or types in this module may not link. However, fixing, recompiling, and redeploying all dependent modules will work--no data migration needed.
-/// `{ check_struct_and_pub_function_linking: true, check_struct_layout: false, check_friend_linking: true }`: Attempting to read structs published by this module will now fail at runtime. However, dependent modules will continue to link. Requires data migration, but no changes to dependent modules.
-/// `{ check_struct_and_pub_function_linking: false, check_struct_layout: false, check_friend_linking: false }`: Everything is broken. Need both a data migration and changes to dependent modules.
+/// `{ check_struct_and_pub_function_linking: true, check_struct_layout: true, check_friend_linking: true, check_private_entry_linking: true }`: fully backward compatible
+/// `{ check_struct_and_pub_function_linking: true, check_struct_layout: true, check_friend_linking: true, check_private_entry_linking: false }`: Backwards compatible, private entry function signatures can change
+/// `{ check_struct_and_pub_function_linking: true, check_struct_layout: true, check_friend_linking: false, check_private_entry_linking: true }`: Backward compatible, exclude the friend module declare and friend functions
+/// `{ check_struct_and_pub_function_linking: true, check_struct_layout: true, check_friend_linking: false, check_private_entry_linking: false }`: Backward compatible, exclude the friend module declarations, friend functions, and private and friend entry function
+/// `{ check_struct_and_pub_function_linking: false, check_struct_layout: true, check_friend_linking: false, check_private_entry_linking: _ }`: Dependent modules that reference functions or types in this module may not link. However, fixing, recompiling, and redeploying all dependent modules will work--no data migration needed.
+/// `{ check_struct_and_pub_function_linking: true, check_struct_layout: false, check_friend_linking: true, check_private_entry_linking: _ }`: Attempting to read structs published by this module will now fail at runtime. However, dependent modules will continue to link. Requires data migration, but no changes to dependent modules.
+/// `{ check_struct_and_pub_function_linking: false, check_struct_layout: false, check_friend_linking: false, check_private_entry_linking: _ }`: Everything is broken. Need both a data migration and changes to dependent modules.
 #[derive(PartialEq, Eq, Debug, Clone, Copy)]
 pub struct Compatibility {
     /// if false, do not ensure the dependent modules that reference public functions or structs in this module can link
-    check_struct_and_pub_function_linking: bool,
+    pub check_struct_and_pub_function_linking: bool,
     /// if false, do not ensure the struct layout capability
-    check_struct_layout: bool,
-    /// if false, treat `friend` as `private` when `check_struct_and_function_linking`.
-    check_friend_linking: bool,
+    pub check_struct_layout: bool,
+    /// if false, treat `friend` as `private` when `check_struct_and_pub_function_linking`.
+    pub check_friend_linking: bool,
+    /// if false, treat `entry` as `private` when `check_struct_and_pub_function_linking`.
+    pub check_private_entry_linking: bool,
 }
 
 impl Default for Compatibility {
@@ -35,6 +39,7 @@ impl Default for Compatibility {
             check_struct_and_pub_function_linking: true,
             check_struct_layout: true,
             check_friend_linking: true,
+            check_private_entry_linking: true,
         }
     }
 }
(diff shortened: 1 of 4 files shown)


language/move-binary-format/src/compatibility.rs
```diff
@@ -14,19 +14,23 @@ use move_core_types::vm_status::StatusCode;
...
+/// `{ check_struct_and_pub_function_linking: false, check_struct_layout: true, check_friend_linking: false, check_private_entry_linking: _ }`: Dependent modules that reference functions or types in this module may not link. However, fixing, recompiling, and redeploying all dependent modules will work--no data migration needed.
+/// `{ check_struct_and_pub_function_linking: true, check_struct_layout: false, check_friend_linking: true, check_private_entry_linking: _ }`: Attempting to read structs published by this module will now fail at runtime. However, dependent modules will continue to link. Requires data migration, but no changes to dependent modules.
+/// `{ check_struct_and_pub_function_linking: false, check_struct_layout: false, check_friend_linking: false, check_private_entry_linking: _ }`: Everything is broken. Need both a data migration and changes to dependent modules.
 #[derive(PartialEq, Eq, Debug, Clone, Copy)]
 pub struct Compatibility {
     /// if false, do not ensure the dependent modules that reference public functions or structs in this module can link
```
tnowacki, line 26: Should we just make these fields public and kill off `new`? Would make it a bit more readable.

language/move-binary-format/src/unit_tests/compatibility_tests.rs
```diff
@@ -464,6 +469,100 @@ fn deprecated_add_script_visibility() {
...
+    let module = mk_module_entry(Visibility::Public as u8, true);
+    let mut updated_module = module.clone();
+    // Update the signature of the entry fun to now take a u64 argument.
+    updated_module
+        .exposed_functions
+        .get_mut(ident_str!("fn"))
+        .unwrap()
+        .parameters = vec![Type::U64];
+
+    assert!(Compatibility::new(true, true, true, false)
```
tnowacki, line 525: nit, bit hard for to reason about these tests with just these raw bools here.

language/move-binary-format/src/compatibility.rs
tzakian: Should be all fixed now :)

language/move-binary-format/src/unit_tests/compatibility_tests.rs
```diff
@@ -464,6 +469,145 @@ fn deprecated_add_script_visibility() {
...
+    assert!(Compatibility {
+        check_struct_and_pub_function_linking: true,
+        check_struct_layout: true,
+        check_friend_linking: true,
+        check_private_entry_linking: true
+    }
+    .check(&module, &updated_module)
+    .is_err());
+}
+
```
amnn, line 610: Let's also add tests for removing an entry function outright, and removing an `entry` modifier on a function.

amnn (commented):
I think there's one more part that needs to pay attention to the new flag, but otherwise looks good!

language/move-binary-format/src/compatibility.rs
```diff
@@ -49,36 +54,27 @@ impl Compatibility {
...
-            check_friend_linking,
+            check_private_entry_linking: false,
         }
     }
 
     pub fn need_check_compat(&self) -> bool {
         self.check_struct_and_pub_function_linking
             || self.check_friend_linking
             || self.check_struct_layout
+            || self.check_private_entry_linking
```
amnn, line 65: nit: To avoid some boilerplate:

```suggestion
        self != &Self::no_check()
```

language/tools/move-cli/src/sandbox/utils/mod.rs
```diff
@@ -345,16 +345,26 @@ pub(crate) fn explain_publish_error(
             let old_api = normalized::Module::new(&old_module);
             let new_api = normalized::Module::new(module);
 
-            if Compatibility::new(false, true, false)
-                .check(&old_api, &new_api)
-                .is_err()
+            if (Compatibility {
```
amnn, line 348: nit: unnecessary parens?

language/tools/move-cli/src/sandbox/utils/mod.rs
```diff
@@ -345,16 +345,26 @@ pub(crate) fn explain_publish_error(
...
+            .check(&old_api, &new_api)
+            .is_err()
             {
                 // TODO: we could choose to make this more precise by walking the global state and looking for published
                 // structs of this type. but probably a bad idea
                 println!("Layout API for structs of module {} has changed. Need to do a data migration of published structs", module_id)
-            } else if Compatibility::new(true, false, false)
-                .check(&old_api, &new_api)
-                .is_err()
+            } else if (Compatibility {
```
amnn, line 360: ditto

language/move-binary-format/src/compatibility.rs
```diff
@@ -128,14 +124,18 @@ impl Compatibility {
...
-                        struct_and_pub_function_linking = false;
+                    } else if old_func.visibility != Visibility::Private {
+                        struct_and_function_linking = false;
+                    } else if old_func.is_entry && self.check_private_entry_linking {
+                        // This must be a private entry function. So set the link breakage if we're
+                        // checking for that.
+                        private_entry_linking = false;
                     }
                     continue;
                 }
```
amnn, line 141: nit, `let ... else`?  Also, some consistency nits:

 - use `matches!` to check equivalence against `Visibility::Friend` but `==` on `Visibility::Private` -- might as well use equality on both.
 - checking for `self.check_private_entry_linking` -- it seems like the pattern for all the other checks is to do this once, at the end, so we might as well conform to that here?

```suggestion
            let Some(new_func) = new_module.functions.get(name) else {
               /* body of None match arm */
            }
```

Or even perhaps: 

```suggestion
            let Some(new_func) = new_module.functions.get(name) 
            else if old_func.visibility == Visibility::Friend {
                friend_linking = false;
                continue
            } else if old_func.visibility != Visibility::Private {
                struct_and_function_linking = false;
                continue
            } else if old_func.is_entry {
                private_entry_linking = false;
                continue
            }
```

Not sure if Rust allows that syntax...

language/move-binary-format/src/compatibility.rs
```diff
@@ -156,11 +156,19 @@ impl Compatibility {
             {
                 // if it was public(script), it must remain public(script)
                 // if it was not public(script), it _cannot_ become public(script)
-                old_func.is_entry == new_func.is_entry
+                // if it was private before, it can become public(script)
+                old_func.visibility == Visibility::Private || old_func.is_entry == new_func.is_entry
```
amnn, line 160: Not sure if we can/should change this behaviour as it's already baked into bytecode format 5, cc @tnowacki for thoughts.

language/move-binary-format/src/compatibility.rs
```diff
@@ -156,11 +156,19 @@ impl Compatibility {
...
             } else {
-                // If it was an entry function, it must remain one.
                 // If it was not an entry function, it is allowed to become one.
-                !old_func.is_entry || new_func.is_entry
+                // If it was a private entry function and we have `check_private_entry_linking` then it must remain one
+                // If it was a public entry function, it must remain one.
```
amnn, line 164: I would decouple `public` and  `entry` here -- i.e. we should not have a different rule for `public entry` to just `entry`, it should just be that if we have `check_private_entry_linking` enabled then a function cannot lose its `entry` modifier.

amnn: So the rule should be something like below, although it might be better off in another place:
```
if old_func.is_entry && !new_func.is_entry {
    private_entry_linking = false
}
```

(Possible you want to rename it to just `entry_linking` given that it is truly orthogonal to `public` vs `(private)`).

amnn (commented):
Thanks @tzakian -- this is almost there, but currently there is too much coupling between `public` and private and `entry` and non-entry -- the two should be orthogonal (also, I don't think we should change how compatibility works in v5 or below).

language/tools/move-cli/src/sandbox/utils/mod.rs
tzakian: Needed sadly for the Rust parser to be happy :'(

language/move-binary-format/src/compatibility.rs
tzakian: Sadly Rust syntax doesn't allow this :'( But moving the if-else into the `else { .. }` of the let-else looks OK (or, at least better than it does now).

language/move-binary-format/src/compatibility.rs
tzakian: Much prefer this. Thank you for the suggestion!

language/move-binary-format/src/compatibility.rs
amnn: Discussed this offline -- this change is actually necessary to maintain backwards compatibility because of the change from iterating over only `exposed` functions to iterating over all functions.

language/move-binary-format/src/compatibility.rs
```diff
@@ -156,10 +147,15 @@ impl Compatibility {
...
+                // if it was private before, it can become public(script)
+                old_func.visibility == Visibility::Private || old_func.is_entry == new_func.is_entry
             } else {
-                // If it was an entry function, it must remain one.
                 // If it was not an entry function, it is allowed to become one.
+                // If it was a entry function, it must remain one if unless we set
+                // `check_private_entry_linking` to false.
+                if old_func.is_entry && !new_func.is_entry {
+                    entry_linking = false;
+                }
```
amnn, line 158: This check is literally the opposite of the boolean condition below, right?  That makes me think something duplicative/redundant is going on here.

amnn (commented):
Almost there!

language/move-binary-format/src/compatibility.rs
tzakian: Good spot. What we actually want to do here is say `entry_linking &&= is_entry_compatible` I believe.

language/move-binary-format/src/compatibility.rs
```diff
@@ -128,53 +118,56 @@ impl Compatibility {
...
-            let is_entry_compatible = if old_module.file_format_version < VERSION_5
+
+            // Check visibility compatibility
+            match (old_func.visibility, new_func.visibility) {
+                (Visibility::Public, Visibility::Private | Visibility::Friend) => {
+                    struct_and_function_linking = false
+                }
+                (Visibility::Friend, Visibility::Private) => friend_linking = false,
+                _ => (),
+            }
```
amnn, line 142: nice!

language/move-binary-format/src/unit_tests/compatibility_tests.rs
```diff
@@ -464,24 +474,295 @@ fn deprecated_add_script_visibility() {
         .is_err());
 }
 
+#[test]
+fn private_entry_to_public_entry_allowed() {
+    let private_module = max_version(mk_module_entry(Visibility::Private as u8, true));
```
amnn, line 479: nit: It seems like you could do the cast to `u8` within the body of `mk_module_entry`?

language/move-binary-format/src/unit_tests/compatibility_tests.rs
```diff
@@ -464,24 +474,295 @@ fn deprecated_add_script_visibility() {
...
+            check_struct_and_pub_function_linking: true,
+            check_struct_layout: true,
+            check_friend_linking: true,
+            check_private_entry_linking: false,
+        }
+        .check(prev, new)
+        .is_ok());
+    }
+
+    // Every
```
amnn, line 622: nit: comment trailled off

language/move-binary-format/src/unit_tests/compatibility_tests.rs
```diff
@@ -464,24 +474,295 @@ fn deprecated_add_script_visibility() {
...
+    for (prev, new) in invalid_private_entry_breakages.into_iter() {
+        assert!(Compatibility::full_check().check(prev, new).is_err());
```
amnn, line 624: Is it worth adding `full_check` invocations for the valid combos as well?

amnn (approved):
I think we're there. Wooo, thanks @tzakian !  There's one case in the tests that I think we need to add, but otherwise, we're good to go.

PR #1017, opened 2023-03-28
[transactional tests] Support multi module publishing

tnowacki:
## Motivation

- This probably should have been done a while ago...
- Will be used for PT testing of publish and upgrade

## Test Plan

- One new tests, existing tests are not broken


rvantonder (approved):
sounds-reasonable

PR #1021, opened 2023-03-29
[model-compiler] Enabling basic type checking of Move functions

wrwg:
This is a very first step to enable the move model to process Move functions from source (and not just bytecode), and thus act as a compiler.

In this initial approach, the old Move compiler is run together with the new one. This allows us to continue to have a `CompiledModule` in the `GlobalEnv`. Untangling the model from compiled module is a larger refactoring which we can factor out in different PRs.

A new model builder option `compile_via_model` is added. If set, the body of function definitions will be piped through the model's exp builder to translate expansion AST into model AST, which is now stored in the model and accessible via `FunctionEnv::get_def()`. The model AST has been extended to handle new imperative expression constructs, but the translator has not yet been extended to support imperative constructs. Test infra has been set up as well, to be refined as we proceed.

### DIFF
--- a/language/move-model/src/ast.rs
+++ b/language/move-model/src/ast.rs
@@ -406,6 +406,18 @@ pub enum ExpData {
     Block(NodeId, Vec<LocalVarDecl>, Exp),
     /// Represents a conditional.
     IfElse(NodeId, Exp, Exp, Exp),
+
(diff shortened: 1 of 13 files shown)


meng-xu-cs (approved):
Thanks for kickstarting the effort!

PR #1022, opened 2023-03-29
[session] Expose struct loading

amnn:
## Motivation

This operation loads the struct type information.  It differs from `load_type` which, when loading a generic struct type, will also instantiate type parameters.  Having this separate entry into the loader allows us to implement type loading support in the presence of package upgrades from the adapter layer.

## Test Plan

```
$ env SOLC=... cargo nextest --no-fail-fast
```


awelc (approved):
LGTM. I was wondering if it would make sense to expose `verify_ty_args` as well to avoid re-implementing it (as the idea is to have custom implementation of `load_type` I believe), but it does sound like too fine level of granularity.

amnn:
Yeah I did consider it, but ended up going for reimplementing it, because we do expose get_type_abilities on the session.

PR #1023, opened 2023-03-30
[move-package] Support dependency overrides

awelc:
## Motivation

When building a dependency graph, different versions of the same (transitively) dependent package can be encountered. If this is indeed the case, a single version must be chosen by the developer to be the override, and this override must be specified in a manifest file whose package dominates all the conflicting "uses" of the dependent package. These overrides are taken into consideration during the dependency graph construction.

When constructing the graph (top to bottom) for internal dependencies (external dependencies are batch-processed at the end of the graph construction), we maintain a set of the current overrides collected when processing dependencies (starting with an empty set), and do the following when processing a package (starting at root):

- process overrides first to collect a list of overrides to be added to the set (if they are not yet in the set - outer overrides "win") and then process non-overridden dependencies using a freshly updated overrides set                                                                    
- when trying to insert a package into the graph, mark it as being on the "override path" if the set of overrides is non-empty:                                                                   
   - if no package exists in the graph, insert it                                                
   - if a conflicting package already exists in the graph, override it if an override can be found in the set (if it does not, report an error)                                            
   - if the same package already exists in the graph, and this package is on the "override path" keep checking its dependencies to make sure that previously used overrides are correct (dominate all uses of the package)                                                            
- after a package is fully processed, remove its own overrides from the set.

External dependencies are provided by external resolvers as fully formed dependency sub-graphs that need to be inserted into the "main" dependency graph being constructed. Whenever an external dependency is encountered, it's "recorded" for future batch-processing along with the set of overrides available at the point of sub-graph insertion. When processing an individual sub-graph insertion, we do a depth first search of the subgraph in order to:

- detect which of the sub-graph's packages need to be overridden (in which case their dependencies in the sub-graph no longer should be inserted into the main graph)
- avoid inserting sub-graph edges into the main dependency graph if they belong to overridden    
packages

Yes

## Test Plan

Tests added


language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -42,6 +42,44 @@ use super::{
 ///
 /// In order to be `BuildConfig` agnostic, it contains `dev-dependencies` as well as `dependencies`
 /// and labels edges in the graph accordingly, as `DevOnly`, or `Always` dependencies.
+///
+///
```
amnn, line 46: nit: just one line break between paragraphs?

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -415,45 +491,106 @@ impl DependencyGraph {
...
-                    PackageWithResolverTOML(&ext_pkg),
-                );
-            }
+        // unwrap safe as the table must have the package if the graph has it
+        let merged_pkg = ext_table.get(&merged_pkg_name).unwrap();
+        // unwrap is safe as all edges have a Dependency weight
+        let merged_dep = ext_graph.edge_weight(from, merged_pkg_name).unwrap();
+        self.package_graph
+            .add_edge(from, merged_pkg_name, merged_dep.clone());
+        self.merge_pgk(
```
amnn, line 505:
```suggestion
        self.merge_pkg(
```

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -415,45 +491,106 @@ impl DependencyGraph {
...
-                    ext_name,
-                    PackageWithResolverTOML(&ext_pkg),
-                );
-            }
+        // unwrap safe as the table must have the package if the graph has it
+        let merged_pkg = ext_table.get(&merged_pkg_name).unwrap();
+        // unwrap is safe as all edges have a Dependency weight
+        let merged_dep = ext_graph.edge_weight(from, merged_pkg_name).unwrap();
+        self.package_graph
+            .add_edge(from, merged_pkg_name, merged_dep.clone());
```
amnn, line 504: nit: Is it possible to merge the package first, and then add the edge if that succeeds?

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -385,15 +454,22 @@ impl DependencyGraph {
         Ok(())
     }
 
-    /// Add the graph in `extension` to `self` consuming it in the process.  Assumes the root of
-    /// `extension` is the only shared node between the two, and fails if this is not the case.
-    /// Labels packages coming from `extension` as being resolved by `resolver`.
+    /// Add the graph in `extension` to `self. Assumes the root of `extension` is the only shared
+    /// node between the two, and fails if this is not the case.  Labels packages coming from
+    /// `extension` as being resolved by `resolver`.
```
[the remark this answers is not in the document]
amnn: Also, can you document what `from` and `merged_pkg_name` are used for here? `overrides` is more or less straightforward, but those other two are not.

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -415,45 +491,106 @@ impl DependencyGraph {
...
 
-        // Because all the packages in `extensions`'s package table didn't exist in `self`'s, all
-        // `ext_graph`'s edges are known to not occur in `self.package_graph` and can be added
-        // without worrying about introducing duplicate edges.
-        for (from, to, dep) in ext_graph.all_edges() {
-            self.package_graph.add_edge(from, to, dep.clone());
+            Entry::Occupied(mut entry) => {
+                if let Some(overridden_pkg) = overrides.get(&ext_name) {
+                    // override found - use it and return - no reason to process its dependencies
+                    entry.insert(overridden_pkg.clone());
```
amnn, line 565: Do we need to do the acyclicity check here?

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -415,45 +491,106 @@ impl DependencyGraph {
...
+        )?;
 
-            match self.package_table.entry(ext_name) {
-                Entry::Vacant(entry) => {
-                    entry.insert(ext_pkg);
-                }
+        Ok(())
+    }
+
+    fn merge_pgk(
```
amnn, line 517: This is quite a complicated recursive function, it deserves a pretty hefty doc comment I think!  (Could be that part of the big doc comment that you have at the top could go here?)

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -493,55 +630,154 @@ impl DependencyGraph {
...
+        progress_output: &mut Progress,
+    ) -> Result<()> {
+        // partition dep into overrides and not
+        let (overridden_deps, deps): (Vec<_>, Vec<_>) =
+            dependencies.iter().partition(|(_, dep)| {
+                if let PM::Dependency::Internal(d) = dep {
+                    d.dep_override
+                } else {
+                    false
+                }
```
amnn, line 684: Not 100% sure this is actually neater, but:

```suggestion
                matches!(dep, PM::Dependency::Internal(InternalDependency {
                    dep_override: true,
                    ..
                }))
```

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -682,24 +927,41 @@ impl DependencyGraph {
...
+            // that the overrides introduced on this path correctly dominate all "uses" of a given
+            // package
             Entry::Occupied(entry) if entry.get() == &pkg => {
-                return Ok(());
+                if entry.get().overridden_path {
+                    // check if acyclic to avoid infinite recursion
+                    self.check_acyclic()?;
+                    self.override_verify(&pkg, name, overrides)?;
+                }
+                return Ok(pkg);
```
amnn, line 947: Should this be guarded by the `else` of the `if` above?  The comment above seems to suggest so.

This also implies a missing test case.

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -711,16 +973,67 @@ impl DependencyGraph {
...
+    fn override_verify(
+        &self,
+        pkg: &Package,
+        name: PM::PackageName,
+        overrides: &BTreeMap<PM::PackageName, Package>,
+    ) -> Result<()> {
+        // check if any (should be 0 or 1) edges are the overrides of pkg
+        let pkg_overrides: Vec<_> = self
+            .package_graph
+            .neighbors_directed(name, Direction::Incoming)
```
amnn, line 1000: nit: use `edges_directed` to avoid the lookup on `edge_weight` and the `unwrap`.

amnn (commented):
Did the initial pass, I think it's largely there, thanks @awelc, I especially appreciate the comprehensive test suite !  High-level notes:

- Some of the new functions need some love vis-a-vis doc comments -- it's a bit of a struggle to see how they all fit together without it, especially the recursive ones.
- There's a comment that suggests that we should not exit early if we are on an overridden path that seems to exit early anyway -- potential bug and missing test case there?
- We're generally missing tests for overrides that themselves contain dependencies, which certainly has potential for some hairy edge cases.
- The tests make heavy use of the `version` field to create different versions of the same package even though they all point to the same place.  I'm not sure how legitimate that is, because we don't really pay attention to the version during resolution (if anything, we should probably get rid of the version as a legitimate field on dependencies).  Consider using different locations to represent different versions of a package.
- Copybara -- we should probably port this PR over to the `sui` repo land it there and let `copybara` port it back.  I was able to do that before by generating a patch file from my commit, and then applying it in `sui`.

language/tools/move-package/src/resolution/dependency_graph.rs
awelc: That was kind of intentional to separate to somewhat different parts of this large doc comment but I agree that it does not look that great

language/tools/move-package/src/resolution/dependency_graph.rs
awelc: Good point. I assumed we might as I don't think we have any guards against actively creating cycles during graph construction nor does `override_verify` include such check when traversing the graph, which could lead to infinitely traversing the cycle.

I will try creating an example of such behavior to either confirm the use of  acyclicity here or rule it out.

language/tools/move-package/src/resolution/dependency_graph.rs
awelc: I was trying to find a better fitting function, but I don't think that `edges_directed` is implemented for `GraphMap` (nor is the `IntoEdgesDirected`)

language/tools/move-package/src/resolution/dependency_graph.rs
awelc: > Should this be guarded by the `else` of the `if` above? The comment above seems to suggest so.

This was not my intention. The idea was that if  package already in the graph (the same as package we are trying to insert)  is NOT on the upgrade path, we are OK to return early. Otherwise, either `override_verify` will return an error or not, in which case it is also OK to return (early as there are no more packages to analyze and insert here).

language/tools/move-package/src/resolution/dependency_graph.rs
awelc: Will do!

language/tools/move-package/src/resolution/dependency_graph.rs
awelc: Should be!

language/tools/move-package/tests/test_sources/diamond_problem_dep_incorrect_override_occupied/Move.resolved
```diff
@@ -0,0 +1,3 @@
+Failed to resolve dependencies for package 'Root': Resolving dependencies for package 'B': Conflicting dependencies found:
```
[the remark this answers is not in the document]
awelc: I agree that this would be ideal but it's currently impossible... It's due to the restriction that when we process an overridden dependency we do not add this dependency to the set of overrides:
```
        // It's also pretty important that we do not extend overrides with the override being
        // currently processed as we have no way of knowing if it's the correct one (dominating all
        // package "uses"). If we did, we could override an existing (non-overridden) entry in the
        // graph with an incorrect override without being able to detect it.
```
In other words, in this case, there isn't even a notion of override when processing this particular pair of conflicting packages...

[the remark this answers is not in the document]
awelc: I have updated the doc comment to explain the restriction leading to different error messages a bit better:
```
        // It's also pretty important that we do not extend overrides with the override being
        // currently processed. The reason for it is that in order to detect incorrect overrides
        // (such that do not dominate all package "uses") we rely on the package being reachable via
        // different paths:
        // - if it's reached via an overridden path again for the same override, it's OK
        // - if it's reached via an overridden path again for a different override, it's an error
        // - if it's reached via a non-overridden path, it's an error (insufficient override)
        //
        // While the first type of error could still be detected if we injected the currently
        // processed override into the overrides set, the second one would not.  Consider
        // diamond_problem_dep_incorrect_override_occupied example (in tests) to see a situation
        // when a non-overridden path is chosen first to insert the package and then insufficient
        // override could be considered correct if we injected it into the overrides set (as we will
        // not have another path to explore that would disqualify it).

```

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -535,62 +534,36 @@ impl DependencyGraph {
...
+            self.merge_pkg(
                 dst_pkg.clone(),
                 dst,
                 ext_graph,
                 ext_table,
                 resolver,
                 overrides,
             )?;
+            // unwrap is safe as all edges have a Dependency weight
+            let ext_dep = ext_graph.edge_weight(ext_name, dst).unwrap();
```
amnn, line 565: nit: Is it possible to avoid the `unwrap` by iterating over `edges_directed` instead of `neighbors_direceted`?

language/tools/move-package/tests/test_sources/diamond_problem_dep_incorrect_override_cycle/Move.toml
```diff
@@ -0,0 +1,27 @@
+# Dependency graph and and override (ov) - when override is processed there is no entry in the graph
```
amnn, line 1: nit: comment

```suggestion
# Dependency graph and override (ov) - when override is processed there is no entry in the graph
```

language/tools/move-package/src/resolution/dependency_graph.rs
```diff
@@ -930,64 +917,122 @@ impl DependencyGraph {
...
             }
 
             // Seeing the same package again, but pointing to a different dependency: Not OK unless
             // there is an override.
             Entry::Occupied(mut entry) => {
                 if let Some(overridden_pkg) = overrides.get(&name) {
-                    // override found - use it
-                    entry.insert(overridden_pkg.clone());
-                    entry.into_mut()
+                    // override found - use it and return - its dependencies have already been
```
amnn, line 1012: It seems like we only care to check if there's an override when there's a potentially conflicting set of dependencies, but if the package table is empty we just insert the package, and don't check for overrides.  Is this because if there is an override, it will be in the package table already, or is this something that we need to change? (If the former, we should add a comment on the `Vacant` case as well.

language/tools/move-package/tests/test_sources/diamond_problem_dep_conflict/deps_only/C/Move.toml
```diff
@@ -0,0 +1,3 @@
+[package]
+name = "C"
+version = "0.0.0"
```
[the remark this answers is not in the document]
amnn: Recording a conversation we had offline:  I think the test set-up here is fine, but it's highlighted a peculiarity with dependency specification (that we treat dependencies as different based on the version field, even though we don't use the version field during dependency resolution), which we should change in future (probably by removing the `version` field).

language/tools/move-package/src/resolution/dependency_graph.rs
amnn: I seem to see it here, in case I'm missing something: 

https://docs.rs/petgraph/latest/petgraph/graphmap/struct.GraphMap.html#method.edges_directed

language/tools/move-package/src/resolution/dependency_graph.rs
amnn: Got it, that makes sense, thanks for clarifying!

amnn (commented):
Okay, just one final question from me, about how we handle overrides in the case where the package table is vacant at the entry for that overridden package, and then let's ship this thing!

language/tools/move-package/src/resolution/dependency_graph.rs
awelc: Good point! And your assessment is correct - the override would have already been inserted (and at the point of the override insertion, the package in question is not in the overrides set). I have put a comment on the `Vacant` case accordingly.

awelc:
> Okay, just one final question from me, about how we handle overrides in the case where the package table is vacant at the entry for that overridden package, and then let's ship this thing!

As we discussed offline, this situation cannot happen as a package is inserted into the overrides set after it's "processed" that is after its insertion is attempted in `process_dependency` . In fact, we return a package from the `package_table` entry to have it inserted into the overrides set.

language/tools/move-package/tests/test_sources/diamond_problem_dep_with_deps_conflict/.#Move.toml
```diff
@@ -0,0 +1 @@
+adamwelc@Adams-Laptop.local.85810
```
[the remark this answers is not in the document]
awelc: Dang!!! Removed!

PR #1034, opened 2023-04-12
[bug fix] Remove incorrect overflow guard

tnowacki:
## Motivation

Thanks to [Zellic](https://www.zellic.io/) and @fcremo for the find and fix! 

The guard in `get_successors` was intended to prevent an overflow, but the check erroneously assumed that the last instruction was a return or an abort. In the case that it was an unconditional jump, the successors would not be captured in the CFG. And as such, the target code would not be visited (if this was the only predecessor). This means that locals safety and reference safety (and any custom, control flow based checks) would not be run on the skipped code. 

## Test Plan

- New tests


runtian-zhou (approved):
Thanks Todd!

Issue #1038, opened 2023-04-15
[Feature Request] Support  multiple languages in comments


WGB5445:
# 🚀 Feature Request

## Motivation

When writing code, you may write comments in multiple languages.
but it is not supported in Move files, which is very inconvenient

## Pitch

**Are you willing to open a pull request?** (See [CONTRIBUTING](../../CONTRIBUTING.md))
yes

zsluedem:
https://github.com/move-language/move/issues/262#issuecomment-1190378746 for your reference

Issue #262, opened 2022-07-12
[Bug] utf8 is not allowed


zsluedem:
# 🐛 Bug

The compiler currently only accepts ASCII characters.

https://github.com/move-language/move/blob/a8383d88fa48f4e1e0e91264cffbbd27136e4732/language/move-compiler/src/parser/comments.rs#L15-L37

## To reproduce

**Code snippet to reproduce**
```rust
module 0xCAFE::BasicCoin {
    struct Coin has key {
        value: u64,
    }
    // 你好啊
    public fun mint(account: signer, value: u64) {
        move_to(&account, Coin { value })
    }
}
```

**Stack trace/error message**
```
error[E01001]: invalid character
  ┌─ ./sources/FirstModule.move:5:8
  │
5 │     // 你好啊
  │        ^^ Invalid character '你' found when reading file. Only ASCII printable characters, tabs (\t), lf (\n) and crlf (\r\n) are permitted.


```

## Expected Behavior

No compiler error should be showed.

## System information

**Please complete the following information:**
move-cli 0.1.0
rustc 1.60.0 (7737e0b5c 2022-04-04)
ubuntu18

Although the messages are not necessary.

## Additional context
No

sblackshear:
We intentionally restrict comments to ASCII characters to prevent [trojan source](https://trojansource.codes/) vulnerabilities.

PR #1039, opened 2023-04-15
[move-compiler]Support utf8 language in comments 

WGB5445:
## Motivation

In code writing, there is no way to use other languages as comments. So I hope to support other languages

So, I changed the character control for comments to allow readable characters and newline tabs as well as carriage return line feeds

yes

## Test Plan
Unit Test

## Issue
#1038


WGB5445:
#262

PR #1055, opened 2023-05-24
[RFC] [abi] implemented a abi for structs using serde-reflection

runtian-zhou:
## Motivation

Compile struct layout into serde-reflection format that could be codegened using serde-generate crate.

Yes

## Test Plan

Tests added.


sblackshear:
Always love seeing more serde-generate integration! But could you help me understand the difference between this and the existing layout generation functionality? https://github.com/move-language/move/blob/main/language/tools/move-bytecode-utils/src/layout.rs is the layout generator, and is exposed via the CLI https://github.com/move-language/move/blob/main/language/tools/move-cli/src/sandbox/commands/generate.rs#L14 with `move sandbox generate struct-layouts`. This was the PR that added this https://github.com/move-language/move/pull/652

PR #1069, opened 2023-08-12
[move-compiler] Fixed abstract interpreter bug with continue statement

ksolana:
continue in while loop causes an unexpected error.

Cherry-pick: MystenLabs/sui#13349
Bugfix: move-language#1068


sblackshear (approved):

sblackshear:
Thanks for the pick!

Issue #1078, opened 2023-10-03
[Bug]: update rust version to 1.66.0 from 1.65.0


bajpai244:
# 🐛 Bug

When someone attempts to install the `move-cli` with `cargo install --path language/tools/move-cli` , it will fail with `package `toml_edit v0.19.15` cannot be built because it requires rustc 1.66.0 or newer, while the currently active rustc version is 1.65.0`

More context: I am following the [move tutorial](https://github.com/move-language/move/tree/main/language/documentation/tutorial)l

## To reproduce

- clone the repo
- cd and then `./scripts/dev_setup.sh -ypt`
- `cargo install --path language/tools/move-cli`

**Stack trace/error message**
```
failed to compile `move-cli v0.1.0 (/Users/harshbajpai/Desktop/Projects/move/language/tools/move-cli)`, intermediate artifacts can be found at `/Users/harshbajpai/Desktop/Projects/move/target`

Caused by:
  package `toml_edit v0.19.15` cannot be built because it requires rustc 1.66.0 or newer, while the currently active rustc version is 1.65.0
  Either upgrade to rustc 1.66.0 or newer, or use
  cargo update -p toml_edit@0.19.15 --precise ver
  where `ver` is the latest version of `toml_edit` supporting rustc 1.65.0
```

## Expected Behavior

The compilation should succeed.

## System information

**Please complete the following information:**
- Mac M2

wrwg:
Please see #1076

ducthann:
In the rust-toolchain file, change '1.65.0' to '1.67.0' before installing. It should work.

dongowu:
I also encountered this situation today，change 1.66.0 or newer is ok ,

PR #1079, opened 2023-10-03
fix(#1078): fix move-cli installation

bajpai244:
## Motivation

This PR aims to fix the current compilation issue of the move-cli as described in #1078.


wrwg:
Can you fix the error which CI discovered by new rules coming with this Rust version? You can run `cargo xclippy --workspace --all-targets` locally to see the issue. Thanks!

PR #1083, opened 2023-11-06
Move VM Transfer callback 

35359595:
### Concept implementation of Move -> Pallet methods glue.
* Callback method as member of 'Mvm' to execute transfer;
* type alias and error enums to be used as references on Substrate side;


35359595:
wrong repo

PR #1100, opened 2024-03-01
[tutorial] Advise using `--locked` in `cargo install`

wiktor-k:
Hi :wave: 

## Motivation

This patch makes cargo use the lockfile for dependencies. Without that, at the time of writing this patch, the install fails with the following error:

    error: failed to compile `move-cli v0.1.0 (/home/wiktor/src/open-source/move/language/tools/move-cli)`, intermediate artifacts can be found at `/home/wiktor/tmp/cargo`

    Caused by:
      package `toml_datetime v0.6.5` cannot be built because it requires rustc 1.67 or newer, while the currently active rustc version is 1.65.0

See: https://github.com/rust-lang/cargo/issues/7169#issuecomment-514361495

Alternative solution would be upgrading Rust version but adding `--locked`  to the tutorial is far less invasive.

Yes :+1: 

## Test Plan

Try `cargo install --path language/tools/move-cli` and see it fail, now try `cargo install --locked --path language/tools/move-cli`. It should succeed.

### DIFF
--- a/language/documentation/tutorial/README.md
+++ b/language/documentation/tutorial/README.md
@@ -49,7 +49,7 @@ source ~/.profile
 Next, install Move's command-line tool by running this commands:
 
 ```bash
-cargo install --path language/tools/move-cli
+cargo install --locked --path language/tools/move-cli
 ```
 
 You can check that it is working by running the following command:


wrwg (approved):

wiktor-k:
Hmm... funny thing, one of the CI jobs failed with the same exact reason ( --locked flag missing on cargo install). I could fix it in this PR if you don't mind.

PR #1105, opened 2024-03-05
Update rust toolchain to 1.70.0

ksolana:
Rust version is old and blocking other changes like https://github.com/move-language/move/pull/1104 which blocks https://github.com/move-language/move/pull/1103


wrwg (commented):
Hi @ksolana , the main branch is kind of frozen at this point. It is behind Sui and Aptos Move branches and fixes are not longer upstreamed or downstreamed. I suggest to work on a fork until you are ready to integrate with Aptos or Sui. Also @sblackshear FYI.

PR #1111, opened 2024-04-27
Update rust-toolchain

NadigerAmit:
## Motivation

Fix the compilation error that occurs when executing the below step :

`cargo install --path language/tools/move-cli`

The reason is that some of the packages required a newer version of Rust, but the current one is pinned at 1.65.0 in rust-toolchain.

Hence changed the  rust-toolchain to 1.77.2. It works good.

Yes

## Test Plan

Changed the rust-toolchain from 1.65.0 to 1.77.2.
The compilation is successful.


NadigerAmit:
since there is already a PR on the same I canceled my pull request: 

Existing PR:  https://github.com/move-language/move/pull/1108

